PR #3, opened 2021-02-18
Unity

jsarzy:


runner.py
```diff
@@ -0,0 +1,647 @@
...
+
+class Color:
+    FAIL = '\033[91m'
+    OK = '\033[92m'
+    IGNORE = '\033[93m'
+    END = '\033[0m'
+
+    @staticmethod
+    def colorify(string, color):
+        return color + string + Color.END
```
kemonats, line 49:
```suggestion
        return f"{color}{string}{Color.END}"
```
It would be better  to use f-string.

runner.py
```diff
@@ -0,0 +1,647 @@
...
+def args_log_level(arg):
+    levels = {'DEBUG': 1, 'INFO': 2, 'WARNING': 3, 'ERROR': 4}
+    arg = levels.get(arg)
+
+    if not arg or arg not in (1, 2, 3, 4):
+        logging.warning(f"{arg} is not valid verbosity level")
+        sys.exit(1)
+
+    # Convert to logging module level (10, 20, 30, 40)
+    return arg * 10
```
kemonats, line 559:
```suggestion
    levels = {'DEBUG': 1, 'INFO': 2, 'WARNING': 3, 'ERROR': 4}
    arg1 = levels.get(arg)

    if not arg1 or arg1 not in (1, 2, 3, 4):
        logging.warning(f"{arg} is not valid verbosity level")
        sys.exit(1)

    # Convert to logging module level (10, 20, 30, 40)
    return arg1 * 10
```
Now if  `arg` does not contain a valid verbosity level a confusing message is displayed:
`WARNING:root:None is not valid verbosity level`

kemonats, line 559:
```suggestion

    if arg not in level:
        logging.warning(f"{arg} is not valid verbosity level")
        sys.exit(1)

    # Convert to logging module level (10, 20, 30, 40)
    return level[arg] * 10
```

runner.py
jsarzy: I agree it is more consistent.

runner.py
jsarzy: I changed a little bit arguments handling.

runner.py
```diff
@@ -0,0 +1,653 @@
...
+                        default='info',
+                        choices=sum(verbosity.items(), ()),
+                        help="Specify verbosity level. By default uses level info (2).")
+
+    args = parser.parse_args()
+
+    if isinstance(args.verbose, str):
+        args.verbose = verbosity[args.verbose]
+    # Convert to logging module's level (10, 20, 30, 40)
+    args.verbose *= 10
```
kemonats, line 591: I don't understand why you use two variable types `str` and `int` in the `choices` argument. 
There can only be one type, default `str` or set by `type` argument.
Now there is:
```
python3 runner.py -v 1
usage: runner.py [-h] [-T {ia32-generic}] [-t TEST] [--no-build]
                 [-v {debug,1,info,2,warning,3,error,4}]
runner.py: error: argument -v/--verbose: invalid choice: '1' (choose from 'debug', 1, 'info', 2, 'warning', 3, 'error', 4)
```
My proposition:
```suggestion
    parser.add_argument("-v", "--verbose",
                        default='info',
                        choices=tuple(verbosity.keys()),
                        help="Specify verbosity level. By default uses level 'info'.")

    args = parser.parse_args()

    # Convert to logging module's level (10, 20, 30, 40)
    args.verbose = verbosity[args.verbose] * 10
```

runner.py
```diff
@@ -0,0 +1,653 @@
...
+        for test_config in config['tests']:
+            # Test consist of one pair (key, value) which is (name, config)
+            if len(test_config.keys()) != 1:
+                raise YAMLParserError(self, "test name not found in test config")
+
+            name = list(test_config.keys())[0]
+            test_config = test_config[name]
+
+            # Add useful staff to config
+            name = remove_prefix(str(self.path), str(PRTOS_PROJECT_DIR) + '/') + '/' + name
```
kemonats, line 501: f-string

runner.py
```diff
@@ -0,0 +1,653 @@
...
+            except yaml.YAMLError as exc:
+                raise YAMLParserError(self, exc) from exc
+
+        self.parsed_tests = []
+        # Current path is /example/test/test.yaml - get rid of test.yaml in path
+        self.path = path.parents[0]
+
+    def parse_test_case(self, config):
+        allowed_keys = ['syspage', 'type', 'harness', 'name', 'path', 'attribute']
+        mandatory_keys = ['syspage', 'name', 'path']
```
kemonats, line 448:
```suggestion
        allowed_keys = ('syspage', 'type', 'harness', 'name', 'path', 'attribute')
        mandatory_keys = ('syspage', 'name', 'path')
```
`tuple` are faster and take up less  memory.

runner.py
```diff
@@ -0,0 +1,653 @@
...
+    def create(target):
+        if target == 'ia32-generic':
+            return IA32Builder()
+
+        return TargetBuilder(target)
+
+
+class YAMLParserError(Exception):
+    def __init__(self, parser, message="Parser error"):
+        message = f"An error occured while parsing {parser.path}:\n" + message
```
kemonats, line 430:
```suggestion
        message = f"An error occured while parsing {parser.path}:\n{message}"
```

runner.py
```diff
@@ -0,0 +1,653 @@
...
+        )
+
+        out, err = proc.communicate()
+        if proc.returncode != 0 and handle_error:
+            logging.error(f"Command {' '.join(args)} for {self.target} failed!")
+            logging.error(out)
+            logging.error(err)
+            # Fail at build stage should not happen, exit immediately
+            sys.exit(1)
+        elif proc.returncode != 0:
```
kemonats, line 378: `proc.returncode != 0` is this a success or error?

runner.py
```diff
@@ -0,0 +1,653 @@
...
+            logging.info(f"Command {' '.join(args)} for {self.target} success!")
+
+        return proc.returncode, out, err
+
+    def build(self, syspage=None):
+        if syspage:
+            syspage_with_defaults = list(set(syspage + TargetBuilder.SYSPAGE[self.target]))
```
kemonats, line 385: No checking  for valid target.

runner.py
```diff
@@ -0,0 +1,653 @@
...
+        self.unit_test_results = []
+
+    def report(self):
+        super().report()
+
+        for test in self.unit_test_results:
+            if test.status == UnitTest.FAIL:
+                logging.info('\t' + str(test))
+            else:
+                logging.debug('\t' + str(test))
```
kemonats, line 282:
```suggestion
                logging.info(f"\t{test})
            else:
                logging.debug(f"\t{test})
```
f-string

runner.py
```diff
@@ -0,0 +1,653 @@
...
+
+
+class QemuRunner:
+    def __init__(self, run_cmd):
+        self.run_cmd = run_cmd
+        # Remove existing serial flag and replace it with stdio
+        self.run_cmd = [flag for flag in self.run_cmd if flag[0] != '-serial']
+        self.run_cmd.append(('-serial', 'stdio'))
+        # Flatten list and turn it into string
+        self.run_cmd = " ".join(list(sum(self.run_cmd, ())))
```
kemonats, line 81:
```suggestion
        # Remove existing serial flag and replace it with stdio
        self.run_cmd = [flag for flag in run_cmd if flag[0] != '-serial']
        self.run_cmd.append(('-serial', 'stdio'))
        # Flatten list and turn it into string
        self.run_cmd = " ".join(sum(self.run_cmd, ()))
```
`list()` is no needed, `sum()` return `tuple`.
Instead of `sum()` you  can use `itertools.chain()`

runner.py
```diff
@@ -0,0 +1,653 @@
...
+            'uart16550',
+            'pc-ata',
+            'psh'
+        ],
+        'riscv64-virt': [
+            'uart16550',
+            'dummyfs',
+            'psh'
+        ]
+    }
```
kemonats, line 330: I suggest making this `class` more abstract.
Here define `SYSPAGE` e.g empty `SYSPAGE = []`.
Specific `SYSPAGE` values should be defined in classes for particular targets e.g. for `ia32-generic`
```
class IA32Builder(TargetBuilder):
    SYSPAGE = ( 'uart16550', 'pc-ata', 'psh' )

    def __init__(self):
        super().__init__('ia32-generic')
```

runner.py
jsarzy: I agree.

runner.py
jsarzy: I agree that immutable tuples are betters here. Changed

runner.py
```diff
@@ -0,0 +1,707 @@
...
+                        help="Filter targets on which test will be build and ran. "
+                             "By default uses environment variable TARGET. If variable "
+                             "is not set then runs tests on all aviable targets. "
+                             "Flag can be used multiple times.")
+
+    parser.add_argument("-t", "--test",
+                        default=[], action='append', type=args_file,
+                        help="Specify directory in which test are will be searched. "
+                             "If flag is not used then runner search for tests in "
+                             "phoenix-rtos-tests dictionary. Flag can be used multiple times.")
```
kemonats, line 659:
```suggestion
                             "phoenix-rtos-tests directory. Flag can be used multiple times.")
```

nalajcie:
if the pull request is ready for review, then remove the WIP status (click `ready for review`).

I'm assuming this is the case now, so I'm removing the WIP status now.

runner.py
```diff
@@ -0,0 +1,707 @@
...
+        # We run syspage progs from filesystem, set only default syspage
+        self.env['SYSPAGE'] = " ".join(self.SYSPAGE[self.target])
+
+    def postbuild_action(self, syspage=None, **kwargs):
+        if not syspage:
+            return
+
+        logging.info(f"Installing {' '.join(syspage)} to filesystem for {self.target}")
+        build_dir = PRTOS_PROJECT_DIR / f"_build/{self.target}/prog.stripped"
+        syspage_dir = pathlib.Path("/root/syspage")
```
nalajcie, line 438: no, install in `/bin`, `/syspage` is for programs which are embedded in kernel binary

runner.py
```diff
@@ -0,0 +1,707 @@
...
+            color = Color.IGNORE
+
+        status = Color.colorify(self.status, color)
+        return f"{self.target}: {self.name}: {status}"
+
+    def ignored(self):
+        return self.status == TestCase.IGNORE
+
+    def failed(self):
+        return self.status == TestCase.FAIL
```
nalajcie, line 212: or `TestCase.FAIL_TIMEOUT`

Right now timeouted tests are considered not an error.

runner.py
```diff
@@ -0,0 +1,707 @@
...
+class TestCase:
+    PASS = "PASS"
+    FAIL = "FAIL"
+    FAIL_TIMEOUT = "FAIL TIMEOUT"
+    IGNORE = "IGNORE"
+
+    def __init__(self, name, target, syspage=None, status=None):
+        self.name = name
+        self.target = target
+        self.syspage = syspage
```
nalajcie, line 190: `syspage` is an invalid name, this is the binary path to be executed - it has no connection with the actual kernel syspage

runner.py
```diff
@@ -0,0 +1,707 @@
...
+        self.status = TestCase.FAIL
+
+        if not self.harness:
+            return False
+
+        if self.syspage:
+            try:
+                self.exec_syspage(proc)
+            except pexpect.exceptions.TIMEOUT as exc:
+                logging.info(f"{self.target}: Executing syspage {self.syspage} failed!")
```
nalajcie, line 247: binary, not syspage

runner.py
```diff
@@ -0,0 +1,707 @@
...
+                    raise ValueError("Unit test harness: results do not match")
+
+                return test_results
+
+
+class TestCase:
+    PASS = "PASS"
+    FAIL = "FAIL"
+    FAIL_TIMEOUT = "FAIL TIMEOUT"
+    IGNORE = "IGNORE"
```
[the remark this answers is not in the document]
nalajcie: do we differentiate between test not executed and test executed but the result ignored (eg. because of unity test case configuration?)

runner.py
```diff
@@ -0,0 +1,707 @@
...
+
+        self.postbuild_action(syspage)
+
+
+class IA32Builder(TargetBuilder):
+    def __init__(self):
+        super().__init__('ia32-generic')
+
+    def prebuild_action(self, syspage=None, **kwargs):
+        # We run syspage progs from filesystem, set only default syspage
```
nalajcie, line 429: so: they are not syspage progs, just standard ELF binaries

runner.py
```diff
@@ -0,0 +1,707 @@
...
+        self.env['SYSPAGE'] = " ".join(self.SYSPAGE[self.target])
+
+    def postbuild_action(self, syspage=None, **kwargs):
```
nalajcie, line 432: invalid naming - `syspage`

runner.py
```diff
@@ -0,0 +1,707 @@
...
+                        default=[], action='append', type=args_file,
+                        help="Specify directory in which test are will be searched. "
+                             "If flag is not used then runner search for tests in "
+                             "phoenix-rtos-tests dictionary. Flag can be used multiple times.")
+
+    parser.add_argument("--no-build",
+                        default=False, action='store_true',
+                        help="Omitt building step.")
+
+    parser.add_argument("-v", "--verbose",
```
nalajcie, line 665: hmm, I would expect `-v` to set verbosity level to DEBUG, maybe `-l` like log-level? - or `-s[s]` for decreasing verbosity (silent)

sample/helloworld/test.py
```diff
@@ -0,0 +1,3 @@
+def harness(p):
+    p.expect_exact('Hello world!')
```
nalajcie, line 2: this is not the best idea, because to detect FAIL we need to wait 30s.

Please provide some more sensible examples (maybe using `readline()` for this and similar tests is enough?)

Maybe all harnesses should detect `psh >` and conclude the program has ended/failed with return code? (should we detect the error code?)

sample/test/fails/test-fail-output-1.c
```diff
@@ -0,0 +1,8 @@
+#include <stdio.h>
+
+int main(int argc, char **argv)
+{
+	puts("[Test FAIL TIMEOUT started]");
+	puts("1");
+	puts("This line will not be parsed");
```
nalajcie, line 7: add `while (1) 
;`

Program ending before the expected string should be detected as immediate failure instead of timeout

missing return?

sample/test/fails/test.yaml
```diff
@@ -0,0 +1,13 @@
+test:
+    attribute: ignore
+    tests:
+        - fail-output-1:
+            syspage: fail_output_1
+            harness: test-fail-output-1.py
+        - fail-output-2:
+            syspage: fail_output_2
+            harness: test-fail-output-2.py
+        - fail-unity:
```
nalajcie, line 10: please add test which will fail with exception (eg NULL dereference) and think about how to process it. Might be done in the separate PR to speed up merging of this one.

(i think it would be enough to have full output of every failed test)

runner.py
```diff
@@ -0,0 +1,707 @@
...
+            logging.warning(f"Test {self.name} reached EOF")
+            logging.warning(str(exc))
+        except Exception as exc:
+            logging.warning(f"Test {self.name} exception:\n {exc}")
+            raise exc
+
+        if res:
+            self.status = TestCase.PASS
+
+        return res
```
nalajcie, line 268: not sure, but return value is not being used (not being passed up from `DeviceRunner`?)

runner.py
```diff
@@ -0,0 +1,707 @@
...
+            for target, tests in self.tests_per_target.items():
+                target_builder = TargetBuilderFactory.create(target)
+
+                syspage = [test_case.syspage for test_case in tests]
+                syspage = list(set(syspage))
+                target_builder.build(syspage)
+
+        for target, tests in self.tests_per_target.items():
+            for test_case in tests:
+                self.runners[target].run(test_case)
```
nalajcie, line 597: The test case name should be printed before the actual result, so I know what I'm waiting for (`print_test_name(); run(); print_result();`)

nalajcie (changes requested):
Good job, need some polishing - but overall it's working :+1:.

Please address the most pressing issues so we can merge this PR, additional features might be done at a later time by separate PRs.

nalajcie:
also: please update `.gitignore` to ignore python compiled files

runner.py
```diff
@@ -0,0 +1,707 @@
+import argparse
```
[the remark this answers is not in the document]
jsarzy: I split it into multiple submodules which I placed into `trunner` directory. I left args parsing and setting log level in `runner.py`, but I can also move it into `trunner` directory.

runner.py
```diff
@@ -0,0 +1,707 @@
...
+import signal
+import subprocess
+import sys
+
+import yaml
+import serial
+import pexpect
+import pexpect.fdpexpect
+
+SCRIPT_PATH = pathlib.Path(__file__).absolute().resolve()
```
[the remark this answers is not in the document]
jsarzy: I changed it to `getcwd()` but for now runner will fail being executed outside `phoenix-rtos-project`.  I'm going to fix it at a later time.

runner.py
```diff
@@ -0,0 +1,707 @@
...
+PRTOS_TEST_DIR = SCRIPT_PATH.parents[0]
+PRTOS_PROJECT_DIR = SCRIPT_PATH.parents[1]
+_BOOT_DIR = PRTOS_PROJECT_DIR / "_boot"
+
+
+QEMU_CMD = {
```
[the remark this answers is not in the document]
jsarzy: Yes, I save it for the future.

runner.py
```diff
@@ -0,0 +1,707 @@
...
+    OK = '\033[92m'
+    IGNORE = '\033[93m'
+    END = '\033[0m'
+
+    @staticmethod
+    def colorify(string, color):
+        return f"{color}{string}{Color.END}"
+
+
+class DeviceRunner:
```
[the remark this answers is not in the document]
jsarzy: I agree. For now `Runners` are not much complex so I left it as it is. I will save it for next PR.

runner.py
```diff
@@ -0,0 +1,707 @@
...
+            color = Color.OK
+        elif status in (TestCase.FAIL, TestCase.FAIL_TIMEOUT):
+            color = Color.FAIL
+        elif status == TestCase.IGNORE:
+            color = Color.IGNORE
+
+        status = Color.colorify(self.status, color)
+        return f"{self.target}: {self.name}: {status}"
+
+    def ignored(self):
```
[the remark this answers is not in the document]
jsarzy: For now I removed `ignore` and `fail` methods.

runner.py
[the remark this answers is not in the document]
jsarzy: We do not differentiate these two cases. Test are only skipped if they has keyword 'ignore' in yaml config. Unity test will execute and if all units are ignored then test will receive `PASSED` status.

I changed `IGNORED` to `SKIPPED` in `TestCase` but I left `IGNORE` in `UnitTestResult` at is represents Unity test case which uses word `IGNORE`.

runner.py
```diff
@@ -0,0 +1,707 @@
...
+                self.config = yaml.safe_load(f_yaml)
+            except yaml.YAMLError as exc:
+                raise YAMLParserError(self, exc) from exc
+
+        self.parsed_tests = []
+        # Current path is /example/test/test.yaml - get rid of test.yaml in path
+        self.path = path.parents[0]
+
+    def parse_test_case(self, config):
+        allowed_keys = ('syspage', 'type', 'harness', 'name', 'path', 'attribute')
```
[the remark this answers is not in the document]
jsarzy: I will add YAML templates and some basic tests for parser in the next PR.

I changed YAML layouts (inspired by Github Actions) from:
```YAML
tests:
    attribute: ignore
    tests:
        - test.a:
            syspage: a
            harness: a.py
        - test.b:
            syspage: b
            type: unit
```
to:
```YAML
tests:      
    ignore: true
    tests:
        - name: a
          syspage: a
          harness: a.py
          
        - name: b
          syspage: b
          type: unit
```
Let me know what you think about these changes.

runner.py
```diff
@@ -0,0 +1,707 @@
...
+        try:
+            res = test.handle(proc)
+        finally:
+            if not res:
+                proc.kill(signal.SIGKILL)
+            else:
+                proc.kill(signal.SIGTERM)
+
+
+class UnitTest:
```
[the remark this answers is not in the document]
jsarzy: `UnitTestResult` represents result of single Unity unit test. It is used to interpret if unit test failed and to nicely print it. I will add few docstrings.

runner.py
jsarzy: I changed it to `/bin` however "syspage" handling have to be changed in the future PR. For now it's impossible to include multiple programs (like some driver and user app) into one test case (we do not choose which one we want to execute).

runner.py
jsarzy: I changed it to `exec_bin`.

runner.py
jsarzy: I left here `user_syspage`. We use `syspage` word in YAML config so it also have to be changed to `exec` or something. On the other hand only `ia32-generic` arch has fs system for now so on the other architectures it will be embedded for real into syspage. Let me know what you think.

In other words we have to distinguish:
1. Binaries embedded into syspage
2. Binaries embedded into syspage that are executed on OS start (like drivers)
3. Binaries embedded into filesystem

runner.py
jsarzy: I changed it to `--log-level`, `-l`

sample/helloworld/test.py
jsarzy: Now after invoking `exec prog` shell won't return to us (or I don't know how to do it). So is impossible to detect `psh >` after program execution.

sample/test/fails/test.yaml
jsarzy: This will require parsing output buffer in case of `FAIL_TIMEOUT` for exception pattern. Will be done in separate PR.

runner.py
jsarzy: For now I used `res` in `QemuRunner` to make decision if we should send `SIGKILL` or `SIGTERM` but now I think it won't be necessary.

runner.py
jsarzy: I added `log_test_sart` and `log_test_status`.

nalajcie (commented):
Ok, I like the current state of the code - I have one important feature request before merging this - I need to see the full output for the tests that have failed (some tests may fail only once in 1000 executions, we need as much debug info as we can get). 

I'll post comments to previous unresloved conversations in a second because I can't do it as a part of a code review.

Other things to consider:
 - rebase/squash the changes into fewer distinct commits for clearer history
 - add some docstrings describing the intended usage of the classes
 - I would add interfaces as soon as possible (may be in the separate PR) because they make architecture of the solution much more easy to grasp

runner.py
```diff
@@ -0,0 +1,707 @@
...
+
+    def build(self, syspage=None):
+        if syspage:
+            syspage_with_defaults = list(set(syspage + TargetBuilder.SYSPAGE[self.target]))
+            self.env['SYSPAGE'] = " ".join(syspage_with_defaults)
+
+        self.prebuild_action(syspage)
+
+        logging.info(f"Building {self.env['TARGET']} with syspage: {self.env['SYSPAGE']}")
+        self.run_command(['./docker-build.sh', 'clean', 'core', 'fs', 'image', 'project'])
```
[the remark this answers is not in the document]
nalajcie: ok, I think the build part of the script would be hardly used, but I might be wrong :)

runner.py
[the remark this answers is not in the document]
nalajcie: Yes, for that we would need do install the full package in system/virtualenv, so currently it's not worth the hassle

runner.py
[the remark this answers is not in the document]
nalajcie: I was thinking about proposing that in the previous review (as it simplifies the parser), but decided to leave the decision to You. I like the explicit name more ;)

runner.py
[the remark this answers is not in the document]
nalajcie: waiting for the docstrings :P

runner.py
nalajcie: I think that the "putting the binaries/files onto the device" and `executing programs` will need to be part of the `DeviceRunner` in the future. For some "big" targets it will be just appending them to the `rootfs` image, but for smaller (like iMX.RT1064) it might involve uploading binaries via USB/serial using `plo` bootloader during the bootup.

Regarding the "test binary execution" part - for now let's assume the "base" image would include all necessary drivers and tools, if the test would need multiple threads/processes it can spawn them internally.

runner.py
nalajcie: Linked to my previous comment - the `DeviceRunner` should provide the abstraction over "uploading files" (putting them on target) and executing programs. For sure we shouldn't use `syspage` naming, as for most targets we will not be using syspage for test files.

I assume building base image with all necessary drivers to run on a particular target is the responsibility of the build system, not a responsibility of this script, so we don't need to know set of necessary drivers for each of the targets to be workable.

Right now regarding different targets we can upload files by:
 - putting them in the rootfs image (`ia32-generic` (ext2), `armv7a7-imx6ull` (jffs), `riscv64-*` in the near future)
 - uploading files via `plo` - phfs on serial / USB CDC (`armv7a9-zynq7000`, `armv7m7-imxrt106x`, `armv7m7-imxrt117x`)
 - `armv7m4-stm32l4x6` - currently only by appending directly to syspage (AFAIK)

sample/helloworld/test.py
nalajcie: uch, I've always used `/bin/helloworld` which spawns new process, so after it's finish we're getting back to `psh`. I haven't checked the psh `exec` implementation, but I believe it uses `execve`, so swaps the current process contents with new one.

I think spawning new processes for "big" targets is a good execution strategy.

sample/test/fails/test.yaml
nalajcie: I don't care about automatically parsing the exception buffer, I need to see the full output of the failed testcase execution, eg:
Screenshot_20210406_164447

nalajcie:
Additional note: when rebasing/squashing ensure each commit has JIRA issue assigned

sample/test/fails/test.yaml
jsarzy: I added printing output buffer and pattern that we were looking for.

sample/helloworld/test.py
jsarzy: I added spawning new process. Parsing `psh >` after execution looks like a nice idea for additional check. I've written it down in my todo list for next PR.

nalajcie (approved):
Ok, the output is indeed visible in some failed cases (and i some - not).

I think this PR is good enough to be merged, we will tackle down remaining issues and important features as a separate PRs.

Please create JIRA tasks for bigger things on TODO list.

PR #5, opened 2021-04-13
runner: add 'install' target to Makefile

jsarzy:
- Add 'dir' keyword for config which specifies where test are placed on phoenix-rtos. By default if our path is `/home/user/phoenix-rtos-project/myproject/path/to/test/test.yaml` then test are run from `/bin/myproject/path/to/test/{syspage}`. Using 'dir' keyword we can change default path and run tests from wherever we want.
- Add `install_fs` target to Makefile.
- Decrease pyexpect timeout to 8 seconds

Needed by [ci: add test runner job #61](https://github.com/phoenix-rtos/phoenix-rtos-project/pull/61)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ TARGET ?= ia32-generic
 include ../phoenix-rtos-build/Makefile.common
 include ../phoenix-rtos-build/Makefile.$(TARGET_SUFF)
 
+INSTALL_PROGS :=
 
 .PHONY: clean
 clean:
@@ -32,7 +33,7 @@ ifneq ($(filter clean,$(MAKECMDGOALS)),)
 	$(shell rm -rf $(BUILD_DIR))
 endif
 
-T1 := $(filter-out clean all,$(MAKECMDGOALS))
+T1 := $(filter-out clean all install,$(MAKECMDGOALS))
 ifneq ($(T1),)
 	include $(T1)/Makefile
 .PHONY: $(T1)
@@ -46,9 +47,16 @@ else
 	include fs/Makefile
 	include disk/Makefile
 	#include virtio/Makefile
-	include sample/helloworld/Makefile
-	include sample/test/Makefile
-	include sample/test/fails/Makefile
+	include sample/Makefile
 	include setjmp/Makefile
 
 endif
+
+PREFIX_INSTALL = $(PREFIX_FS)bin/
+
+$(addprefix $(PREFIX_INSTALL), $(INSTALL_PROGS)): $(PREFIX_INSTALL)% : $(PREFIX_PROG_STRIPPED)%
+	$(INSTALL_FS)
+
+all: $(addprefix $(PREFIX_PROG), $(INSTALL_PROGS))
+
+install: $(addprefix $(PREFIX_INSTALL), $(INSTALL_PROGS))
--- a/sample/helloworld/test.yaml
+++ b/sample/helloworld/test.yaml
@@ -1,5 +1,5 @@
 test:
-    syspage: test-helloworld
     tests:
         - name: main
           harness: test.py
+          exec: test-helloworld
--- a/sample/helloworld/test.yaml
+++ b/sample/helloworld/test.yaml
@@ -1,5 +1,5 @@
 test:
-    syspage: helloworld
+    syspage: test-helloworld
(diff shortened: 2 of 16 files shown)


Makefile
```diff
@@ -46,9 +46,22 @@ else
...
+
+PROGS_INSTALL_FS := $(addprefix $(PREFIX_FS)bin/phoenix-rtos-tests/, $(TEST_HARNESS_PROGS) $(TEST_UNITY_PROGS))
+
+$(addprefix $(PREFIX_PROG), $(TEST_HARNESS_PROGS)): $(PREFIX_PROG)% : $(PREFIX_O)%.o
+	$(LINK)
+
+$(addprefix $(PREFIX_PROG), $(TEST_UNITY_PROGS)): $(PREFIX_PROG)% : $(PREFIX_O)%.o $(PREFIX_A)unity.a
+	$(LINK)
+
+$(PROGS_INSTALL_FS): $(PREFIX_FS)bin/phoenix-rtos-tests/% : $(PREFIX_PROG)%
```
nalajcie, line 62: - we should probably install stripped versions of the binaries in the filesystem
- I don't like the idea of having subdirs in `/bin`. I don't have any subdirs on my host computer there :no_good: 
- I think we either should install all tests in `/bin` or - If You really want to - we can make it `/test`
- we don't really need to be afraid of overwriting tests (they should have unique names), I think `make` might warn us if there are multiple rules for the same target (I don't know if it will work with the current rules) - You can easily test it locally :)

Makefile
```diff
@@ -46,9 +46,22 @@ else
...
+	$(INSTALL_FS)
+
+all: $(addprefix $(PREFIX_PROG), $(TEST_HARNESS_PROGS) $(TEST_UNITY_PROGS))
+
+install_fs: $(PROGS_INSTALL_FS)
```
nalajcie, line 67: maybe just `install`?

sample/helloworld/test.yaml
```diff
@@ -1,5 +1,5 @@
 test:
-    syspage: helloworld
+    syspage: main
```
nalajcie, line 2: maybe we could change it to `exec` or something like that, as we still overuse the `syspage` naming (see description of this PR :P)

trunner/device.py
```diff
@@ -6,6 +6,7 @@
 
 from .config import PHRTOS_PROJECT_DIR
 
+TIMEOUT = 8
```
nalajcie, line 9: add comment that it's default timeout for any input from the device (or even provide a better name)

nalajcie (commented):
We're going in good direction :+1:

trunner/device.py
jsarzy: I changed name and added keyword 'timeout' to YAML config

sample/helloworld/test.yaml
jsarzy: Done. I also removed some part of building code related to that "syspage" naming.

Makefile
jsarzy: Now tests are placed into `/bin` without subdirs. Also I changed a little bit building strategy. Let me know what you think.

trunner/testcase.py
```diff
@@ -176,15 +177,17 @@ def create(test):
...
-                    exec_bin=test['syspage'],
+                    timeout=test['timeout'],
+                    exec_bin=test['exec'],
                     status=status
             )
         if test['type'] == 'harness':
             return TestCaseCustomHarness(
                     name=test['name'],
                     target=test['target'],
+                    timeout=test['timeout'],
```
nalajcie, line 188: timeout is a good idea, but I would expect it's a timeout for the whole test, not for each expect() separately. I can't think of a better name straight away.

Btw, I don't know if we need full-test timeout for now, but maybe in the future we will. We may also add total test execution time in the test status output (right now in CI I don't know what took the most time).

nalajcie (approved):
Looks good to merge :+1: 
Pointed out some minor issues to be fixed with the next features :).

Regarding the building - seems sane enough, there are deeper changes to be made with the build system (eg. right now in concurrent build we can't be 100% sure unity will be fully built/installed before compiling the tests - but that needs to be solved globally in the build system).

Merging this now.

PR #6, opened 2021-04-14
tests: add a few tests of psh

jsarzy:
Some few examples how tests for psh can be made.

### DIFF
--- a/psh/test-mkdir.py
+++ b/psh/test-mkdir.py
@@ -1,0 +1,157 @@
+import random
+import string
+
+from collections import namedtuple
+
+File = namedtuple('File', ['name', 'owner', 'is_dir'])
+
+def ls(p, dir=''):
+    files = []
+
+    p.sendline(f'ls -la {dir}')
+    p.expect_exact('ls -la')
+    if dir:
+        p.expect_exact(f' {dir}')
+
+    p.expect_exact('\n')
+
+    while True:
+        idx = p.expect([r'\(psh\)\% ', r'(.*)(\r+)\n'])
+        if idx == 0:
+            break
+
+        line = p.match.group(0)
+
+        try:
+            permissions, _, owner, _, _, _, _, _, name = line.split()
+        except ValueError:
+            print(f'psh mkdir: wrong ls output')
+            return []
+
+        # Name is printed with ascii escaped characters - remove them
+        if name.startswith('\x1b[34m'):
+            name = name[5:]
+        if name.endswith('\x1b[0m'):
+            name = name[:-4]
+
+        f = File(name, owner, permissions[0] == 'd')
+        files.append(f)
+
+    return files
+
+def mkdir(p, dir):
+    p.sendline(f'mkdir {dir}')
+    p.expect_exact(f'mkdir {dir}')
+    p.expect_exact('\n')
+
+    idx = p.expect(['\(psh\)\% ', 'mkdir: failed to create (.*) directory'])
+
+    if idx == 1:
+        # Failed, read one more time (psh)%
+        p.expect_exact('(psh)% ')
+
+    return idx != 1
+
+def dir_present(dir, files):
+    for file in files:
+        if dir == file.name:
+            if file.owner != 'root':
+                print(f'psh ls: {dir} not owned by root')
+                return False
+            if not file.is_dir:
+                print(f'psh ls: {dir} is not directory')
+                return False
+
+            break
+    else:
+        print(f'psh ls: failed to find {dir}')
+        return False
+
+    return True
+
+def assert_dir_created(p, dir):
+    if not mkdir(p, dir):
+        print(f'psh mkdir: failed to create {dir}')
+        return False
+
+    path = dir.rsplit('/', 1)
+    if len(path) == 2:
+        dir = path[1]
+        path = path[0]
+    else:
+        path = '/'
+
+    files = ls(p, path)
+    if not files:
+        return False
+
+    return dir_present(dir, files)
+
+def random_dirs(p, pool, count=20):
+    if not mkdir(p, '/random_dirs'):
+        print(f'psh mkdir: failed to create /random_dirs')
+        return False
+
+    dirs = {''.join(random.choices(pool, k=random.randint(8, 16))) for _ in range(count)}
+    for dir in dirs:
+        if not mkdir(p, f'/random_dirs/{dir}'):
+            print(f'psh mkdir: failed to create {dir}')
+            return False
+
+    files = ls(p, '/random_dirs')
+    if not files:
+        return False
+
+    if len(files) > len(dirs) + 2:      # +2 because of . and ..
+        print(f'psh ls: too many dirs')
+        return False
+
+    for dir in dirs:
+        if not dir_present(dir, files):
+            return False
+
+    return True
+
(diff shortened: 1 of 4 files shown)


psh/test-mkdir.py
```diff
@@ -0,0 +1,143 @@
...
+
+    p.sendline(f'ls -la {dir}')
+    p.expect_exact('ls -la')
+    if dir:
+        p.expect_exact(f' {dir}')
+
+    p.expect_exact('\n')
+
+    while True:
+        idx = p.expect(['\(psh\)\% ', '(.*)(\r+)\n'])
```
kemonats, line 19:
```suggestion
        idx = p.expect([r'\(psh\)\% ', '(.*)(\r+)\n'])
```
I think it should be `raw string`.

psh/test-mkdir.py
```diff
@@ -0,0 +1,143 @@
...
+        if idx == 0:
+            break
+
+        line = p.match.group(0)
+
+        permissions, _, owner, _, _, _, _, _, name = line.split()
```
kemonats, line 25: Maybe this line in the block `try ... except`

psh/test-mkdir.py
```diff
@@ -0,0 +1,143 @@
...
+        # Line is printed with ascii escaped characters - remove them
```
kemonats, line 26: Not line, only name.

psh/test-mkdir.py
```diff
@@ -0,0 +1,143 @@
...
+    return idx != 1
+
+def dir_present(dir, files):
+    for file in files:
+        if dir == file.name:
+            if file.owner != 'root':
+                print(f'psh ls: {dir} not owned by root')
+                return False
+            if not file.is_dir:
+                print(f'psh ls: {dir} is not dictionary')
```
kemonats, line 57:
```suggestion
                print(f'psh ls: {dir} is not directory')
```

psh/test-mkdir.py
```diff
@@ -0,0 +1,143 @@
...
+                return False
+
+            break
+    else:
+            print(f'psh ls: failed to found {dir}')
+            return False
```
kemonats, line 63:
```suggestion
        print(f'psh ls: failed to found {dir}')
        return False
```
Indentation

psh/test-mkdir.py
```diff
@@ -0,0 +1,143 @@
...
+    return dir_present(dir, files)
+
+def random_dirs(p, pool, count=20):
+    if not mkdir(p, '/random_dirs'):
+        print(f'psh mkdir: failed to create /random_dirs')
+        return False
+
+    dirs = {''.join(random.choices(pool, k=random.randint(8, 16))) for _ in range(count)}
+    for dir in dirs:
+        if not mkdir(p, '/random_dirs/' + dir):
```
kemonats, line 90:
```suggestion
        if not mkdir(p, '/random_dirs/{dir}'):
```
f-string

psh/test-mkdir.py
```diff
@@ -0,0 +1,143 @@
...
+
+    return True
+
+
+def get_first_prompt(p):
+    prompt = '\r\x1b[0J' + '(psh)% '
+
+    got = p.read(len(prompt))
+    if got != prompt:
+        print(f'Expected:\n\t{prompt}\nGot:\n\t{got}')
```
kemonats, line 112:
```suggestion
        print(f'Expected:\n{prompt}\nGot:\n{got}')
```
The char `\t` is not needed, in `prompt` is char '\r'.

nalajcie:
@Maxez I heard You've tried to run these tests, please report here if they are working as expected ;).

psh/test-mkdir.py
```diff
@@ -0,0 +1,143 @@
...
+            print(f'psh ls: failed to found {dir}')
+            return False
+
+    return True
+
+def assert_dir_created(p, dir):
+    if not mkdir(p, dir):
+        print(f'psh mkdir: failed to create {dir}')
+        return False
+
```
nalajcie, line 71: maybe also check if `..` works? (something like: `files = ls(p, dir + "/..")`)

psh/test-mkdir.py
```diff
@@ -0,0 +1,143 @@
...
+    dirs = {''.join(random.choices(pool, k=random.randint(8, 16))) for _ in range(count)}
+    for dir in dirs:
+        if not mkdir(p, '/random_dirs/' + dir):
+            print(f'psh mkdir: failed to create {dir}')
+            return False
+
+    files = ls(p, '/random_dirs')
+
+    if len(files) > len(dirs) + 2:      # +2 because of . and ..
```
nalajcie, line 96: hmm... this might fail if we would be running the same test 2 times on the same image (where `/random_dirs` will be on persistent storage (like ext2 partition)). Maybe we should remove the dir at the beginning of the test?

psh/test.yaml
```diff
@@ -0,0 +1,11 @@
+test:
+    syspage: psh
```
nalajcie, line 2: please update the YAML (use `exec`), so we can merge it

nalajcie (commented):
Looks good, please fix all small issues found so we can merge this :+1:

psh/test.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+    tests:
+        - name: prompt
+          harness: test-prompt.py
+        - name: top
```
Maxez, line 7:
```suggestion
        - name: ps
```
I'd change test name to ps.

Maxez:
> @Maxez I heard You've tried to run these tests, please report here if they are working as expected ;).

The tests appear to work correctly:
psh-tests

psh/test-mkdir.py
Maxez: I'd suggest removing test folders **after** we verify they were successfully created by mkdir command. We should cleanup the mess we create.

psh/test-mkdir.py
jsarzy: Unfortunately, there is no such param.

psh/test-mkdir.py
jsarzy: I added such a test

psh/test-mkdir.py
jsarzy: There is no `rm` command in psh.

psh/test-mkdir.py
jsarzy: Exception will be raised. It will be changed in the next PR with assertion handling.

psh/test-mkdir.py
nalajcie: ok, maybe it would be a good feature enhancement for `psh`

psh/test-mkdir.py
nalajcie: @Maxez This was my initial idea, but currently we don't have "cleanup" hook (some code which would be called regardless of the test outcome) - currently if the test would fail or raise an exception we would need complicated code constructs to always cleanup. Because of that I proposed to "clean at the beginning" for this particular test :).

This is not a merge blocker as potential FS changes during testing and how to address them is a subject for broader discussion ;).

nalajcie (approved):
LGTM, can I merge this now?

jsarzy:
Yeap!

PR #9, opened 2021-04-19
runner: add 'host-pc' target

jsarzy:
These changes allows to run tests on the host-pc target. To do this I added keyword `target` to config.

Runner choose target following these rules:
1. If there is a keyword `target` in config then use it.
2. If target is no specified in config then run test on all available targets excluding `host-pc`. Available targets are defined in `config.py`.

If I understand correctly most of the tests that will be run on `host-pc` are created only for that platform. Therefore, we don't want to include `host-pc` in all default targets in order to avoid writing some `exclude` keyword in most of the configs:
```YAML
tests:
     test:
        name: test-for-all-platform
        exec: psh
        exclude: host-pc
```
On the other hand, if we would like to run test on `host-pc` and other available platforms then instead of writing explicitly all targets:
```YAML
tests:
     test:
        name: test-for-all-platform-and-host-pc
        exec: helloworld
        target:
            - host-pc
            - ia32-generic
            # If there is new platfrom we have to update all configs because of excluding host-pc from all targets
```
we can define configs like this:
```YAML
tests:
     test:
        name: test-for-all-platform-and-host-pc
        exec: helloworld
        target:
            - host-pc
    test:
        name: test-for-all-platform-and-host-pc
        exec: helloworld
```
I also removed some unused Makefiles and tried to organize what should be build on specified platform but I will appreciate any feedback how it can be done better.


nalajcie:
Just putting the comment regarding the new YAML syntax - I would go for more explicit approach (similar to Github Actions or Homeassistant):

explicitly stating targets:
```yaml
  target:
    targets: [host-pc, ia32-generic]   # can also be written multiline
```


adding or removing from default targets:
```yaml
  target:
    # implicit "targets" -> array of default targets
    include:
      - host-pc
    exclude:
      - riscv64-virt
```

Probably it would be beneficial to support simple wildcards in the future so I could write:
```yaml
# eg. tests for JFFS filesystem which is unavailable on "small ARM" targets
  target:
    exclude:
      - armv7m.* # or "armv7m*"  - some simplified regexes?
```
(just an idea for the future, not really right away, so might not be addressed in this PR)

Similarity with Github Actions (matrix stategy): https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations

Similarity with some of the HomeAssistant modules: https://www.home-assistant.io/integrations/recorder/#configure-filter

I will review the code later (hopefully today :))

jsarzy:
I like the idea of include/exclude keywords. I also added a few tests of handling keyword 'target'.

I only have a little problem with combining "global" and "local" config for tests. As I assume if we have config defined like this:
```YAML
test:
    target:
        targets: [ia32-generic, riscv64-virt]
        include: [armv7a7-imx6ull]

    tests:
        - name: test1
          exec: test1.elf
          target:
              include: [host-pc]

        - name: test2
          exec: test2.elf
          target:
              targets: [ia32-generic]
```

Then `test1` should be executed on the targets `[ia32-generic, riscv64-virt, armv7a7-imx6ull, host-pc]` and `test2` only on the target `[ia32-generic]`? Or maybe `test1` should be handled on `[all available targets] + [host-pc]`?

nalajcie:
> Then `test1` should be executed on the targets `[ia32-generic, riscv64-virt, armv7a7-imx6ull, host-pc]` and `test2` only on the target `[ia32-generic]`? Or maybe `test1` should be handled on `[all available targets] + [host-pc]`?

I think currently all *global* configs are **inherited** by the *local* configs, so I would assume the `target` would work the same - *local* inclusions/exclusions would be against the *global* configuration (not the *default* one). 

I think it's uncommon scenario that one sub-test would have completely different set of targets, but than we can either:
 - use separate YAML file for that test (so no problem at all)
 - introduce special keyword `default` that writing `targets: default` will reset the targets to the hard-coded default, so You can use then `include`/`exclude` against the default targets set

Makefile
```diff
@@ -40,16 +40,12 @@ ifneq ($(T1),)
 $(T1):
```
nalajcie, line 40: as a part of generic Makeflies cleanup:
 - You can remove all these `TARGET ?= ` a the beginning of the file
 - `$(T1)` rule can be simplified: https://github.com/phoenix-rtos/phoenix-rtos-devices/commit/2bb99186b080d99f5b8dfd6571d0a4d499e96846#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

Makefile
```diff
@@ -40,16 +40,12 @@ ifneq ($(T1),)
...
-	include stdio/Makefile
-	include proc/Makefile
-	include mem/Makefile
-	include fs/Makefile
-	include disk/Makefile
-	#include virtio/Makefile
-	include sample/Makefile
-	include setjmp/Makefile
 
+ifneq ($(TARGET_SUFF),host)
```
nalajcie, line 44: maybe instead of making special conditionals here we should add `include Makefile.common` to `Makefile.arm` (and others)?

trunner/config.py
```diff
@@ -12,6 +12,9 @@
 # Default time after pexpect will raise TIEMOUT exception if nothing matches an expected pattern
 PYEXPECT_TIMEOUT = 8
 
+# Available targets for test runner. By default test are run on every target except host-pc
+TARGETS = ['ia32-generic', 'host-pc']
```
nalajcie, line 16: maybe we should have `ALL_TARGETS` and `DEFAULT_TARGETS` instead of some logic while parsing configs?

trunner/config.py
```diff
@@ -56,30 +60,74 @@ def parse_test_case(self, config):
...
+        if not config['harness'].exists():
+            raise YAMLParserError(self, f"harness {config['harness']} file not found")
+
+        if not str(config['harness']).endswith('.py'):
+            raise YAMLParserError(self, "harness file must has .py extension")
+
+    def parse_target(self, config):
+        config.setdefault('target', dict())
+
+        target = config['target']
```
nalajcie, line 82:
```python
target = config.get('target', dict())
```

trunner/config.py
```diff
@@ -56,30 +60,74 @@ def parse_test_case(self, config):
...
 
-            if not config['harness'].exists():
-                raise YAMLParserError(self, f"harness {config['harness']} file not found")
+        return test_case
 
-            if not str(config['harness']).endswith('.py'):
-                raise YAMLParserError(self, "harness file must has .py extension")
+    def copy_global_keywords(self, glob, local):
+        if 'targets' in local:
+            # Always take local 'targets' without copying include/exclude from global config
```
nalajcie, line 123: as per my previous comment, I think selectively not-inheriting `target` property might be counter-intuitive

trunner/config.py
```diff
@@ -56,30 +60,74 @@ def parse_test_case(self, config):
...
+            test_case.append(test)
 
-            if not config['harness'].exists():
-                raise YAMLParserError(self, f"harness {config['harness']} file not found")
+        return test_case
 
-            if not str(config['harness']).endswith('.py'):
-                raise YAMLParserError(self, "harness file must has .py extension")
+    def copy_global_keywords(self, glob, local):
+        if 'targets' in local:
```
nalajcie, line 122: shouldn't it be `target`? ...
now that I see it - the `target` and `targets` are too similar, maybe we should rename `targets` so it won't be confusing. `value`? `array`?

trunner/test/test_config.py
```diff
@@ -0,0 +1,187 @@
+import pytest
+import yaml
+
+from trunner.config import YAMLParser, YAMLParserError, TARGETS
+
+
+@pytest.fixture
+def parser(tmp_path, config, targets, request):
```
nalajcie, line 8: oh, tests for tests runner! I :heart: the inception! :+1:

trunner/test/test_config.py
```diff
@@ -0,0 +1,187 @@
...
+    return YAMLParser(path, targets)
+
+
+class TestTarget:
+    @pytest.fixture
+    def targets(self):
+        return TARGETS
+
+    @pytest.mark.parametrize('config, expect_target', [
+        (
```
nalajcie, line 23: might have been easier/more readable to store raw yaml exernally, but I'm not against the current solution either.

nalajcie (approved):
Good direction, I left some minor remarks but they are not a merge blocker. Let me know if You want to merge this now or experiment with it more.

jsarzy:
I will add few changes (I hope today), so do not merge yet.

trunner/config.py
jsarzy: I used `value` keyword. In parser I used word `array` referring to keyword of type `{value: something, include: something, exclude: something}`. Also let me know if I understood inheritance correctly.

jsarzy:
In this push I also changed a little bit YAMLParser to make it easier to use in tests. In few cases I changed variable name `config` to `test`. So with variable `config` I mean the whole config loaded from .yaml (specially global options) and by word `test` the description of the test (from `tests` list). Sorry for the mid-review changes but I think parser will be more readable.

trunner/config.py
```diff
@@ -30,73 +37,135 @@ def __init__(self, parser, message="Parser error"):
...
+            return
+
+        array_keywords = ('value', 'include', 'exclude')
+        global_array = self.config[keyword]
+        array = test[keyword]
+
+        for key in array_keywords:
+            inherited = global_array.get(key, [])
+            if inherited:
+                array.setdefault(key, []).extend(inherited)
```
nalajcie, line 147: I would go for computing global array and setting it for test->array->values 

For me intuitive would be to:

```yaml
test:
  target:
    value: [a, b]
    include: [c]
  tests:
    - name: test1  # expected to be run on d
      target:
        value: [d]
    - name: test2  # expected to be run on a, b, c, e
      target:
        include: [e]
    - name: test3  # expected to be run on a, b
       target:
        exclude: [c]

```
If I read correctly the current code:
```yaml
test1 # run on a, b, c, d
test2 # run on a, b, c, e
test3 # run on a, b (but only because of implementation detail)
```

FYI, first I've used `targets` keyword instead of `target` as it was better mnemonic for "array of targets". Fixed this only after re-reading the source code. Maybe `targets` would be better?

nalajcie (commented):
one remark regarding the targets inheritance, other than that - LGTM.

trunner/config.py
jsarzy: Now targets should inherit in more intuitive way. I'm a little bit dissatisfied with `YAMLParser` as a whole -- is hard to test it in more "modular" way, but it should be a topic for another PR.

trunner/config.py
```diff
@@ -30,73 +37,134 @@ def __init__(self, parser, message="Parser error"):
...
-                    test_config[key] = config[key]
+        if 'targets' in self.config:
+            self.parse_target(self.config)
 
-            test = self.parse_test_case(test_config)
-            self.parsed_tests.append(test)
+        tests_configs = self.config.pop('tests')
+        for test_config in tests_configs:
+            # Combine 'local' test config and 'global' config
+            self.inherit_global_keywords(test_config)
```
nalajcie, line 166: small note - the naming `inherit_global_keywords` might be misleading, because it's not `self` that is inheriting some properties from the `test_config` - it's the other way around. So maybe it should be `propagate` or something like that.

Might be changed in the future, let's leave it as-is for now.

nalajcie (approved):
LGTM.

Issue #10, opened 2021-04-22
fix tests compilation warnings for host-pc target


nalajcie:
```
Warning: sample/test/test-echo.c:11:3: warning: ignoring return value of 'read', declared with attribute warn_unused_result [-Wunused-result]
   11 |   read(STDIN_FILENO, &c, 1);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~
```

https://github.com/phoenix-rtos/phoenix-rtos-project/pull/63/checks?check_run_id=2408557131#step:4:86

Issue #7, opened 2021-04-14
Unity throws warnings when being built


nalajcie:
Sample build log:

https://github.com/phoenix-rtos/phoenix-rtos-build/pull/45/checks?check_run_id=2332083723#step:5:307

PR #11, opened 2021-04-27
Add CI for runner. Fix warnings from GCC and flake

jsarzy:
Comments to some other warnings:
1. Build on riscv64-virt produces the following warning:
```
unity/unity_internals.h:903:121: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
[omitted output]
sample/test/test-unity.c:26:2: note: in expansion of macro 'TEST_ASSERT_EQUAL_PTR'
   26 |  TEST_ASSERT_EQUAL_PTR(p, NULL);
      |  ^~~~~~~~~~~~~~~~~~~~~
```
Unity uses `UINTPTR_MAX` to detect pointer size. Then it casts it to int of the right size. In libphoenix `UINTPTR_MAX` is defined as `UINT32_MAX` but `sizeof(void*)` is 8. Also `uintptr_t` is defined as `uint64_t` so it might some inconsistency.  

2. No return warning mentioned in https://github.com/phoenix-rtos/phoenix-rtos-tests/issues/10
It might be triggered because of lack `noreturn` attribute in `longjmp` definition in libphoenix.

Changes fixe https://github.com/phoenix-rtos/phoenix-rtos-tests/issues/7.

### DIFF
--- a/.github/workflows/runner-ci.yaml
+++ b/.github/workflows/runner-ci.yaml
@@ -1,0 +1,44 @@
+name: Runner CI
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - '**.py'
+      - 'trunner/**'
+  pull_request:
+    branches:
+      - master
+    paths:
+      - '**.py'
+      - 'trunner/**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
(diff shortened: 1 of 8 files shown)


nalajcie:
> Unity uses `UINTPTR_MAX` to detect pointer size. Then it casts it to int of the right size. In libphoenix `UINTPTR_MAX` is defined as `UINT32_MAX` but `sizeof(void*)` is 8. Also `uintptr_t` is defined as `uint64_t` so it might some inconsistency.


Should probably be fixed in `libphoenix`, can You prepare PR?
 
> 1. No return warning mentioned in #10
>    It might be triggered because of lack `noreturn` attribute in `longjmp` definition in libphoenix.

That's because our `libphoenix` includes lack function attributes enabling additional compiler checks (`__attribute_warn_unused_result__` in this particular case). I believe they should be introduced in the future as a part of making `libphoenix` more user friendly.

Because of the above building for `host-pc` enables us to find possibly-bad code quicker.

Regarding python CI - i'm all in, personally I'm using `pylint` but I assume `flake8` is as robust as the other tool :).

nalajcie (approved):
LGTM

PR #14, opened 2021-04-28
Add assertion handling

jsarzy:
Now we do not have to return error value from harnesses. Test interrupted by assertion produces following exception message:
image

Test interrupted by python exception other than assertion/pyexpect are handled and marked as failed.

### DIFF
--- a/initfini/test.py
+++ b/initfini/test.py
@@ -1,17 +1,12 @@
 def harness(p):
-    if 'Constructor 0' not in p.readline():
-        return False
+    scenario = [
+        'Constructor 0',
+        'Constructor 1',
+        'Main function',
+        'Destructor 1',
+        'Destructor 0'
+    ]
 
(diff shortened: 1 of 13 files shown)


nalajcie (approved):
Overall looks good, I'm wondering if we shouldn't distinguish between test execution problems and test data problem (different exception types) - if test executes but fails You know it's the problem with the code being tested - but if there is a problem with test execution it might be anything (problems with test runner, problems with infrastructure, etc). 

Currently there is only a minor chance of exception in the runner, but when we will move to testing on HW it might happen more often. I would appreciate knowing the probable source of failed test without digging into the traceback :).

LGTM.

jsarzy:
I agree we can distinguish exceptions caused by internal runner problems (like failed spawn command) and failed test execution (caused by assert or pyexpect timeout). I will add it with hardware handling part.

PR #15, opened 2021-04-29
runner: add tests for pyexpect exceptions handling

jsarzy:
Additional tests to cover all exception types that can be raised during execution of harness.


nalajcie (approved):
LGTM.

PR #18, opened 2021-05-07
meterfs: added test case for multiple lookup calling

Diabliz:
Test case for lookup bug. Bug resolve in PR [#33](https://github.com/phoenix-rtos/phoenix-rtos-filesystems/pull/33).


anglov (approved):
Ok test for checking bug behavior.

PR #19, opened 2021-05-07
runner: add flashing imxrt106x target

jsarzy:


trunner/device.py
```diff
@@ -44,6 +64,110 @@ def run(self, test):
...
+
+    def high(self):
+        RPi.GPIO.output(self.pin, RPi.GPIO.HIGH)
+
+    def low(self):
+        RPi.GPIO.output(self.pin, RPi.GPIO.LOW)
+
+
+class IMXRT106xRunner(DeviceRunner):
+    SDP = 'plo-ram-armv7m7-imxrt106x.sdp'
```
nalajcie, line 83: please add docstring regarding how RT106x should be connected to be useable with this class.

Should we make some runner for testing board connected directly to the PC?

trunner/device.py
```diff
@@ -44,6 +64,110 @@ def run(self, test):
...
+        else:
+            self.boot_gpio.high()
+
+        self.reset()
+
+    def flash(self):
+        self.serial_downloader = True
+        self.boot()
+
+        logging.info("psu run!\n")
```
nalajcie, line 113: I think `\n` is not be needed here (if so - applies also in other places)

trunner/device.py
```diff
@@ -44,6 +64,110 @@ def run(self, test):
...
+
+        self.boot_gpio = GPIO(4)
+
+    def reset(self):
+        self.reset_gpio.low()
+        time.sleep(0.050)
+        self.reset_gpio.high()
+
+    def boot(self):
+        if self.serial_downloader:
```
nalajcie, line 102: I don't see any real benefit of keeping the current state as an instance attribute, maybe just
```python
def boot(self, serial_downloader=False):
```
would be enough?

trunner/device.py
nalajcie, line 113: I'm wondering about detecting if we're running inside github actions and styling the output with specific "tags" - I'm thinking about `::group::` to create collapsible debug logs (https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines) - like the current use case when we're flushing the device and are interested in the logs only if it failed.

Just an idea to experiment with in the future.

trunner/device.py
```diff
@@ -1,12 +1,23 @@
...
 import signal
+import subprocess
+import time
 
 import pexpect
 import pexpect.fdpexpect
 import serial
 
+try:
+    import RPi.GPIO
```
nalajcie, line 12: should this be added to requirements.txt?

trunner/device.py
```diff
@@ -44,6 +64,110 @@ def run(self, test):
...
+
+    def boot(self):
+        if self.serial_downloader:
+            self.boot_gpio.low()
+        else:
+            self.boot_gpio.high()
+
+        self.reset()
+
+    def flash(self):
```
nalajcie, line 109: I would add some of these functions to the base class `DeviceRunner` and make them no-op in Qemu runners

nalajcie (commented):
Looks good, I've added some minor remarks.

Do You want to merge it or You're planning to add more commits to this PR?

trunner/device.py
jsarzy: Done

trunner/device.py
jsarzy: By default logging adds new line but I changed it. Logging doesn't have keyword similar to python `end` in `print` so I couldn't log info about starting test without new line.

jsarzy: Wow, I didn't know we can style output in github actions. I noted it for the future.

trunner/device.py
jsarzy: I added class `Runner`.

trunner/device.py
jsarzy: I added docstrings.

On PC I don't know how to solve problem with restarting board. Should we print message "Now you have 10 seconds to restart board manually"?

nalajcie: I don't know how it should be done, I'm just stating the problem (and trigger discussion whether we would like to do it now - maybe we should ask if/how it might help the people working with rt1064 right now?).

Possible solutions I can think of:
 - "interactive mode" ("Restart board into serial-downloader and press ENTER" && wait)
 - (in the future) configurable restart method (eg. by JLink, by `psh reboot`, etc.)?

jsarzy: Yeap, I was thinking about "interactive mode". In the next PR I will add building for rt1064 and then I will be thinking about handling rt1064 connected directly to the PC - is hard to write tests when we can't run them locally;).

trunner/device.py
nalajcie: yeah, we can also emit warnings, errors and other things :)

trunner/device.py
```diff
@@ -19,7 +30,27 @@
...
+    while True:
+        output = proc.stdout.readline().decode('utf-8')
+        if proc.poll() is not None and output == '':
+            break
+        if output:
+            logging.info(output)
+
+
+class Runner:
+    def boot(self):
```
nalajcie, line 43: - please add one-sentence docstring regarding what the given function should do (so: describe the interface)
- i'm not sure if `boot` is not rt1064-specific, maybe `reboot` would be more desirable in different Runners (like Qemu)?

trunner/device.py
nalajcie: Ok, still it would be good to gather feedback from other people working on rt1064 ;).

nalajcie (commented):
I think the code is good for now, more generalization will come with adding new targets.

Please add some :memo: docstrings and I think we can merge it.

trunner/device.py
jsarzy: 1. I added docstrings
2. I removed `boot` from `Runner`. Maybe I'll put it into the `DeviceRunner` class in the future when more targets arrives.

nalajcie:
docstring should be put as a first line inside the function - https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

nalajcie:
note for the future: it should be possible to use some `flake8` action supporting problem-matcher to show the problems inline.

runner.py
```diff
@@ -64,6 +64,11 @@ def parse_args():
                         choices=logging_level,
                         help="Specify verbosity level. By default uses level info.")
 
+    parser.add_argument("-s", "--serial",
+                        default=config.DEVICE_SERIAL,
+                        help="Specify serial to communicate with device board. "
+                             "By default uses /dev/ttyACM0.")
```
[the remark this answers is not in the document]
jsarzy: Done

nalajcie (approved):

PR #20, opened 2021-05-14
makes: use templates for binaries and static libs compilation

nalajcie:
Rewrote Makefiles according with the new convention

JIRA: CI-45

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,43 +1,37 @@
 #
-# Makefile for phoenix-rtos-devices
+# Makefile for phoenix-rtos-tests
 #
 # Copyright 2018, 2019 Phoenix Systems
 #
 # %LICENSE%
 #
 
-SIL ?= @
-MAKEFLAGS += --no-print-directory
-
-TARGET ?= ia32-generic
-
 include ../phoenix-rtos-build/Makefile.common
+# FIXME: this include should be done by Makefile.common
 include ../phoenix-rtos-build/Makefile.$(TARGET_SUFF)
 
-INSTALL_PROGS :=
+.DEFAULT_GOAL := all
 
-.PHONY: clean
-clean:
-	@echo "rm -rf $(BUILD_DIR)"
+# shortcut for unity dependence
+define add_unity_test
+$(call add_test,$(1),$(2),unity)
+endef
 
-ifneq ($(filter clean,$(MAKECMDGOALS)),)
-	$(shell rm -rf $(BUILD_DIR))
-endif
+# default path for the programs to be installed in rootfs
+DEFAULT_INSTALL_PATH := /bin
 
-T1 := $(filter-out clean all install,$(MAKECMDGOALS))
-ifneq ($(T1),)
-	include $(T1)/Makefile
-.PHONY: $(T1)
-$(T1): all
-else
-	include Makefile.$(TARGET_SUFF)
-endif
+# read out all components
+ALL_MAKES := $(shell find -mindepth 2 -name Makefile)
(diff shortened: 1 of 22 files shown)


Makefile
```diff
@@ -1,43 +1,41 @@
...
 
-$(addprefix $(PREFIX_INSTALL), $(INSTALL_PROGS)): $(PREFIX_INSTALL)% : $(PREFIX_PROG_STRIPPED)%
-	$(INSTALL_FS)
+# read out all components
+ALL_MAKES := $(shell find -mindepth 2 -name Makefile)
+include $(ALL_MAKES)
 
-all: $(addprefix $(PREFIX_PROG), $(INSTALL_PROGS))
+# by default compile all tests, but allow custom values on per-TARGET_FAMILY basys
+# TODO: prepare tool which will read YAML files and return the components which should be compiled for a given platform
```
nalajcie, line 33: @jsarzy - idea for the future: 
write separate tool (which uses some classes from the runner) to compute the subset of components / tests to be run on a given target and use it here instead of keeping the per-target configuration on two places (YAML files and Makefiles).

Makefile
```diff
@@ -1,43 +1,42 @@
...
-ifneq ($(T1),)
-	include $(T1)/Makefile
-.PHONY: $(T1)
-$(T1): all
-else
-	include Makefile.$(TARGET_SUFF)
+ifneq ($(filter %clean,$(MAKECMDGOALS)),)
+$(info cleaning targets, make parallelism disabled)
+.NOTPARALLEL:
 endif
```
kemonats, line 18: What do you think about moving this to Makefile.common?

meterfs/Makefile
```diff
@@ -6,20 +6,26 @@
...
+  LOCAL_SRCS += file_phx.c
 endif
 
-OBJS_COMMON := $(patsubst %, $(PREFIX_O)meterfs/%.o, $(OBJS_COMMON))
-LIBS := $(patsubst %, $(PREFIX_A)%.a, $(LIBS))
+include $(static-lib.mk)
 
-$(PREFIX_PROG)%: $(PREFIX_O)meterfs/%.o $(OBJS_COMMON) $(LIBS)
-	$(LINK)
+define add_metersrv_test
```
anglov, line 24: You mean meterfs I think :)

sample/test/fails/Makefile
```diff
@@ -1,13 +1,8 @@
-HARNESS_PROGS := test-fail-output-1 test-fail-output-2 test-fail-exception
+$(eval $(call add_test, test-fail-output-1))
+$(eval $(call add_test, test-fail-output-2))
+$(eval $(call add_test, test-fail-exception))
 
-UNITY_PROGS := test-fail-unity test-fail-unity-exc
+$(eval $(call add_unity_test, test-fail-unity, unity))
+$(eval $(call add_unity_test, test-fail-unity-exc, unity))
```
jsarzy, line 6: Redundant `unity`

jsarzy (approved):
LGTM

Makefile
nalajcie: yeah, good idea, I'll add it in phoenix-rtos-build and will be removing from other Makefiles with other updates.

meterfs/Makefile
nalajcie: thx, fixed

sample/test/fails/Makefile
nalajcie: thx, fixed

PR #24, opened 2021-05-28
runner: imxrt106x: add loading tests using plo

jsarzy:
These changes add possibility of running tests automatically. Runner uses new plo to load test ELF (from _boot/test) to the syspage.

Some remarks:
1.  usb0 is not ready when the plo prompt is printed first time. Normally in the initscript there is some wait (like wait 2000), so end user is not affected by this. Unfortunately, plo occasionally crashes executing this command. Therefore, for now, tests will be build without wait in the preinit script and runner will wait 1 second for usb initialization (but maybe is possible to make some lookup for /dev).
2. `phoenixd` needs ~1 second to be ready to handle requests from plo. Here we could parse output from phoenixd and look for the "dispatcher ready" string but `phoenixd` doesn't flush too often. Setting small buffers in `Popen` doesn't help. I would like to handle this in the next PR (and also how to log `phoenixd` output in the GA and to the user in case of failure -- in general logging).
3. Im going to think of how to handle the `psh` case because we can't copy it directly to the syspage.

### DIFF
--- a/trunner/device.py
+++ b/trunner/device.py
@@ -2,6 +2,8 @@ import logging
 import os
 import signal
 import subprocess
+import sys
+import threading
 import time
 
 import pexpect
@@ -11,9 +13,10 @@ import serial
 try:
     import RPi.GPIO
 except ImportError:
-    print("RPi.GPIO module can't be imported!")
+    pass
 
 from .config import PHRTOS_PROJECT_DIR, DEVICE_SERIAL
+from .tools.color import Color
 
 
 _BOOT_DIR = PHRTOS_PROJECT_DIR / '_boot'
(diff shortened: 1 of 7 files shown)


jsarzy:
I converted PR to draft - I'm going to add some changes to phoenixd handling in **this** PR. Sorry for the confusion

jsarzy:
I added `phoenixd` output handling. There aren't any sleep in code. I also changed directory name which from test are loaded to `_boot/test/{target}`.

nalajcie:
@xvuko I'll start adding You to PRs with python code, so You can share Your knowledge regarding modern Python

trunner/device.py
```diff
@@ -101,7 +294,7 @@ class IMXRT106xRunner(DeviceRunner):
     SDP = 'plo-ram-armv7m7-imxrt106x.sdp'
     IMAGE = 'phoenix-armv7m7-imxrt106x.disk'
 
-    def __init__(self, port, phoenixd_port='/dev/ttyUSB0'):
+    def __init__(self, port, phoenixd_port='/dev/ttyACM1'):
```
nalajcie, line 297: probably we should also think about some "device configs" in the future, but it's good for now.

One tip: it might be easier to use `/dev/serial/by-path/*` or `/dev/serial/by-name/*` to always use the desired peripheral regardless of connecting order.

trunner/device.py
```diff
@@ -30,13 +33,203 @@
...
+        return output
+
+    def kill(self):
+        os.killpg(os.getpgid(self.proc.pid), signal.SIGTERM)
+        self.reader_thread.join()
+        if self.reader_lock.locked():
+            self.reader_lock.release()
+
+    def __enter__(self):
+        self.run()
```
xvuko, line 150: If exception is raised inside run function after invoking spawn than kill method will not be called (exception in \__enter\__  does not result in \__exit\__ call).

nalajcie (approved):
overall LGTM, I'm eager to see it in action. We will hammer out other potential problems "in production" :).

Let's wait for @xvuko  review as he could have some valid points regarding python syntax.

trunner/device.py
```diff
@@ -30,13 +33,203 @@
...
+            raise
+
+        self.plo = pexpect.fdpexpect.fdspawn(self.serial, timeout=8)
+        return self
+
+    def close(self):
+        self.serial.close()
+
+    def __enter__(self):
+        return self.open()
```
xvuko, line 191: See comment about Phoenixd.\__enter\__

trunner/device.py
```diff
@@ -30,13 +33,203 @@
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.DEVNULL,
+            cwd=self.cwd
+        )
+
+        self.read_output()
+        self.proc.wait()
+        if self.proc.returncode != 0:
+            logging.error(f'Command {" ".join(self.proc.args)} with pid {self.proc.pid} failed!\n')
+            raise Exception('Flashing IMXRT106x failed\n')
```
xvuko, line 74: Trailing newline in exception string

trunner/device.py
```diff
@@ -30,13 +33,203 @@
...
+    """Wrapper for psu program"""
+
+    def __init__(self, script, cwd=_BOOT_DIR):
+        self.script = script
+        self.cwd = cwd
+        self.proc = None
+
+    def read_output(self):
+        if is_github_actions():
+            logging.info('::group::psu\n')
```
xvuko, line 50: Unless intention is to output empty line trailing newlines should be removed from logging functions. There are more cases in this commit.

trunner/device.py
jsarzy: Changed to `/dev/serial/by-id/*`

trunner/device.py
```diff
@@ -30,13 +33,203 @@
...
+    def output(self):
+        output = self.output_buffer
+        if is_github_actions():
+            output = '::group::Pheonixd\n' + output + '::endgroup::\n'
+
+        return output
+
+    def kill(self):
+        os.killpg(os.getpgid(self.proc.pid), signal.SIGTERM)
+        self.reader_thread.join()
```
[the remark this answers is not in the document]
jsarzy: I added timeout to join, then SIGKILL is send.

trunner/device.py
jsarzy: Do you suggest that I should put threading code into `try: except:`? I also changed a little bit `Lock` handling.

trunner/device.py
jsarzy: I added catching exception from `pexpect.fdspawn` and closing serial port if exception occures.

trunner/device.py
jsarzy: Fixed.

nalajcie:
please also fix codespell errors: https://github.com/phoenix-rtos/phoenix-rtos-tests/compare/d44fc52aa457aff775e269cab0ae7e10e7983db6..f18c71b1184aee645ea96d929c027e88772be923#diff-ca95dae7b90239ed0a4ca2685d5c104a1ce5248e9cc7fa145ee7472eb6a0f9d5R103

trunner/device.py
jsarzy: Is a more complex problem right now. Logging doesn't provide `end` arguments as print, so I set logging to not use a new line by default. I'm going to change it and provide more coherent way of logging. So I suggest leaving it as it is right now and fix it for the whole runner in a separate PR.

jsarzy:
I smuggled some few small changes:
1. `Phoenixd.run` throws exception if message dispatcher don't start.
2. `Phoenixd.run` use `threading.Event` instead of `threading.Lock` - I think the intention is more clear.
3. I added `PloTalker.from_pexpect` class method.
4. `go!` comamnd is send after second opening of serial (in `DeviceRunner.run`), so we are sure that we will read the whole output.

nalajcie (approved):
From my point of view - LGTM (as I'm eager to start running tests on HW ;))

@xvuko do You see any other issues which should to be resolved in this PR?

trunner/device.py
xvuko: I was mostly concerned about timeout exception from Lock.acquire. This is not actually a case as acquire simply returns false on timeout. 

It is fine now - can be merged as is.

### for future reference: 

Generic solution to deal with this case would be to use:
```python
def __enter__(self):
    try:
        self.run()
    except Exception:
        self.__exit__(None, None, None) # if exit uses args than None should be changed to proper values
```
Note that `__exit__` in this case needs to do additional checks (ie. self.proc could be None if exception occurred before spawn)

My preferred way of implementing context managers is to use [contextlib.contextmanager decorator](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) as it allows using try and with statements in standard way.

```python
class Phoenixd:
    @contextlib.contextmanager
    def running(self):
        try:
            self.run()
            yield self # all code inside with statement happens here
        except Exception:
            self.kill()
```

and than to use it:

```python
pxd = Phoenixd(...)
with pxd.running():
    # ...
```

trunner/device.py
```diff
@@ -30,13 +33,228 @@
...
+             '-b', str(self.baudrate),
+             '-s', self.dir],
+            cwd=self.cwd,
+            encoding='utf-8'
+        )
+
+        self.dispatcher_event = threading.Event()
+        if not self.wait_dispatcher:
+            # If we don't want to wait for a message dispatcher, just mark it as started
+            self.dispatcher_event.set()
```
xvuko, line 133: [nit] would be more readable if `if self.wait_dispather` block would be around lines 138-143

xvuko (approved):
LGTM - can be merged as is.

trunner/device.py
jsarzy: I agree, done!

PR #27, opened 2021-06-21
tests: random lifespan threads test

niewim19:
Testing if multiple threads with pseudorandom lifespan start and close correctly on imxrt106x

JIRA: CI-65

## Description
New test addition for imxrt106x

## Motivation and Context
To test if adding test with random lifespan crashes the imxrt106x platform

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt106x.

## Checklist:

## Special treatment


niewim19:
By `I  will merge this PR by myself when appropriate.` I mean that this PR is waiting for syspage name size correction as for now it is named incorrectly:
-is named: `tthr_rand`
-should be: test-thread-rand
But the correct name exceeds the maximum application name length in syspage or imxrt106x

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
...
+
+#include "unity_fixture.h"
+
+
+typedef struct threadss_self {
+	uint16_t time;
+	uint16_t id;
+} threadss_self_t;
+
+int fin;
```
nalajcie, line 13: static

nalajcie (changes requested):
not sure if intentionally left as a draft, but got here via email and did review anyway :).

I've pointed some things that need to be fixed.

Other notes:
 - when You fix coding style issue reported by reviewdod - please "resolve conversation" so the reviewer would know it was taken care of
 - I believe the the binary names can be longer now (see https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/199)
 - this test should also work on ia32 (there is nothing platform-specific here) so please include it in the testing.

threads/tthr_rand.c
niewim19: Used static

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
...
+
+
+void test_thread(void *arg)
```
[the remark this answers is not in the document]
niewim19: Used static

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
...
+	/* run threads */
+	for (int i = 0; i < cthread; i++) {
+		beginthread(test_thread, 6, &(stack[i * 512]), 512, &(params[i]));
+		usleep(100);
+	}
+
+	/* 	all threads should end in under one second */
+	sleep(1);
+	/* printf("End main thread %d\n",fin); */
+	TEST_ASSERT_TRUE(fin == cthread);
```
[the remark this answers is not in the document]
niewim19: You mean `TEST_ASSER_EQUAL_INT` :) Fixed, thanks!

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	fin = 0;
+	/* run threads */
+	for (int i = 0; i < cthread; i++) {
+		beginthread(test_thread, 6, &(stack[i * 512]), 512, &(params[i]));
+		usleep(100);
+	}
+
+	/* 	all threads should end in under one second */
```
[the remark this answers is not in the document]
niewim19: fixed, thanks!

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
+#include <string.h>
+#include <unistd.h>
+#include <sys/threads.h>
+
+#include "unity_fixture.h"
+
+
+typedef struct threadss_self {
```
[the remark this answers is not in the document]
niewim19: - yes, this was a typo that was propagated by mistake
 - removed struct name

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
...
+	uint16_t id;
+} threadss_self_t;
+
+int fin;
+
+
+void test_thread(void *arg)
+{
+	threadss_self_t *self = (threadss_self_t *)arg;
+	/* printf("Starting thr: %d for %d ms\n", self->id, self->time); */
```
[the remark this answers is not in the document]
niewim19: Commented code was removed

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
...
+
+	fin = 0;
+	/* run threads */
+	for (int i = 0; i < cthread; i++) {
+		beginthread(test_thread, 6, &(stack[i * 512]), 512, &(params[i]));
+		usleep(100);
+	}
+
+	/* 	all threads should end in under one second */
+	sleep(1);
```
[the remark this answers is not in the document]
niewim19: `threadJoin` used

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
...
+TEST(threads_rand, test_1)
+{
+	int cthread = 10, seed = 1087, n = 107;
+	threadss_self_t params[cthread];
+	void *stack[cthread * 512];
+
+	/* assign threads parameters (quick pseudorandom) */
+	for (int i = 0; i < cthread; i++) {
+		params[i].id = i; /* assign thread id */
+		n = (n * seed) % 200;
```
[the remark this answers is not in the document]
niewim19: `rand()` used. Also used srand to ensure the same seed for every run

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
...
+TEST_TEAR_DOWN(threads_rand)
+{
+}
+
+
+TEST(threads_rand, test_1)
+{
+	int cthread = 10, seed = 1087, n = 107;
+	threadss_self_t params[cthread];
+	void *stack[cthread * 512];
```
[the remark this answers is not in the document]
niewim19: Type changed to `char` and switched to static allocation

threads/tthr_rand.c
```diff
@@ -0,0 +1,83 @@
...
+int fin;
+
+
+void test_thread(void *arg)
+{
+	threadss_self_t *self = (threadss_self_t *)arg;
+	/* printf("Starting thr: %d for %d ms\n", self->id, self->time); */
+	usleep(1000 * self->time);
+	/* printf("Ending %d\n", self->id); */
+	fin++;
```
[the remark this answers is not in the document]
niewim19: Used __sync_fetch_and_add to reassure correct behavior

threads/threads.yaml
```diff
@@ -0,0 +1,7 @@
+test:
+    tests:
+        - name: test_threads_rand
+          exec: tthr_rand
+          type: unit
+          targets:
+              value: [armv7m7-imxrt106x]
```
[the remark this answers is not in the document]
niewim19: Yes. I started with a test only for imxrt106x, but it can be run also on ia32

threads/test_thread_rand.c
```diff
@@ -0,0 +1,88 @@
...
+typedef struct {
+	uint16_t time;
+	uint16_t id;
+} threads_self_t;
+
+
+static int cthread = 10, fin = 0;
+
+
+static char stack[10 * 512];
```
kemonats, line 18:
```suggestion
static char stack[cthread * 512];
```

threads/test_thread_rand.c
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/* run threads */
+	for (i = 0; i < cthread; i++) {
+		beginthread(test_thread, 6, &(stack[i * 512]), 512, &(params[i]));
+		usleep(100);
+	}
+
+	/* all threads should end in under one second */
+	for (i = 0; i < 10; i++) {
```
kemonats, line 62:
```suggestion
	for (i = 0; i < cthread; i++) {
```

pawelpisarczyk:
I think that this test should be located in proc subdirectory. General remark - please, before start writing a code, perform a short analysis of current project structure, coding convention etc. This will reduce our effort in making reviews.

threads/test_thread_rand.c
```diff
@@ -0,0 +1,88 @@
+#include <string.h>
```
pawelpisarczyk, line 1: File should start with the properly constructed label as it is described in documentation.

threads/test_thread_rand.c
```diff
@@ -0,0 +1,88 @@
...
+#include <sys/threads.h>
+
+#include "unity_fixture.h"
+
+
+typedef struct {
+	uint16_t time;
+	uint16_t id;
+} threads_self_t;
+
```
pawelpisarczyk, line 13: I don't understand the reason of using types with defined bit size. TIme and id should be defined using common C types e.g. unsigned int.

threads/test_thread_rand.c
pawelpisarczyk, line 18: We agreed to use common structures holding common. global attributes attributes. For code clarity is better to use two dimensional array.

threads/test_thread_rand.c
```diff
@@ -0,0 +1,88 @@
...
+
+static char stack[10 * 512];
+
+
+static void test_thread(void *arg)
+{
+	threads_self_t *self = (threads_self_t *)arg;
+	usleep(1000 * self->time);
+	__sync_fetch_and_add(&fin, 1);
```
pawelpisarczyk, line 25: Please don't use function embedded into the GNU CC compiler. Please synchronize global counter using mutex defined in _common structure. I don't understand why we should rely on GNU CC primitives. In this particular test we are not aiming to write high-performance code minimizing context switching. Rather we would like to test  correctness of constructing and destroying threads.

threads/test_thread_rand.c
```diff
@@ -0,0 +1,88 @@
...
+	/* assign threads parameters */
+	for (i = 0; i < cthread; i++) {
+		params[i].id = i;
+		params[i].time = 100 + rand() % 200; /* assign thread lifespan in milliseconds */
+	}
+
+	/* run threads */
+	for (i = 0; i < cthread; i++) {
+		beginthread(test_thread, 6, &(stack[i * 512]), 512, &(params[i]));
+		usleep(100);
```
pawelpisarczyk, line 58: What is the reason of adding here 100us delay?

threads/test_thread_rand.c
```diff
@@ -0,0 +1,88 @@
...
+
+
+TEST_TEAR_DOWN(threads_rand)
+{
+}
+
+
+TEST(threads_rand, test_1)
+{
+	threads_self_t params[cthread];
```
pawelpisarczyk, line 45: This is the inconsequence. Thread stacks are defined globally in BSS, but thread parameters are define locally on stack? Why?

threads/test_thread_rand.c
```diff
@@ -0,0 +1,88 @@
...
+	int ret, joined = 0, i;
+
+	srand(7);
+	/* assign threads parameters */
+	for (i = 0; i < cthread; i++) {
+		params[i].id = i;
+		params[i].time = 100 + rand() % 200; /* assign thread lifespan in milliseconds */
```
pawelpisarczyk, line 52: I think that time randomization range should be defined properly using defined scale projected on RAND_MAX macro.

threads/threads.yaml
```diff
@@ -0,0 +1,7 @@
+test:
+    tests:
+        - name: test_threads_rand
+          exec: test_thread_rand
+          type: unit
+          targets:
+              include: [armv7m7-imxrt106x]
```
pawelpisarczyk, line 7: This test should be accessible on all hardware platforms. Marek mentioned this in his review.

pawelpisarczyk (dismissed):
Please change the code according to our requests.

threads/test_thread_rand.c
niewim19: As `cthread` and `stack` are now stored in struct i am not able to define size of `stack` base on `cthread` value. Switching to dynamic allocation allows for it but I want to stay simple and do not allocate and then free memory for stack.

threads/test_thread_rand.c
niewim19: Thanks! Changed.

threads/test_thread_rand.c
niewim19: Yes, my mistake, label added.

threads/test_thread_rand.c
niewim19: At first i was experimenting with `uint16_t` instead of `char` while defining stack. Changed to unsigned int.

threads/test_thread_rand.c
niewim19: Two dimensional array implemented. `thread_common` structure also created.

threads/test_thread_rand.c
niewim19: Changed global counter synchronization. Now using `mutexLock()` and `mutexUnlock()` with lock stored in `thread_common` structure.

threads/test_thread_rand.c
niewim19: Yes it is unnecessary here. Removed.

threads/test_thread_rand.c
niewim19: Stack was moved to BSS because it was necessary to even store it. Parameters were left as local. Now with common struct `params` they are also stored in BSS for consistency. 

Only variables `joined` and `i` were left on stack as they scope is limited only for use in main test function `TEST(threads_rand, test_1)`

threads/test_thread_rand.c
niewim19: Yes it would be more professional, but i would rather stay with this simple approach with just `%` and added constant value as we only need an integer and we do not really bother with edge cases. This saves us couple of casts (from integer to double and then finally to integer for `sleep` functions) and keeps the code clean.

threads/threads.yaml
niewim19: Previous Marek`s note was resolved with change from
```
targets:
    value: [armv7m7-imxrt106x]
```
to
```
targets:
    include: [armv7m7-imxrt106x]
```
as `value:` specifies the only architecture to run test on, and `include:` adds it to test targets.

proc/test_thread_rand.c
```diff
@@ -0,0 +1,108 @@
+/*
+ * Phoenix-RTOS
+ *
+ * phoenix-rtos-tests
```
nalajcie, line 4: Please put a short description here. What You're testing?

proc/test_thread_rand.c
```diff
@@ -0,0 +1,108 @@
...
+typedef struct {
+	unsigned int time;
+	unsigned int id;
+} threads_self_t;
+
+
+typedef struct {
+	unsigned int cthread;
+	unsigned int fin;
+	char stack[10][512];
```
nalajcie, line 31: aligned properly by accident, please add explicit alignment

proc/test_thread_rand.c
```diff
@@ -0,0 +1,108 @@
...
+	srand(7);
+	/* assign threads parameters */
+	for (i = 0; i < threads_common.cthread; i++) {
+		threads_common.params[i].id = i;
+		threads_common.params[i].time = 100 + rand() % 200; /* assign thread lifespan in milliseconds */
+	}
+
+	/* run threads */
+	for (i = 0; i < threads_common.cthread; i++)
+		beginthread(test_thread, 6, &(threads_common.stack[i]), 512, &(threads_common.params[i]));
```
nalajcie, line 79: maybe proper `sizeof` instead of hardcoded value?

proc/test_thread_rand.c
```diff
@@ -0,0 +1,108 @@
...
+}
+
+
+TEST(threads_rand, test_1)
+{
+	int joined = 0, i;
+
+	srand(7);
+	/* assign threads parameters */
+	for (i = 0; i < threads_common.cthread; i++) {
```
nalajcie, line 72: could be done in a single `for` loop with starting the thread

proc/threads.yaml
```diff
@@ -0,0 +1,7 @@
+test:
+    tests:
+        - name: test_threads_rand
```
nalajcie, line 3: why `threads` / `thread`?

proc/test_thread_rand.c
```diff
@@ -0,0 +1,108 @@
...
+
+
+typedef struct {
+	unsigned int time;
+	unsigned int id;
+} threads_self_t;
+
+
+typedef struct {
+	unsigned int cthread;
```
nalajcie, line 29: `cthread` is misleading, when I change the value below the program will break.

I would prefer to make it `static const int` - then it can be used as a const (so also in structs definition):
```c
static const int cthread = 10;

typedef struct {
	unsigned int fin;
	char stack[cthread][512];
	[...]
}
```

You can also group (sort) struct fields according to their usage (eg. `fin` and `fin_lock`, `stack` and `params`),

proc/test_thread_rand.c
```diff
@@ -0,0 +1,108 @@
...
+
+static threads_common_t threads_common = { .cthread = 10, .fin = 0 };
+
+
+static void test_thread(void *arg)
+{
+	threads_self_t *self = (threads_self_t *)arg;
+	usleep(1000 * self->time);
+
+	mutexLock(threads_common.fin_lock);
```
nalajcie, line 45: I don't see this mutex being created anywhere. Does this even work!?

nalajcie (changes requested):
Some code fixes are still needed.

I don't see this test being run automatically - **this is a requirement for merging this test**:
https://github.com/phoenix-rtos/phoenix-rtos-tests/runs/2892877920?check_suite_focus=true#step:6:47

threads/threads.yaml
nalajcie: still not being run on ia32-generic

threads/test_thread_rand.c
nalajcie: label needs fixing, please read what You're copying!

threads/test_thread_rand.c
nalajcie: obsolete now.

threads/tthr_rand.c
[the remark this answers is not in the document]
nalajcie: obsolete.

proc/test_thread_rand.c
nalajcie: sorry, my bad, not possible with GCC and pure C (clang actually compiles that even if it's not C standard-compilant). 

We either need to use `#define` or switch to dynamic allocation.

proc/test_thread_rand.c
```diff
@@ -0,0 +1,108 @@
+/*
+ * Phoenix-RTOS
+ *
+ * phoenix-rtos-tests
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Hubert Buczynski
```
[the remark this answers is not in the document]
niewim19: Yes, my bad. Changed.

proc/test_thread_rand.c
niewim19: Description added. However some of other test files need checking cause they lack short description.

proc/test_thread_rand.c
niewim19: Added explicit alignment for `stack`

proc/test_thread_rand.c
niewim19: Used `sizeof`. Thanks

proc/test_thread_rand.c
niewim19: This loop was divided into two for better clarity, but you are right: one loop is shorter and still clear to read.

proc/threads.yaml
niewim19: My mistake. Changed plural form to singular `thread`

proc/test_thread_rand.c
niewim19: I moved `stack` into `thread_self_t` and separated out `time` and `id` into `thread_params_t` structure. Now we can define the amount of threads in one place and still have static allocation. 
Also the split was done to avoid passing thread `stack` as parameter to itself.  

As for `cthread` being misleading: yes its true. Now `cthread` variable is only local for main test run and not global and its value is based on size of `threads` array.

proc/test_thread_rand.c
niewim19: Surprisingly it works. Added proper creation and destroy of mutex with `mutexCreate()` and `resourceDestroy()`.

niewim19:
Tests on `ia32` were blocked because of _not matching the pattern_ filename of `.yaml` script. It was changed to `test.yaml`.

niewim19:
Removed unnecessary includes

proc/test_thread_rand.c
nalajcie: Actually - You don't know as You're not checking the result code of `mutexLock` so You don't know if it succeeded.

Please add checking result code to all native calls (`mutex*`, `begin/endthread`, `resourceDestroy`)

nalajcie (changes requested):
Please add checking result code to all native calls (`mutex*`, `begin/endthread`, `resourceDestroy`).

I would like to see more thread-related unit tests in the same file/binary in the future (we can change file names then).

niewim19:
Added checking the state. Went straight for checking the output value using _unity_ functions (not using if-s) as we have to mainly catch errors, not process them.

`endthread()` returns void so no check for it.

proc/test_thread_rand.c
```diff
@@ -0,0 +1,121 @@
...
+
+
+static void test_thread(void *arg)
+{
+	thread_params_t *params = (thread_params_t *)arg;
+	int err;
+	usleep(params->time * 1000);
+
+	err = mutexLock(threads_common.fin_lock);
+	TEST_ASSERT_EQUAL_INT(err, EOK);
```
nalajcie, line 51: please note that the `TEST_ASSERT_EQUAL_INT` is defined as: `TEST_ASSERT_EQUAL_INT(expected, actual)`, so You need to change the order of params for some of the calls).

proc/test_thread_rand.c
```diff
@@ -0,0 +1,121 @@
...
+	for (i = 0; i < cthread; i++) {
+		threads_common.threads[i].params.id = i;
+		threads_common.threads[i].params.time = 100 + rand() % 200; /* assign thread lifespan in milliseconds */
+		err = beginthread(test_thread, 6, &(threads_common.threads[i].stack), sizeof(threads_common.threads[i].stack), &(threads_common.threads[i].params));
+		TEST_ASSERT_EQUAL_INT(err, EOK);
+	}
+
+	/* all threads should end in under one second */
+	for (i = 0; i < cthread; i++) {
+		threadJoin(0);
```
nalajcie, line 93: please check return code

nalajcie (commented):
Looks much better, pointed out small issues to be fixed.

proc/test_thread_rand.c
niewim19: `threadJoin(0);` here returns id (not local) of thread. https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/964cd68eac3da0175fd98abc25cb65538dbf1825/proc/threads.c#L1138-L1146

So there is nothing to check the return value against.

proc/test_thread_rand.c
niewim19: Checked and changed if necessary. Thanks.

proc/test_thread_rand.c
niewim19: Ok. After consulting the error check is added.

pawelpisarczyk:
niewim19 what is the status of solving issues?

proc/test_thread_rand.c
jsarzy: I think we should rather not use `TEST_ASSERT*` in threads. It uses `longjmp` and/or sets a global state which is not protected by any synchronization primitives.

proc/test_thread_rand.c
```diff
@@ -0,0 +1,123 @@
...
+
+typedef struct {
+	thread_params_t params;
+	char stack[512] __attribute__((aligned(8))); /* stack needs to be aligned to 8 */
+} thread_self_t;
+
+
+typedef struct {
+	unsigned int fin;
+	handle_t fin_lock;
```
pawelpisarczyk, line 37: Please change this name to word without `_` to be inline with our coding standards.

proc/test.yaml
```diff
@@ -0,0 +1,7 @@
+test:
+    tests:
+        - name: test_thread_rand
+          exec: test_thread_rand
+          type: unit
+          targets:
+              include: [armv7m7-imxrt106x]
```
pawelpisarczyk, line 7: Marek and me asked you to compile this test on all targets. When this issue will be addressed?

proc/test_thread_rand.c
```diff
@@ -0,0 +1,123 @@
...
+	cthread = sizeof(threads_common.threads) / sizeof(threads_common.threads[0]);
+	err = mutexCreate(&(threads_common.fin_lock));
+	TEST_ASSERT_EQUAL_INT(EOK, err);
+	threads_common.fin = 0;
+
+	srand(7);
+	/* assign threads parameters */
+	for (i = 0; i < cthread; i++) {
+		threads_common.threads[i].params.id = i;
+		threads_common.threads[i].params.time = 100 + rand() % 200; /* assign thread lifespan in milliseconds */
```
pawelpisarczyk, line 87: I've asked you to use RAND_MAX constant to scale the randomization space to defined by constant.

proc/test_thread_rand.c
niewim19: Yes, that name is not within our coding standards. I will fix that.

proc/test.yaml
niewim19: Test is now build and tested on all targets that our runner supports (ia32 and imxrt106x). When the automatic tests will be available for imxrt106x (soon) even that include will not be necessary. The same goes with other automatic tests currently available.

The term `include:` does not specify _the only_ architecture but it adds it. Even if I specify all architectures here, tests do not build everywhere - only on imxrt106x, ia32 and host-pc

proc/test_thread_rand.c
niewim19: Ok, added random number generation based on RAND_MAX and rand().

Also, for better randomization:
 - switched to microseconds
 - added `timemax` and `timemin` to `threads_common`

proc/test_thread_rand.c
niewim19: Ok, so the reasonable option is to in some way escape the error from thread. I can make another parameter in `thread_params_t` for that.

niewim19:
Test runner fail **not** due to `test_thread_rand` fail

proc/test_thread_rand.c
niewim19: @nalajcie , is this an issue?

nalajcie: Oh, sorry, didn't know that it uses `longjmp`. As doing `longjmp` from different thread that `setjmp` is undefined behavior - it may (and probably will) result in kernel panic or some strange errors (You may test it by Yourself ny failing on purpose).

For now we can add some result code in `thread_params_t` and verify it in the main thread, maybe we would come up with better solution for multithreading support in `unity` framework in the future.

proc/test.yaml
nalajcie: @pawelpisarczyk this is taken care of, `targets` is a dict which supports attributes (`include`, `exclude` and `values`). The default value for `values` is `DEFAULT_TARGETS` (currently `ia32-generic` as all tests should work on this platform).

The final set of targets is `targets['values'] + targets['include'] - targets['exclude']` (yeah, seems unnaturally complex, but common pattern in YAML configs) - works great when You have multiple tests in single YAML and You globally set targets for all of them and then just include/exclude one or two for a certain target)

nalajcie:
> Test runner fail **not** due to `test_thread_rand` fail. Also I am awaiting commentary on using `TEST_ASSERT*` inside threads. When that issue will be resolved i will push final commit together with reviewdog suggestion.

There were some changes in new toolchain and respective docker images. Please rebase Your branch on top of current master - then the tests will work again.

proc/test_thread_rand.c
niewim19: Thanks! Going on with new field in `thread_params_t` then.

niewim19:
Status of this PR recent changes:
 - added proper error handling from `thread*` and `mutex*` functions, which are thread safe.
 - changed the resolution of time variable by changing to microseconds and using `RAND_MAX` for greater randomization

nalajcie (approved):
You could use `float` instead of `double`, but apart from that te test is ready to merge.

Other improvements might be done as more threading unit tests will be added to this test group.

PR #34, opened 2021-07-09
busybox: add running busybox testsuite

damianloew:
JIRA: RTOS-59

## Description

- busybox: add running busybox testsuite
- busybox: add a check if the test suite is in the bin directory
- busybox: add strerror and _exit from child process when failed 
- busybox: add posixsrv exec error handling 
- busybox: allow test_busybox to be run using test runner as single test 
- busybox: add long_test argument for runner to allow running busybox test suite
- busybox: temporarily disable check if posixsrv has been run properly, because of waitpid WNOHANG problem
- busybox: add export bindir variable, because of test suite location change
- busybox: allow to run each busybox command tests seperately using test runner
- runner: add alternative option for disabling long tests, check long_test arg earlier

## Motivation and Context

Test_busybox runs a busybox testsuite (which is copied to /usr/test/busybox)
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic - qemu).

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/18).

### DIFF
--- a/busybox/test_busybox.c
+++ b/busybox/test_busybox.c
@@ -1,0 +1,48 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <dirent.h>
+#include <errno.h>
+
+int main(int argc, char **argv)
+{
+	int ret;
+	char cmdargs[20];
+	char cmd[140] = "cd /usr/test/busybox/testsuite/ && export PATH=/bin:/sbin:/usr/bin:/usr/sbin && export bindir=/bin && ./runtest";
+	DIR *dir = opendir("/usr/test/busybox/testsuite");
+
+	if (dir) {
+		closedir(dir);
+	}
+	else if (ENOENT == errno) {
+		fprintf(stderr, "There is no busybox test suite to run, build project with \"LONG_TEST = 'y'\"\n");
+		return 1;
+	}
+	else {
+		fprintf(stderr, "There is problem with opening existing /bin/testsuite directory: %s\n", strerror(errno));
+		return 1;
+	}
+
+	if (argc == 2 && (strlen(argv[1]) <= (sizeof(cmdargs) - 5)))
+		sprintf(cmdargs, " -v %s", argv[1]);
+	else if (argc == 1)
+		sprintf(cmdargs, " -v");
+	else {
+		fprintf(stderr, "The argument is too long!");
+		return 1;
+	}
+	strncat(cmd, cmdargs, sizeof(cmd) - strlen(cmd) - 1);
+
+	if ((ret = system(cmd)) < 0) {
+		fprintf(stderr, "system function failed: %s\n", strerror(errno));
+		return 1;
+	}
+
+	if (argc == 2)
+		printf("\n****A single test of the Busybox Test Suite completed****\n\n");
+	else if (argc == 1)
+		printf("\n****The Busybox Test Suite completed****\n\n");
+
+	return 0;
+}
--- a/runner.py
+++ b/runner.py
@@ -78,6 +78,14 @@ def parse_args():
                         default=False, action='store_true',
                         help="Board will not be flashed by runner.")
 
+    # Add alternative option "--long-test" "" for long-test argument, which is False
+    # Needed for proper disabling long tests in gh actions
+    # because of problem with passing empty argument through gh workflows
+    parser.add_argument("--long-test",
+                        type=bool,
+                        nargs='?', default=False, const=True,
+                        help="Long tests will be run")
+
     args = parser.parse_args()
 
     args.log_level = logging_level[args.log_level]
@@ -107,7 +115,8 @@ def main():
                          build=args.build,
                          flash=not args.no_flash,
                          serial=(args.serial, args.baudrate),
-                         log=(args.log_level == logging.DEBUG)
+                         log=(args.log_level == logging.DEBUG),
+                         long_test=args.long_test
                          )
 
     passed, failed, skipped = runner.run()
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -65,11 +65,24 @@ def array_value(array: Dict[str, List[str]]) -> List[str]:
 @dataclass
 class ParserArgs:
     targets: List[str]
+    long_test: bool
     yaml_path: Path
     path: Path = field(init=False)
 
     def __post_init__(self):
         self.path = self.yaml_path.parent
+
+
+def filter_tests(tests: List['TestConfig'], args: ParserArgs) -> List['TestConfig']:
+    ''' Removes tests not needed in the current test campaign'''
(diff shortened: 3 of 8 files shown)


damianloew:
https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/18

damianloew:
https://github.com/phoenix-rtos/phoenix-rtos-build/pull/75

busybox/test_busybox.c
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	if ((ret = vfork()) < 0) {
+		fprintf(stderr, "vfork function failed: %s\n", strerror(errno));
+		exit(EXIT_FAILURE);
+	}
+	else if (!ret) {
+		execv("/bin/posixsrv", argv);
+		fprintf(stderr, "exec function failed: %s\n", strerror(errno));
+		exit(EXIT_FAILURE);
```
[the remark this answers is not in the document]
damianloew: Changes applied, thanks for advice

damianloew:
> 
That's right result is once printed before and another time after test output. I know parsing output of passed tests to error message can be misleading, but I had to do this like that. Because the result can be unknown before printing output I have to catch this output every time. However I named variable like `msg`, not `error_msg` to make it less misleading.

Ok, I will remove this statement about posixsrv.

You are right, that's will be better with a whole test in seprate directory. I've created JIRA task for that to think about it in the future. 

If You mean test-cp, everything works well for me.

That's right, I propose to leave it, because in the near future posixsrv should be run automatically and the problem will dissapear.
https://github.com/phoenix-rtos/phoenix-rtos-project/issues/292

trunner/harness.py
```diff
@@ -33,6 +33,35 @@ def __str__(self):
...
+            color = Color.OK
+        elif self.status == BusyboxTestResult.FAIL:
+            color = Color.FAIL
+        elif self.status == BusyboxTestResult.IGNORE:
+            color = Color.SKIP
+
+        status = Color.colorify(self.status, color)
+
+        res = f"{status}: {self.name}"
+        if self.status == 'FAIL':
```
jsarzy, line 59: be consistent and use `BusyboxTestResult.FAIL` instead of 'FAIL'

trunner/harness.py
```diff
@@ -85,3 +114,46 @@ def harness(proc):
...
+    FINAL = r"\*\*\*\*(The Busybox Test Suite completed|A single test of the Busybox Test Suite completed)\*\*\*\*\r+\n"
+    MESSAGE = r"(.*?)\r+\n"
+
+    @staticmethod
+    def harness(proc):
+        # passed_nr = 0
+        # failed_nr = 0
+        # skipped_nr = 0
+        test_results = []
+        # last_match_was_msg = False
```
jsarzy, line 132: remove these commented variables

trunner/test_runner.py
```diff
@@ -50,6 +51,8 @@ def run(self):
         self.parse_tests()
 
         for test_config in self.test_configs:
+            if test_config["type"] == 'busybox' and not self.long_test:
```
jsarzy, line 54: We are mixing logic here. If there is long-test flag and type of the test is busybox then we should dismiss it at the parsing step.

busybox/test_busybox.c
```diff
@@ -0,0 +1,44 @@
...
+		fprintf(stderr, "There is no busybox test suite to run, build project with \"LONG_TEST = 'y'\"\n");
+		return 1;
+	}
+	else {
+		fprintf(stderr, "There is problem with opening existing /bin/testsuite directory: %s\n", strerror(errno));
+		return 1;
+	}
+
+	if (argc == 2)
+		sprintf(cmdargs, " -v %s", argv[1]);
```
jsarzy, line 28: nit: maybe is worth to check if `argv[1]` fits into `cmdargs` buffer to avoid bugs in the future development

busybox/test_busybox.c
```diff
@@ -0,0 +1,44 @@
...
+	else if (argc == 1)
+		sprintf(cmdargs, " -v");
+	strcat(cmd, cmdargs);
```
jsarzy, line 31: nit: as above, is good practice to use a safer function `strncat`

runner.py
```diff
@@ -78,6 +78,14 @@ def parse_args():
                         default=False, action='store_true',
                         help="Board will not be flashed by runner.")
 
+    # Add alternative option "--long-test" "" for long-test argument, which is False
+    # Needed for proper disabling long tests in gh actions
+    # because of problem with passing empty argument through gh workflows
+    parser.add_argument("--long-test",
+                        type=bool,
+                        nargs='?', default=False, const=True,
+                        help="Long tests will be run")
```
jsarzy, line 87: Can't we build some primitive ternary operator in github actions? Something like ${{ inputs.long_tests && '--long-test' || '' }}

jsarzy (changes requested):

trunner/test_runner.py
damianloew: I don't quite understand. I don't want to run busybox tests when the `long-test` argument is not specified, so I think the condition is ok. Please correct me if am I wrong. But that's right it can be checked earlier, I will correct it.

trunner/config.py
```diff
@@ -3,6 +3,7 @@
 from dataclasses import dataclass, field
 from itertools import chain
 from pathlib import Path
+from xmlrpc.client import Boolean
```
jsarzy, line 6: just use `bool`

trunner/config.py
```diff
@@ -65,13 +66,26 @@ def array_value(array: Dict[str, List[str]]) -> List[str]:
 @dataclass
 class ParserArgs:
     targets: List[str]
+    long_test: Boolean
```
jsarzy, line 69:
```suggestion
    long_test: bool
```

trunner/test_runner.py
```diff
@@ -35,9 +36,10 @@ def search_for_tests(self):
     def parse_tests(self):
         self.test_configs = []
         for path in self.test_paths:
-            args = ParserArgs(yaml_path=path, targets=self.targets)
+            args = ParserArgs(yaml_path=path, targets=self.targets, long_test=self.long_test)
             config = TestCaseConfig.from_yaml(args)
-            self.test_configs.extend(config.tests)
+            for test_config in config.tests:
+                self.test_configs.append(test_config)
```
jsarzy, line 42: there is no need to change it

jsarzy (commented):
few remarks and will be ready to merge

trunner/test_runner.py
damianloew: Thanks, changes applied in the last force push.

jsarzy (approved):

PR #41, opened 2021-07-27
runner: imxrt106x: add relay restart

damianloew:
JIRA: CI-69

## Description

- runner: imxrt106x: change serial port to usb socket reference
- runner: imxrt106x: add _restart_by_poweroff procedure
- runner: imxrt106x: rename boot to reboot, add _restart_by_poweroff to it and add double reboot to load method
- runner: imxrt106x: decrease waiting for dev time step value
- runner: add power_usb_ports(enable: bool) function for power on and off rpi ports
- runner: add unbind_rpi_usb(port_address) function and DEVICE_SERIAL_USB const for port address

## Motivation and Context

It's for running each test on fresh mimxrt1064 eval kit (after power down and power up)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt106x - evk).

## Checklist:

## Special treatment

### DIFF
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -41,6 +41,9 @@ DEVICE_TARGETS = ['armv7m7-imxrt106x']
 # Port to communicate with hardware board
 DEVICE_SERIAL = "/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.1"
 
+# DEVICE_SERIAL USB port address
+DEVICE_SERIAL_USB = "1-1.4"
+
 
 def rootfs(target: str) -> Path:
     return PHRTOS_PROJECT_DIR / '_fs' / target / 'root'
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -39,7 +39,7 @@ DEFAULT_TARGETS = [target for target in ALL_TARGETS
 DEVICE_TARGETS = ['armv7m7-imxrt106x']
 
 # Port to communicate with hardware board
-DEVICE_SERIAL = "/dev/ttyACM0"
+DEVICE_SERIAL = "/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.1"
 
 
 def rootfs(target: str) -> Path:
--- a/trunner/device.py
+++ b/trunner/device.py
@@ -46,6 +46,18 @@ def wait_for_dev(port, timeout=0):
             raise TimeoutError
 
 
+def power_usb_ports(enable: bool):
+    uhubctl = subprocess.run([
+        'uhubctl',
+        '-l', '2',
+        '-a', f'{1 if enable else 0}'],
+        stdout=subprocess.DEVNULL
+    )
+    if uhubctl.returncode != 0:
+        logging.error('uhubctl failed!\n')
+        raise Exception('RPi usb ports powering up/down failed!')
+
+
 class Psu:
     """Wrapper for psu program"""
 
@@ -383,29 +395,13 @@ class IMXRT106xRunner(DeviceRunner):
         f.write("1-1.4")
         f.close()
         # without powering down rpi ports eval kit is partially powered through usb
-        rpi_ports_off = subprocess.run([
-            'uhubctl',
-            '-l', '2',
-            '-a', '0'],
-            stdout=subprocess.DEVNULL
-        )
-        if rpi_ports_off.returncode != 0:
-            logging.error('uhubctl failed!\n')
-            raise Exception('RPi ports power down failed!')
 
+        power_usb_ports(False)
         self.power_gpio.low()
         time.sleep(0.500)
         self.power_gpio.high()
         time.sleep(0.500)
-        rpi_ports_on = subprocess.run([
-            'uhubctl',
-            '-l', '2',
-            '-a', '1'],
-            stdout=subprocess.DEVNULL
-        )
-        if rpi_ports_on.returncode != 0:
-            logging.error('uhubctl failed!\n')
-            raise Exception('RPi ports power up failed!')
+        power_usb_ports(True)
 
         try:
             wait_for_dev(DEVICE_SERIAL, timeout=5)
--- a/trunner/device.py
+++ b/trunner/device.py
@@ -40,8 +40,8 @@ def wait_for_dev(port, timeout=0):
 
     # naive wait for dev
     while not os.path.exists(port):
-        time.sleep(0.5)
-        asleep += 0.5
+        time.sleep(0.01)
+        asleep += 0.01
         if timeout and asleep >= timeout:
             raise TimeoutError
 
--- a/trunner/device.py
+++ b/trunner/device.py
@@ -413,16 +413,19 @@ class IMXRT106xRunner(DeviceRunner):
             logging.error('Serial port not found!\n')
             sys.exit(1)
 
-    def boot(self, serial_downloader=False):
+    def reboot(self, serial_downloader=False, cut_power=False):
         if serial_downloader:
             self.boot_gpio.low()
         else:
             self.boot_gpio.high()
 
-        self._restart_by_jtag()
+        if cut_power:
+            self._restart_by_poweroff()
+        else:
+            self._restart_by_jtag()
 
     def flash(self):
-        self.boot(serial_downloader=True)
+        self.reboot(serial_downloader=True, cut_power=True)
 
         phd = None
         try:
@@ -444,16 +447,19 @@ class IMXRT106xRunner(DeviceRunner):
             logging.info(exception)
             sys.exit(1)
 
-        self.boot()
+        self.reboot(cut_power=True)
 
     def load(self, test):
         """Loads test ELF into syspage using plo"""
 
         phd = None
         load_dir = str(rootfs(test.target) / 'bin')
+        self.reboot(cut_power=True)
         try:
             with PloTalker(self.port) as plo:
-                self.boot()
+                # Because of powering all Rpi ports after powering the board,
+                # there is need to second reboot (without cut power) in order to capture all data
+                self.reboot(cut_power=False)
                 plo.wait_prompt()
 
                 if not test.exec_cmd:
--- a/trunner/device.py
+++ b/trunner/device.py
@@ -9,6 +9,7 @@ import time
 import pexpect
 import pexpect.fdpexpect
 import serial
+import subprocess
 
 from pexpect.exceptions import TIMEOUT, EOF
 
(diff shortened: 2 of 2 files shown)


trunner/device.py
```diff
@@ -374,13 +389,52 @@ def reset(self):
...
+
+        # after power on there is pre-restart device, which will disappear
+        wait_for_disappear(DEVICE_SERIAL, timeout=5)
+        try:
+            wait_for_dev(DEVICE_SERIAL, timeout=5)
+        except TimeoutError:
+            logging.error('Serial port not found!\n')
+            sys.exit(1)
+
+    def boot(self, serial_downloader=False, restart=True):
```
jsarzy, line 428: Maybe instead of `boot(restart=True)` use self.restart() and keep old logic of boot function?

trunner/device.py
```diff
@@ -374,13 +389,52 @@ def reset(self):
...
             self.boot_gpio.high()
 
-        self.reset()
+        if restart:
+            self.restart()
+        else:
+            self.reset()
 
     def flash(self):
         self.boot(serial_downloader=True)
```
jsarzy, line 440: Why we restart a board before each test but not before flashing?

trunner/device.py
```diff
@@ -412,9 +466,10 @@ def load(self, test):
 
         phd = None
         load_dir = str(rootfs(test.target) / 'bin')
+        self.boot()
```
jsarzy, line 469: As I mentioned earlier maybe `self.restart()` (or even `self.hard_restart()`) and keep self.boot as it was.

trunner/device.py
```diff
@@ -374,13 +389,52 @@ def reset(self):
...
+            '-l', '2',
+            '-a', '1'],
+            stdout=subprocess.DEVNULL
+        )
+        if rpi_ports_on.returncode != 0:
+            logging.error('uhubctl failed!\n')
+            raise Exception('RPi ports power up failed!')
+
+        # after power on there is pre-restart device, which will disappear
+        wait_for_disappear(DEVICE_SERIAL, timeout=5)
```
jsarzy, line 421: What if device won't disappear?

damianloew: If device won't disappear after 5 seconds it means that's the right device and wait_for_dev is executed (which doesn't wait, because dev is ready)

trunner/device.py
damianloew: Before flashing for example there is boot without reset, but with restart. In this case the boot (with default reset) and after that self.restart() should be used. I think one function self.boot(restart=true) is better. Btw reset and restart is one abstraction layer lower than boot and that's why I did it that way.

trunner/device.py
damianloew: We restart a board before flashing in self.boot(serial_downloader=True) after 'def flash(self):'

trunner/device.py
nalajcie: ummm... why do we even need rebooting twice here?

trunner/device.py
```diff
@@ -374,13 +390,52 @@ def reset(self):
         time.sleep(0.050)
         self.reset_gpio.high()
 
-    def boot(self, serial_downloader=False):
+    def restart(self):
+        # without powering down rpi ports eval kit is partially powered through usb
+        rpi_ports_off = subprocess.run([
```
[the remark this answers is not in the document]
nalajcie: Is uhubctl needed on RPi host machine or is it a part of runner docker image?

I would much welcome info regarding how to setup new RPi on our internal confluence :).

nalajcie (changes requested):
Overall the change is good, I would like the naming to be more straight-forward (for better readability) and focus on making the reboot process as fast as possible.

trunner/device.py
damianloew: We need to turn Rpi ports on after powering up eval kit (because of partial supplying board through usb port, when it's powered off using relay - this causes some problems, so we avoid that). After starting the Rpi ports (it takes a while, because after restart there is pre-restart device on serial port, which has to disappear and after that there appears proper device) there is some output since restart, which disappears so we reset board second time to capture whole output.

trunner/device.py
damianloew: Thanks, I will test using /sys/bus/usb/drivers/usb/unbind and bind

trunner/device.py
```diff
@@ -374,13 +390,52 @@ def reset(self):
         time.sleep(0.050)
         self.reset_gpio.high()
 
-    def boot(self, serial_downloader=False):
+    def restart(self):
```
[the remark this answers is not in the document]
damianloew: I agree with this approach, I will apply it (without using the strategy for now)

trunner/config.py
```diff
@@ -39,7 +39,7 @@ def resolve_phrtos_dir() -> Path:
 DEVICE_TARGETS = ['armv7m7-imxrt106x']
 
 # Port to communicate with hardware board
-DEVICE_SERIAL = "/dev/ttyACM0"
+DEVICE_SERIAL = "/dev/mimxrt1064-debug-usb"
```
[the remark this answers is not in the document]
damianloew: I added "KERNEL=="ttyACM*", ATTRS{idVendor}=="0d28", ATTRS{idProduct}=="0204", SYMLINK+="mimxrt1064-debug-usb" to 10-local.rules file in /etc/udev/rules.d on both raspberry, because of differences in standard udev serial symlinks on both eval kits ("usb-ARM_DAPLink_CMSIS-DAP_0232000007b01b8800000000000000000000000097969905-if01" and "usb-ARM_DAPLink_CMSIS-DAP_023200001ce1efee00000000000000000000000097969905-if01"). 

Symlinks from by-path directory are different too:
platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.3:1.0-port0
platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.1

trunner/device.py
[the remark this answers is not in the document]
damianloew: That's right, I will try to make it better. Uhubctl need to be part of the docker image

trunner/device.py
nalajcie: If there is an issue like that it needs to be written in the code in comment - othewise this looks like a bug/remnant of old code so I would delete it when working on the code :).

Regarding the issue:
 - the "old devices" issue may go away if You unbind the device directly via sysfs (see my other comment)
 - as I'm striving for as-fast-as possible reboots - we might try to circumvent it in some other way in the future (hmm... maybe cutting 5V from the cable would do it? Or maybe there is some other way of separating it or draining the voltage?)

trunner/config.py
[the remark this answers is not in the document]
nalajcie: - if we need to add specific UDEV rules for it to work, than it needs to be described on confluence
 - regarding the naming 
  - "by serial" being different is expected (they should have different serial)
  - "by path" should be the same if we use the same ports. Please investigate outputs of `lsusb -tv` indeed the same port is being used on every RPi

trunner/device.py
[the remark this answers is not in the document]
nalajcie: ok, so I would welcome PR for changes in docker image needed :)

trunner/device.py
```diff
@@ -374,13 +390,52 @@ def reset(self):
...
+            '-a', '0'],
+            stdout=subprocess.DEVNULL
+        )
+        if rpi_ports_off.returncode != 0:
+            logging.error('uhubctl failed!\n')
+            raise Exception('RPi ports power down failed!')
+
+        self.restart_gpio.low()
+        time.sleep(0.500)
+        self.reset_gpio.low()
```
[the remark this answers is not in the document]
damianloew: I thought I read such a recommendation in the reference manual, but now it seems to me that I may have been wrong and the internal POR_B signal regardless of the state of the external line is low during the power-up sequence. I will remove that reset. 
Screenshot from 2021-08-10 12-43-50

trunner/device.py
damianloew: Ok, I will add some comment about that. I tried with cutting 5V and with desoldering a zero resistor, which connect board 5V_SYS and 5V_USB_OTG. From what I remember then psu fails (port wasn't available)

damianloew:
> Overall the change is good, I would like the naming to be more straight-forward (for better readability) and focus on making the reboot process as fast as possible.

Changes applied

trunner/device.py
```diff
@@ -39,12 +40,44 @@ def wait_for_dev(port, timeout=0):
...
     while not os.path.exists(port):
-        time.sleep(0.5)
-        asleep += 0.5
+        time.sleep(0.01)
+        asleep += 0.01
         if timeout and asleep >= timeout:
             raise TimeoutError
 
 
+def switch_rpi_ports(state):
```
nalajcie, line 49: maybe `power_usb_ports(enabled: bool)`? 

than we can skip string parsing and write 
```python
'-a', f'{1 if enable else 0}'`
```

also calling the function is readable `power_usb_ports(False)`

trunner/device.py
```diff
@@ -39,12 +40,44 @@ def wait_for_dev(port, timeout=0):
...
+    if state.lower() == 'on':
+        bool_state = 1
+    elif state.lower() == 'off':
+        bool_state = 0
+    else:
+        print('state argument should be "on" or "off"')
```
nalajcie, line 55: if the above changes would be introduced no "third" value is possible. Even so - I would throw an error here as it's clearly a programmer error.

trunner/device.py
```diff
@@ -39,12 +40,44 @@ def wait_for_dev(port, timeout=0):
...
+        '-l', '2',
+        '-a', f'{bool_state}'],
+        stdout=subprocess.DEVNULL
+    )
+    if uhubctl.returncode != 0:
+        logging.error('uhubctl failed!\n')
+        raise Exception(f'RPi ports power {state} failed!')
+
+
+def unbind_rpi_usb(port_address):
```
nalajcie, line 68: does unbinding USB device make the restart quicker?

nalajcie (commented):
Some nitpicking, other than that I think it's ready to be merged

trunner/device.py
damianloew: Yes, there is no need to wait for disappear. All tests take about 2 minutes.

trunner/device.py
```diff
@@ -39,12 +40,37 @@ def wait_for_dev(port, timeout=0):
...
+        logging.error('uhubctl failed!\n')
+        raise Exception('RPi usb ports powering up/down failed!')
+
+
+def unbind_rpi_usb(port_address):
+    try:
+        with open('/sys/bus/usb/drivers/usb/unbind', 'w') as file:
+            file.write(port_address)
+    except PermissionError:
+        logging.error("/sys/bus/usb/drivers/usb/unbind: PermissionError\n\
```
nalajcie, line 66: nitpicking:
 - You could have used python multiline string `"""` in here
 - I would go for raising exception for consistency with other error code paths
 - please rewrite/update RPi preparation process in confluence

Not a merge blocker, can be done in separate PR.

nalajcie (approved):

trunner/device.py
damianloew: Thank You, I will remember about this

PR #43, opened 2021-07-28
psh: tab autocompletion test

niewim19:
JIRA: PD-160

tests regarding autoompletion using `tab` in psh.

## Description
Test check how the autocompleting and hint feature work in psh

## Motivation and Context

 - Related issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/134
 - Related PR: https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/78 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work ( https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/78 )

### DIFF
--- a/psh/test-autocompletion.py
+++ b/psh/test-autocompletion.py
@@ -1,0 +1,114 @@
+# Phoenix-RTOS
+#
+# phoenix-rtos-tests
+#
+# "auth" psh applet test
+#
+# Copyright 2021 Phoenix Systems
+# Author: Mateusz Niewiadomski
+#
+# This file is part of Phoenix-RTOS.
+#
+# %LICENSE%
+#
+import re
+
+from psh.tools.basic import run_psh, assert_only_prompt, assert_prompt
+
+
+TAB = '\t'
+ENTER = '\n'
+
+
+def mkdir(p, path):
+    p.send('mkdir ' + path + ENTER)
+    p.expect_exact('mkdir ' + path)
+    assert_prompt(p, msg='Cannot make directory: ' + path)
+
+
+def touch(p, path):
+    p.send('touch ' + path + ENTER)
+    p.expect_exact('touch ' + path)
+    assert_prompt(p, msg='Cannot make file: ' + path)
+
+
+def assert_completion(p, incomplete, complete):
+    p.send(incomplete)
+    p.expect_exact(incomplete)
+    p.send(TAB)
+    if complete.startswith(incomplete):
+        addition = re.sub('^' + incomplete, '', complete)
+        p.expect_exact(addition)
+        p.send(ENTER)
+        assert_prompt(p)
+    else:
+        raise Exception("Error in test: uncomplete path not subset of complete path")
+
+
+def assert_multiple_completion(p, comptab):
+    for pair in comptab:
+        p.send(pair[0])
+        p.expect_exact(pair[0])
+        p.send(TAB)
+        p.expect_exact(pair[1])
+    p.send(ENTER)
+    assert_prompt(p)
+
+
+def assert_hints(p, path, hints):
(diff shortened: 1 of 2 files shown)


psh/test-autocompletion.py
```diff
@@ -0,0 +1,114 @@
...
+    assert_prompt(p, msg='Cannot make directory: ' + path)
+
+
+def touch(p, path):
+    p.send('touch ' + path + ENTER)
+    p.expect_exact('touch ' + path)
+    assert_prompt(p, msg='Cannot make file: ' + path)
+
+
+def assert_completion(p, incomplete, complete):
```
jsarzy, line 35: Maybe use names for variables that more specific to a "string" terminology.
`incomplete` -> `prefix`
`complete` -> `word`
`addition` -> `suffix`

psh/test-autocompletion.py
```diff
@@ -0,0 +1,114 @@
...
+    p.send(incomplete)
+    p.expect_exact(incomplete)
+    p.send(TAB)
+    if complete.startswith(incomplete):
```
jsarzy, line 39: Do we need this check here? This function is only used in this testcase and we can assume that caller will pass reasonable arguments.

psh/test-autocompletion.py
```diff
@@ -0,0 +1,114 @@
...
+        addition = re.sub('^' + incomplete, '', complete)
```
jsarzy, line 40:
```suggestion
        suffix = word[len(prefix):]
```
Thanks to that we do not need to import `re` submodule.

psh/test-autocompletion.py
```diff
@@ -0,0 +1,114 @@
...
+        p.expect_exact(addition)
+        p.send(ENTER)
+        assert_prompt(p)
+    else:
+        raise Exception("Error in test: incomplete path not subset of complete path")
+
+
+def assert_multiple_completion(p, comptab):
+    for pair in comptab:
```
jsarzy, line 49:
```suggestion
    for prefix, suffix in comptab:
```
Could we use `assert_completion` here? I know it expects prompt at the end, but maybe some common function for prefix and suffix completion?

jsarzy (commented):
I know autocompletion is not yet merged, just few comments regarding variables naming. Overall looks good for me

psh/test-autocompletion.py
niewim19: Functions `assert_completion` and `assert_multiple_completion` unified.

psh/test-autocompletion.py
niewim19: Changed. Thanks!

psh/test-autocompletion.py
niewim19: Check removed

psh/test-autocompletion.py
niewim19: Thanks, good idea! Implemented.

niewim19:
As long as submodules will not be updated this test will fail CI. Locally tests pass on ia32.

nalajcie:
just a few things to remember: 
 - the test should clean up after itself (for now only on success unfortunately)
 - we could probably add handling of other special chars to make `psh` usage better (eg. CTRL+U should clear the whole line)
 - I've found out that autocompletion fails to work across symlink to folder, will add description to the issue (no way to test it right now as we don't have native way of symlink creation in `psh`

niewim19:
Tests changed so that they are runnable both on qemu emulator and imxrt106x target platform. Runner response after sending control characters differ between target (qemu runner quits due to not passing control characters through serial)

PR #45, opened 2021-08-05
libc: add exec functions tests

damianloew:
JIRA: PD-175

add exec functions tests
## Description

- create directory harness for harness tests in libc directory
- create text_exec.c, which executes itself using execve, execvpe, and execvp (different test cases)
- create harness test_exec.py, which executes test_exec.c with individual test cases
- add assert_expected function to psh/tools/basic.py

## Motivation and Context

When running busybox tests, there was a suspicion that the execve function was not working properly
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic - qemu).

## Checklist:

## Special treatment

### DIFF
--- a/libc/harness/test_exec.c
+++ b/libc/harness/test_exec.c
@@ -1,0 +1,204 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libc-tests
+ *
+ * executable program for various exec functions
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Damian Loewnau
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <ctype.h>
+#include <unistd.h>
+#include <string.h>
+#include <libgen.h>
+#include <errno.h>
+
+extern char **environ;
+
+
+static void test_exec_execveEnv(bool changeEnv)
+{
+	char *const arg[] = { "/bin/to_exec", NULL };
+	if (changeEnv) {
+		char *const env[] = { "TEST1=exec_value", NULL };
+		setenv("TEST1", "invalid_value", 1);
+		setenv("TEST2", "should_dissapear", 1);
+
+		if (execve("/bin/test_exec", arg, env) == -1) {
+			fprintf(stderr, "execve function failed: %s\n", strerror(errno));
+
+			exit(EXIT_FAILURE);
+		}
+	}
+	else {
+		setenv("TEST1", "unchanged_value", 1);
+
+		if (execve("/bin/test_exec", arg, environ) == -1) {
+			fprintf(stderr, "execve function failed: %s\n", strerror(errno));
+
+			exit(EXIT_FAILURE);
+		}
+	}
+}
+
+
+static void test_exec_execvePath(void)
+{
+	char *const arg[] = { "to_exec", NULL };
+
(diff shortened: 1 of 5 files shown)


libc/harness/test_exec.c
```diff
@@ -0,0 +1,201 @@
...
+static void test_exec_execveEnv(bool changeEnv)
+{
+	char *const arg[] = { "/bin/to_exec", NULL };
+	if (changeEnv) {
+		char *const env[] = { "TEST1=exec_value", NULL };
+		setenv("TEST1", "invalid_value", 1);
+		setenv("TEST2", "should_dissapear", 1);
+
+		if (execve("/bin/test_exec", arg, env) == -1) {
+			fprintf(stderr, "execve function failed\n");
```
nalajcie, line 36: maybe use `perror` or `strerror` to inform about errno encountered?

libc/harness/test_exec.c
```diff
@@ -0,0 +1,201 @@
...
+
+	if (execvp("test_exec", arg) == -1) {
+		fprintf(stderr, "execve function failed\n");
+
+		exit(EXIT_FAILURE);
+	}
+}
+
+
+int main(int argc, char *argv[], char const *envp[])
```
nalajcie, line 135: it would probably be better to use `environ` variable instead of third parameter to `main()` as the first is POSIX-compliant and second is a GNU GCC extension (widely used but even so.. :)).

libc/harness/test_exec.c
```diff
@@ -0,0 +1,201 @@
...
+				test_exec_execvpePath();
+				break;
+
+			case 7:
+				test_exec_execvpEnv();
+				break;
+
+			case 8:
+				test_exec_execvpPath();
+				break;
```
nalajcie, line 193: I would also add here some tests for:
 - `execl`, `execlp`, `execle`
 - passing non-trivial arguments (more than one :))
 - (in the future) execs with interpreter (so a script with a she-bang)
 - intentionally failing cases (and verify the `errno`) - eg. missing file, trying to exec directory, ENOMEM would be probably hard to get :)

libc/harness/test_exec.py
```diff
@@ -0,0 +1,109 @@
...
+                r'(\r\x1b\[0J\(psh\)%)')
+
+    assert_expected(p, cmd, expected, "Wrong output of execve function with changed environment")
+
+
+def assert_execve_env_unchanged(p):
+    cmd = '/bin/test_exec 2'
+    expected = (r'(argc = 1(\r+)\n)'
+                r'(argv\[0\] = /bin/to_exec(\r+)\n)'
+                r'(environ\[0\] = TEST1=unchanged_value(\r+)\n)'
```
nalajcie, line 32: this assumes we will have empty ENV when starting this test (which might not be the case), Maybe it would be better just to check if `TEST1=unchanged_value` exists somewhere in the environment? (so - more complicated regexp).

The alternative approach is to use `clearenv()` explicitly in C code.

nalajcie (commented):
The tests are good and I see they are passing. I've added some small remarks and ideas for other exec-tests (the new test cases can be done either as a part of this PR or as a separate one after merging this on :+1:)

libc/harness/test_exec.py
damianloew: Thanks for review, changes applied. I will add more test cases in the future.

nalajcie: ok, please copy my ideas for new tests to JIRA issue not to forget them in the future :).

nalajcie (approved):
LGTM.

PR #48, opened 2021-09-07
test

damianloew:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


trunner/device.py
```diff
@@ -18,6 +18,7 @@
 
 
 _BOOT_DIR = PHRTOS_PROJECT_DIR / '_boot'
+LOGFILE_PATH = '/actions-runner/logfile'
```
nalajcie, line 21: this is another change which prevents using `trunner` locally!

**to be removed ASAP!**
(as this current PR is a stop-gap, leave it as-is for now, but start thinking how to fix that in the near future)

Please also note that the file would be removed with the container (eg. after docker image update)

trunner/device.py
```diff
@@ -35,6 +36,15 @@ def is_github_actions():
     return os.getenv('GITHUB_ACTIONS', False)
 
 
+def save_in_logfile(threshold_exceeded=False, wait_time=0):
+    if not os.path.exists(LOGFILE_PATH):
+        os.system(f'echo "Does wait time exceed 5s?" > {LOGFILE_PATH}')
```
nalajcie, line 41: no, please open file in python and write directly instead of unnecessary process forking here.
(**this needs to be fixed before the merge**)

I would (always) store raw restart times in file.csv to be able to compute max/min/mean/avg and quickly plot graph from it, the file doesn't need to be human-readable.

trunner/device.py
```diff
@@ -44,6 +52,7 @@ def wait_for_dev(port, timeout=0):
         asleep += 0.01
         if timeout and asleep >= timeout:
             raise TimeoutError
+        save_in_logfile(wait_time=asleep)
```
nalajcie, line 55: definitely should not be inside `while`

nalajcie (changes requested):

trunner/device.py
nalajcie: ok, let's leave it for now. Side remark: `logfile` is a too-generic name (this is not a generic log, You're logging only reboot timeouts here)

PR #50, opened 2021-09-13
test-libc: add getpwuid/getpwnam tests

niewim19:
## Description

Simple `getpwuid()` and `getpwnam()` functional tests

## Motivation and Context

Tests if `getpwuid` and `getpwnam` returnes correctly for `root` user and if NULL is returned for flawed values. These tests should be later extended for input length edge cases and `getpwuid_r` and `getpwnam_r' when implemented

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic)

## Checklist:

## Special treatment


niewim19:
Tests fail due to current, wrong getpwuid implementation. On local machine it is:
Screenshot from 2021-09-13 14-51-31

niewim19:
After submodules update tests pass

libc/getpw.c
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Mateusz Niewiadomski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+/* make compilable against glibc (all these tests were run on host against libc/linux kernel) */
```
nalajcie, line 16: is this true or was it copied from the other C file?

(I believe it's false as the consts You're checking against are not the defaults found on linux)

libc/getpw.c
```diff
@@ -0,0 +1,111 @@
...
+static const char root[] = "root";
+static struct passwd *pw;
+
+
+TEST_GROUP(getpwd);
+
+
+TEST_SETUP(getpwd)
+{
+	ispasswdfile = (access("/etc/passwd", F_OK) == 0) ? 1 : 0;
```
nalajcie, line 35: You know that this gets called before every test? Maybe You should do it once in `TEST_GROUP_RUNNER` - and depending on the result either run every test with result "SKIPPED" or don't run any tests at all.

libc/getpw.c
```diff
@@ -0,0 +1,111 @@
...
+		TEST_ASSERT_EQUAL_STRING("root", pw->pw_name);
+		/* no password hash checking */
+		TEST_ASSERT_EQUAL_INT(0, pw->pw_uid);
+		TEST_ASSERT_EQUAL_INT(0, pw->pw_gid);
+		TEST_ASSERT_EQUAL_STRING("root", pw->pw_gecos);
+		TEST_ASSERT_EQUAL_STRING("/", pw->pw_dir);
+		TEST_ASSERT_EQUAL_STRING("/bin/sh", pw->pw_shell);
+	}
+	else {
+		TEST_ASSERT_NULL(pw);
```
nalajcie, line 60: I don't like that we're marking this test as `PASSED` even if it doesn't check anything relevant.

Maybe we should try exiting with `TEST_IGNORE()` - it might be visible as a skipped one then (or don't run the test at all - as described above).

libc/getpw.c
```diff
@@ -0,0 +1,111 @@
...
+
+#include <pwd.h>
+#include <unistd.h>
+
+#include <unity_fixture.h>
+
+
+static unsigned int ispasswdfile;
+static const char root[] = "root";
+static struct passwd *pw;
```
nalajcie, line 27: pw should be a local variable in every test

libc/getpw.c
```diff
@@ -0,0 +1,111 @@
...
+
+/* Get root account details by uid */
+TEST(getpwd, getpwuid_getroot)
+{
+	pw = getpwuid((uid_t)0);
+	if (ispasswdfile) {
+		TEST_ASSERT_NOT_NULL(pw);
+
+		TEST_ASSERT_EQUAL_STRING("root", pw->pw_name);
+		/* no password hash checking */
```
nalajcie, line 72: You could probably check at least if `pw_passwd` pointer is non-null (could that even happen?)

libc/getpw.c
```diff
@@ -0,0 +1,111 @@
...
+#define _GNU_SOURCE
+
+#include <pwd.h>
+#include <unistd.h>
+
+#include <unity_fixture.h>
+
+
+static unsigned int ispasswdfile;
+static const char root[] = "root";
```
nalajcie, line 26: don't see any real benefit for specifying`root` const here as You hardcode every other value below.

libc/getpw.c
```diff
@@ -0,0 +1,111 @@
...
+	else {
+		TEST_ASSERT_NULL(pw);
+	}
+}
+
+/* Try to reach for non-existing user data with getpwnam */
+TEST(getpwd, getpwnam_getnull)
+{
+	pw = getpwnam("loremipsum");
+	TEST_ASSERT_NULL(pw);
```
nalajcie, line 88: maybe we should check if resulting `errno` code is correct here (that the entry was not found, not there was `ENOMEM` or other error involved?)

nalajcie (changes requested):
Pointed out some issues - the most pressing one for me is that we mark test as `PASSED` when we don't actually test `getpw*` functions which is really misleading.

libc/getpw.c
niewim19: Yes you are right. Leftover copied from other file. I will remove that.

libc/getpw.c
niewim19: Thanks for that idea. Moved the `ispasswdfile` flag initialization to `TEST_GROUP_RUNNER`

libc/getpw.c
niewim19: I propose the middle path, as if there is no `/etc/passwd` file in the filesystem we should still check if the functions behave accordingly. So I propose to:
 - try to execute `getpwuid`/`getpwnam` no matter what
 - if there is no `/etc/passwd` check if `pw` is NULL
 - only if NULL check passes - ignore the test

Additionally add tests: `getpwuid_nopasswdfile` and `getpwnam_nopasswdfile` as this is valid test scenario

libc/getpw.c
niewim19: Changed, thanks!

libc/getpw.c
niewim19: change for checking nullability and then length

libc/getpw.c
niewim19: Removed, thanks!

libc/getpw.c
niewim19: This would be problematic as current getpw* functions do not setup errno by itself ( should at least force errno 0 or ENOENT if nothing was found)v. I will take care of that and then correct tests later. Thanks!

PS. When `getpw*` will setup correct errno the `check_null_and_errno` from `common` may be used.

niewim19:
### Changed:

 - added `getpwnam_nopasswdfile` and `getpwuid_nopasswdfile`:
   - testing `getpwuid` and `getpwnam` with no `/etc/passwd` file
   - if there is no such file - just performs getpwuid and checks `NULL`
   - if there is such file - temporarily renames it
 - some tests performs only partially and then are labeled as ignored if there is no `/etc/passwd` file
 - code cleanup according to review

### Unchanged:
 - no `errno` checking as current implementations of `getpw*`  do not properly set `errno` (only subfunctions used by them). Test will be updated together with changes to `getpw*`

agkaminski (approved):
LGTM - not much more to do if we don't have /etc/passwd imho

libc/getpw.c
```diff
@@ -0,0 +1,150 @@
...
+	if (!ispasswdfile)
+		TEST_IGNORE_MESSAGE("No /etc/passwd file!");
+}
+
+
+/* Try to reach for root user data with no /etc/passwd file with getpwnam */
+TEST(getpwd, getpwnam_nopasswdfile)
+{
+	if (ispasswdfile)
+		rename("/etc/passwd", "/etc/passwd_del");
```
nalajcie, line 117: note: this may fail on targets which will have readonly root partition. I don't know what should be done in that case (should we ignore the test then?). No need to change anything right now, just keep it in mind in the future.

nalajcie (approved):
LGTM.

PR #54, opened 2021-09-28
string.h: strlcat() and strlcpy() tests

niewim19:
JIRA: PD-189

## Description

Testing `strlcat` and `strlcpy` implementations

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (local tests testing on ia320generic qemu).

## Checklist:

## Special treatment

- [x] I will merge this PR by myself when appropriate. This PR is scheduled to merge **after**:
   - https://github.com/phoenix-rtos/libphoenix/pull/164

### DIFF
--- a/libc/string.c
+++ b/libc/string.c
@@ -1,0 +1,189 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libc-tests
+ *
+ * Testing string.h functions
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Mateusz Niewiadomski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <string.h>
+
+#include <unity_fixture.h>
+
+#define SRC_SIZE  5
+#define DEST_SIZE 5
+
+#define SRC1_SIZE 4
+#define SRC2_SIZE 6
+#define BUF_SIZE  12
+
+
+/* fixtures for strlcpy */
+char source[SRC_SIZE];
+char dest[DEST_SIZE];
+
+/* fixtures for strlcat */
+char source1[SRC1_SIZE];
+char source2[SRC2_SIZE];
+char buffer[BUF_SIZE];
+
+/* common fixtures */
+int retval;
+
+
+TEST_GROUP(string_strlcpy);
+
+
+TEST_SETUP(string_strlcpy)
+{
+	memcpy(source, "abcd", SRC_SIZE);
+	memcpy(dest, "xxxx", DEST_SIZE);
+}
+
+
+TEST_TEAR_DOWN(string_strlcpy)
+{
+}
+
+
+TEST(string_strlcpy, strlcpy_fullcopy)
+{
+	/* Test full copy */
+	retval = strlcpy(dest, source, sizeof(source));
+	TEST_ASSERT_EQUAL_INT(sizeof(source) - 1, retval);
+	TEST_ASSERT_EQUAL_STRING(source, dest);
+}
+
+
+TEST(string_strlcpy, strlcpy_shorter)
+{
+	/* Test shorter than source copy */
+	retval = strlcpy(dest, source, sizeof(source) - 2);
+	TEST_ASSERT_EQUAL_STRING("ab", dest);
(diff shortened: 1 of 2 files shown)


libc/string.c
```diff
@@ -0,0 +1,184 @@
...
+
+TEST_TEAR_DOWN(string_strlcpy)
+{
+}
+
+
+TEST(string_strlcpy, strlcpy_fullcopy)
+{
+	/* Test full copy */
+	i = strlcpy(dest, source, 5);
```
agkaminski, line 53:
```suggestion
	i = strlcpy(dest, source, sizeof(dest));
```

libc/string.c
```diff
@@ -0,0 +1,184 @@
...
+	TEST_ASSERT_EQUAL_INT(4, i);
```
agkaminski, line 54:
```suggestion
	TEST_ASSERT_EQUAL_INT(sizeof(source) - 1, i);
```

libc/string.c
```diff
@@ -0,0 +1,184 @@
...
+	TEST_ASSERT_EQUAL_STRING(source, dest);
+}
+
+
+TEST(string_strlcpy, strlcpy_shorter)
+{
+	/* Test shorter than source copy */
+	i = strlcpy(dest, source, 3);
+	TEST_ASSERT_EQUAL_STRING("ab", dest);
+	TEST_ASSERT_EQUAL_INT(4, i);
```
agkaminski, line 64:
```suggestion
	TEST_ASSERT_EQUAL_INT(sizeof(source) - 1, i);
```

libc/string.c
```diff
@@ -0,0 +1,184 @@
...
+}
+
+
+TEST(string_strlcpy, strlcpy_longer)
+{
+	/* Test longer than source copy */
+	source[3] = '\0'; /* source is now "abc" null terminated; */
+	i = strlcpy(dest, source, 5);
```
agkaminski, line 72:
```suggestion
	i = strlcpy(dest, source, sizeof(dest));
```

libc/string.c
agkaminski: etc for other cases

libc/string.c
```diff
@@ -0,0 +1,184 @@
...
+char dest[5];
+char tmplt[5];
+
+/* fixtures for strlcat */
+char source1[4];
+char source2[6];
+char buffer[12];
+
+/* common fixtures */
+int i;
```
agkaminski, line 31: Common iterator name for global variable - I'm not sure if I like it (I think I don't). Maybe `retval`?

libc/string.c
```diff
@@ -0,0 +1,184 @@
...
+
+TEST_GROUP(string_strlcat);
+
+
+TEST_SETUP(string_strlcat)
+{
+	memcpy(source1, "abc", 4);
+	memcpy(source2, "defgh", 6);
+	memcpy(buffer, "klmnopqrstu", 12);
+	i = -1;
```
agkaminski, line 115: I don't think we need this - it's always assigned by each test

libc/string.c
agkaminski: Or maybe it should be a local variable (along with the name change)

agkaminski (commented):
My comments about `sizeof` use are somewhat optional - it would make test more easily changeable in the future and also point out from where this value comes from - right now these are somewhat magic-numbery.

libc/string.c
niewim19: Variable left as global. Name changed to `retval` as suggested. Thanks!

libc/string.c
niewim19: Yes, you are right. Removed `i = -1`. Thanks!

niewim19:
@agkaminski Thanks for review, all suggestions added, used `sizeof()` with correct variable where necessary.

libc/string.c
```diff
@@ -0,0 +1,183 @@
...
+int retval;
+
+
+TEST_GROUP(string_strlcpy);
+
+
+TEST_SETUP(string_strlcpy)
+{
+	memcpy(source, "abcd", sizeof(source));
+	memcpy(dest, "xxxx", sizeof(dest));
```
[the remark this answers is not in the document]
niewim19: Used `#defines` with sizes, thanks!

PR #61, opened 2021-10-19
runner: imxrt106x: switch to restart by jtag

damianloew:
JIRA: CI-88

## Description

- add is_cut_power_used attribute to IMXRT106xRunner class
- set default restart by jtag use
## Motivation and Context

In one of the test units some problem with flashing board has been occurred (when using restart by power off). Otherwise restart by jtag would be a better solution for fast ci (triggered by pr's and merges to master), and restart by power off would be better for long tests scheduled for night hours.
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt106x eval kit).

## Checklist:

## Special treatment

### DIFF
--- a/trunner/device.py
+++ b/trunner/device.py
@@ -381,10 +381,12 @@ class IMXRT106xRunner(DeviceRunner):
     def __init__(
         self,
         port,
-        phoenixd_port='/dev/serial/by-id/usb-Phoenix_Systems_plo_CDC_ACM-if00'
+        phoenixd_port='/dev/serial/by-id/usb-Phoenix_Systems_plo_CDC_ACM-if00',
+        is_cut_power_used=False
     ):
         super().__init__(port)
         self.phoenixd_port = phoenixd_port
+        self.is_cut_power_used = is_cut_power_used
         self.reset_gpio = GPIO(17)
         self.reset_gpio.high()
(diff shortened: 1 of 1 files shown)


trunner/device.py
```diff
@@ -463,14 +463,14 @@ def flash(self):
...
-        self.reboot(cut_power=True)
+        self.reboot(cut_power=False)
 
     def load(self, test):
         """Loads test ELF into syspage using plo"""
 
         phd = None
         load_dir = str(rootfs(test.target) / 'bin')
-        self.reboot(cut_power=True)
+        self.reboot(cut_power=False)
```
nalajcie, line 473: if we're not cutting power we probably don't need the second reboot below

trunner/device.py
```diff
@@ -441,7 +441,7 @@ def reboot(self, serial_downloader=False, cut_power=False):
             self._restart_by_jtag()
 
     def flash(self):
-        self.reboot(serial_downloader=True, cut_power=True)
+        self.reboot(serial_downloader=True, cut_power=False)
```
nalajcie, line 444: maybe instead of changing the parameters in multiple places, we could use instance attribute and change the parameters / code flow accordingly?

PR #64, opened 2021-11-05
amrv7m4-stm32l4 test runner implementation

niewim19:
## Description

This is a test runner implementation for armv7m4-stm32l4 target. It was developed for and tested on [`NUCLEO-L4A6ZG`](https://www.st.com/en/evaluation-tools/nucleo-l4a6zg.html#overview) board

Main differences from other runners:
 - because of slow speed of communication over serial port there is a timeout after each letter send via harness. It is achieved by redefined `send()` and `sendline()` in `oneByOne_fdspawn` nested class in `Stm32l4Runner`
 - it uses openocd to flash image onto target via predefined command in `Stm32l4Runner`
 - additional `proc.send("\n")` (newline) is send to the board to ensure that psh expectance by runner is met.

## Motivation and Context

Expanding the Phoenix-RTOS testing targets universe

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: stm32l4.

## Checklist:

- [x] My change requires a change to the documentation.

## Special treatment


trunner/test_runner.py
```diff
@@ -37,7 +38,7 @@ def parse_tests(self):
             logging.debug(f"File {path} parsed successfuly\n")
 
     def run(self):
-        self.runners = {target: RunnerFactory.create(target) for target in self.targets}
+        self.runners = {target: RunnerFactory.create(target=target, serial=self.serial) for target in self.targets}
```
nalajcie, line 41: seems invalid way of passing serial device. Right now we will create different runners using the same serial communication...

trunner/testcase.py
```diff
@@ -192,6 +199,7 @@ def __init__(
     ):
         super().__init__(name, target, timeout, exec_cmd, use_sysexec, status)
         self.load_module(harness_path)
+        self.type = 'harness'
```
nalajcie, line 202: **if** we would ever want to clearly specify the test type, we would like to use some enumerated type - not a string constant (which can be easily mistyped)

trunner/testcase.py
```diff
@@ -219,6 +227,7 @@ def __init__(
         super().__init__(name, target, timeout, exec_cmd, use_sysexec, status)
         self.harness = UnitTestHarness.harness
         self.unit_test_results = []
+        self.type = 'unit'
```
nalajcie, line 230: please note that this could be a class const, not an instance variable (it just describes the class)

trunner/testcase.py
```diff
@@ -39,10 +39,14 @@ def __init__(
         self.status = status
         self.harness = None
         self.exception = ''
+        self.type = None
 
     def fail(self):
         self.status = TestCase.FAILED
 
+    def skip(self):
```
nalajcie, line 47: deciding which test to run should be done only by inspecting YAML files.

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+import serial
+import time
+import pexpect
+import pexpect.fdpexpect
+
+from .common import DeviceRunner
+from .common import _BOOT_DIR
+
+
+class Stm32l4Runner(DeviceRunner):
```
nalajcie, line 19: maybe `STM32L4Runner` as lowercase `l` looks like `1`

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+# Stm32l4 runner
+#
+# Copyright 2021 Phoenix SYstems
+# Authors: Mateusz Niewiadomski
+#
+import subprocess
+import serial
+import time
+import pexpect
+import pexpect.fdpexpect
```
nalajcie, line 13: unneeded import?

trunner/device.py
```diff
@@ -18,12 +18,14 @@
 
 class RunnerFactory:
     @staticmethod
-    def create(target):
+    def create(target=None, serial=None):
```
nalajcie, line 21: target should not have default value, it is explicitly needed when calling this method

trunner/config.py
```diff
@@ -30,7 +30,7 @@ def resolve_phrtos_dir() -> Path:
 PYEXPECT_TIMEOUT = 8
 
 # Available targets for test runner.
-ALL_TARGETS = ['ia32-generic', 'host-pc', 'armv7m7-imxrt106x']
+ALL_TARGETS = ['ia32-generic', 'host-pc', 'armv7m7-imxrt106x', 'armv7m4-stm32l4']
```
nalajcie, line 33: maybe we should remove `armv7m4-stm32l4` from `DEFAULT_TARGETS` if not all tests are runnable right away (I assume this will change in the near future)

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+
+    OPENOCD_CMD = ['openocd -f interface/stlink.cfg -f target/stm32l4x.cfg -c \
+"reset_config srst_only srst_nogate connect_assert_srst" -c "program ', ' 0x08000000 verify reset exit"']
+    IMAGE = 'phoenix-armv7m4-stm32l4x6.bin'
+
+    def __init__(self, port1, port2=None):
+        super().__init__(port1)
+        self.port2 = port2
+        self.image_flashed = False
+        self.openocd_cmd = self.OPENOCD_CMD[0] + str(_BOOT_DIR / self.IMAGE) + self.OPENOCD_CMD[1]
```
nalajcie, line 40: nah, the filepath is not correctly escaped (if there is a space in the path it would fail).

I would go for providing arguments as an array to `subprocess.Popen()`

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+            shell=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE)
+
+        stdout, stderr = openocd_process.communicate()
+        openocd_process.wait()
+        # Chcecking openocd flashing result
+        verify1 = str(stderr).find('** Verified OK **')
+        verify2 = str(stderr).find('shutdown command invoked')
+        assert verify1 >= 0 and verify2 >= 0, "Stm32l4Runner: OpenOCD failed to flash target"
```
nalajcie, line 55: I assume the `openocd` doesn't report the status by exit value (that would be better to check)

Usability note: maybe we should output logs from flashing for user to see (similarly to rt1064)?

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+
+from .common import DeviceRunner
+from .common import _BOOT_DIR
+
+
+class Stm32l4Runner(DeviceRunner):
+
+    class oneByOne_fdspawn(pexpect.fdpexpect.fdspawn):
+
+        def send(self, s):
```
nalajcie, line 23: invalid implementation, the `send()` should return number of bytes it sent, see the original class

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+            for c in s:
+                super().send(c)
+                time.sleep(0.03)
+
+        def sendline(self, s):
```
nalajcie, line 28: should not be needed, the original implementation uses send() internally also

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+            return
+
+        try:
+            self.serial = serial.Serial(self.port, baudrate=115200)
+        except serial.SerialException:
+            test.handle_exception()
+            return
+
+        proc = self.oneByOne_fdspawn(self.serial, encoding='ascii', timeout=test.timeout)
+        proc.send("\n")
```
nalajcie, line 72: Why this is needed? It needs clear explanation in the code (probably with FIXME as it should not be needed anyway)

nalajcie (changes requested):
Pointed out some issues, the code clearly needs to be better documented before the merge. 

Some things need to be changed as specified; I don't know the full architecture of the solution, maybe @jsarzy would provide more insight.

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class Stm32l4Runner(DeviceRunner):
+
+    class oneByOne_fdspawn(pexpect.fdpexpect.fdspawn):
+
+        def send(self, s):
+            for c in s:
+                super().send(c)
+                time.sleep(0.03)
```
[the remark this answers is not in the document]
niewim19: Yes it is related to our stm32l4 targets not processing characters fast enough.

Comments with relevant information added.

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+
+        def sendline(self, s):
+            self.send(s)
+            super().send('\n')
+
+    OPENOCD_CMD = ['openocd -f interface/stlink.cfg -f target/stm32l4x.cfg -c \
+"reset_config srst_only srst_nogate connect_assert_srst" -c "program ', ' 0x08000000 verify reset exit"']
+    IMAGE = 'phoenix-armv7m4-stm32l4x6.bin'
+
+    def __init__(self, port1, port2=None):
```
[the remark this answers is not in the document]
niewim19: changed names to `OPENOCD_CMD` and `BINARY_PATH`

trunner/runners/Stm32l4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+        # Chcecking openocd flashing result
+        verify1 = str(stderr).find('** Verified OK **')
+        verify2 = str(stderr).find('shutdown command invoked')
+        assert verify1 >= 0 and verify2 >= 0, "Stm32l4Runner: OpenOCD failed to flash target"
+
+    def run(self, test):
+        if test.skipped():
+            return
+
+        if not test.is_type('harness'):
```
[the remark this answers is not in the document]
niewim19: Yes you are right. This is not acceptable. Removed all `is_type` logic as stm32l4 runner as it may cause more chaos

trunner/test_runner.py
niewim19: Yes, but the same **default** serial communication. This PR features fix for not processing `--serial` argument correctly so now any port can be passed instead on the fly.

trunner/testcase.py
niewim19: solved with `is_type` logic removed

trunner/testcase.py
niewim19: solved with `is_type` logic removed

trunner/testcase.py
niewim19: solved with `is_type` logic removed

trunner/runners/Stm32l4Runner.py
niewim19: Good advice. Changed target name to uppercase (only not in ` message where there is a convention of full lowercase). Thanks!

trunner/runners/Stm32l4Runner.py
niewim19: yes indeed, removed.

trunner/device.py
niewim19: Corrected, thanks

trunner/config.py
niewim19: Good point. Removed

trunner/runners/Stm32l4Runner.py
niewim19: I miss the point that return value can be read from `subprocess.wait()`. 

Added that functionality. It works well, returning `!0` if no target present, no stlink present, or target present but flashing failed

trunner/runners/Stm32l4Runner.py
niewim19: Yes, my mistake. Added counter and return

trunner/runners/Stm32l4Runner.py
niewim19: Yes this is redundant redeclaration. Removed `sendline()` from `oneByOne_fdspawn` and it works well

trunner/runners/Stm32l4Runner.py
niewim19: Added FIXME as this is indeed a workaround

trunner/runners/Stm32l4Runner.py
niewim19: used:
```
BINARY_PATH = path.join(_BOOT_DIR ,'phoenix-armv7m4-stm32l4x6.bin')
```

passing arguments to `Popen()` as array seems to be problematic as there are `"` and for some reason escaping them or not makes no difference even using `shlex`

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def __init__(self, port1):
+        super().__init__(port1)
+
+    def flash(self):
+        # Flashing with openocd as a separate process
+        command = self.OPENOCD_CMD[0] + self.BINARY_PATH + self.OPENOCD_CMD[1]
+        openocd_process = Popen(
+            command,
+            shell=True,
```
niewim19, line 51: This is one thing that bothers me, as this is often described as vulnerability as it allows execute shell comands as scripts so if anyone creates a PR where this command will be injected with `rm -rf *` would be catastrophic. But without it paths are not resolved correctly

niewim19:
Code cleared after @nalajcie review. 

Biggest changes:
 - Removed `is_type` logic. It was a shady workaround which at this point only _made more code_ (and runner is still not fully usable e.g hard to implement integration of additional binaries in syspage
 - added necessary commands so the runner (designed around quirks of stm32l4 target) is more reader-friendly
 - changed flash status check to actually checking `openocd` return code (insted of output parsing)

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+            stdout=PIPE,
+            stderr=sys.stdout.fileno()
+            )
+        assert openocd_process.wait() == 0, "Stm32l4Runner: OpenOCD failed to flash target"
```
jsarzy, line 55: Maybe some nice print with error message and then just exit with proper code?

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, port1):
+        super().__init__(port1)
+
+    def flash(self):
+        # Flashing with openocd as a separate process
+        command = self.OPENOCD_CMD[0] + self.BINARY_PATH + self.OPENOCD_CMD[1]
+        openocd_process = Popen(
+            command,
+            shell=True,
+            stdout=PIPE,
```
jsarzy, line 52: we don't want to catch openocd stdout? asking - I don't know what openocd prints

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+            for c in s:
+                r += super().send(c)
+                time.sleep(0.03)
+            return r
+
+    # STM32L4Runner class variables:
+    # - BINARY_NAME: name of Phoenix-RTOS image binary that should be available in _boot/ directory
+    # - OPENOCD_CMD: string with shell command calling openocd with correct parameters to flash target
+    OPENOCD_CMD = ['openocd -f interface/stlink.cfg -f target/stm32l4x.cfg -c \
+"reset_config srst_only srst_nogate connect_assert_srst" -c "program ', ' 0x08000000 verify reset exit"']
```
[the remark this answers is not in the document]
niewim19: The problem is that this command should include a path to binary here:
```
(...) -c "program <PATH_TO_BINARY> 0x08000000 verify (...)
```
and with that it breaks no matter what I do:
 - insert path variable as list argument
 - insert path variable by `'(...)'.format()` method
 - insert path directly into the string as plain text

I think the issue is that openocd interprets commandline arguments in some strange way. List-ifying the command breaks the `openocd` interpreting it. I have tried also making arguments like:
```
'-f interface/stlink.cfg', '-f'
```
instead of 
```
'-f', 'interface/stlink.cfg', '-f'
```
but it just produces different errors (like `openocd` cant figure out where it is launched from as it cannot find config files).

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,77 @@
...
+    BINARY_PATH = path.join(_BOOT_DIR, 'phoenix-armv7m4-stm32l4x6.bin')
+
+    def __init__(self, port1):
+        super().__init__(port1)
```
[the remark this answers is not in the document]
niewim19: Yes, it is a leftover. Constructor removed

trunner/runners/STM32L4Runner.py
niewim19: Good idea. I added nicer error handling here

trunner/runners/STM32L4Runner.py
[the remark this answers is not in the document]
nalajcie: stil, it seems You're doing something wrong, as the approach suggested by @jsarzy is the correct one 9albeit there are some errors - the parameters should not be in double `"` as this is parameter encasing for shell only) . Please share the exact code You're trying out so we could spot potential errors.


I've quickly verified that code like this works:
```python
CMD = ['openocd', '-f', 'interface/stlink.cfg', '-f', 'target/stm32l4x.cfg', '-c', "reset_config srst_only srst_nogate connect_assert_srst",  '-c',  "program {progname} 0x08000000 verify reset exit"]
cmd_run = [x.format(progname='real_program_path') for x in CMD]
Popen(cmd_run)
```

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,84 @@
...
+from os import path
+from subprocess import Popen, PIPE
+
+from .common import DeviceRunner, Color
+from .common import _BOOT_DIR
+
+
+class STM32L4Runner(DeviceRunner):
+
+    # oneByOne_fdspawn class is a workaround for Phoenix-RTOS on stm32l4 targets not processing characters fast enough
```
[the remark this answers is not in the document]
niewim19: Thanks! Added docstrings command to `STM32L4Runner` and `oneByOne_fdspawn` classes and `flash()` method

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,87 @@
...
+        """ Workaround for Phoenix-RTOS on stm32l4 targets not processing characters fast enough
+            This class inherits and is passed to harness instead of pexpect.fdpexpect.fdspawn
+            It redefines send() with addition of timeout after each letter sent. Timeout set to 30ms per character"""
+
+        def send(self, s):
+            r = 0
+            for c in s:
+                r += super().send(c)
+                time.sleep(0.05)
+            return r
```
damianloew, line 37: I suggest more readable names for variables here

trunner/runners/STM32L4Runner.py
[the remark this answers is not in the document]
niewim19: Yes it was a problem. Removing `"` from arguments made all work as it should without `shell=True` parameter. Thanks!

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,87 @@
...
+# Authors: Mateusz Niewiadomski
+#
+
+import serial
+import time
+import pexpect.fdpexpect
+
+import sys
+from os import path
+from subprocess import Popen, PIPE
```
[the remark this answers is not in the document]
niewim19: Changed to module imports and separated built-in from external modules

trunner/runners/STM32L4Runner.py
niewim19: Changed variable names to more descriptive

niewim19:
Main changes after reviews:
 - use list structured command in flashing in `subprocess`
 - removed `shell=True`
 - add nicer error management if `subprocess` fails at flashing
 - minor code cleanup

niewim19:
Timeout in `oneByOne_fdspawn` was brought back down to 30ms

jsarzy (commented):
Some small remarks and I think it will be ready to merge

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,87 @@
...
+        openocd_process = subprocess.Popen(
+            openocd_cmd,
+            stdout=subprocess.PIPE,
+            stderr=sys.stdout.fileno()
+            )
+        try:
+            assert openocd_process.wait() == 0
+        except AssertionError:
+            print(Color.colorify("OpenOCD error: cannot flash target\n", Color.BOLD))
+            exit(1)
```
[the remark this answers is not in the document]
niewim19: Thanks, Removed unnecessary `try/except` and used `if`

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,87 @@
...
+        openocd_cmd = [
+            'openocd',
+            '-f', 'interface/stlink.cfg',
+            '-f', 'target/stm32l4x.cfg',
+            '-c', "reset_config srst_only srst_nogate connect_assert_srst",
+            '-c',  "program {progname} 0x08000000 verify reset exit".format(progname=binary_path)
+            ]
+        openocd_process = subprocess.Popen(
+            openocd_cmd,
+            stdout=subprocess.PIPE,
```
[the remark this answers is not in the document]
niewim19: - Added suitable comment about `openocd` and `stderr`.
 - Piped both `stdout` and `stderr` to main process `stdout` just for consistency

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,87 @@
+#
+# Phoenix-RTOS test runner
+#
+# Stm32l4 runner
+#
+# Copyright 2021 Phoenix SYstems
```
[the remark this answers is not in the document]
niewim19: COrrected :)

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,87 @@
...
+
+class STM32L4Runner(DeviceRunner):
+    """ This class provides interface to run test case on STM32L4 target using RaspberryPi.
+        The ST-Link programming board should be connected to USB port
+        and a target should be connected to ST-Link and powered"""
+
+    class oneByOne_fdspawn(pexpect.fdpexpect.fdspawn):
+        """ Workaround for Phoenix-RTOS on stm32l4 targets not processing characters fast enough
+            This class inherits and is passed to harness instead of pexpect.fdpexpect.fdspawn
+            It redefines send() with addition of timeout after each letter sent. Timeout set to 50ms per character"""
```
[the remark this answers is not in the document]
niewim19: Yes, this was omitted by me. Corrected docstring to `30ms`

anglov:
@nalajcie it is good now for merge or require more work?

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,89 @@
...
+        proc = self.oneByOne_fdspawn(
+            self.serial,
+            encoding='ascii',
+            codec_errors='ignore',
+            timeout=test.timeout
+            )
+
+        # FIXME - race on start of Phoenix-RTOS between dummyfs and psh
+        # sending newline ensures that carret is in newline just after (psh)% prompt
+        proc.send("\n")
```
jsarzy, line 84: Are you are sure it will be handled properly?
For example let's assume that there wasn't race and `psh` prompt have been printed properly, so in a buffer we have
`some drivers output\r\n(psh)%: \r\n(psh)%:`. Then if I understand correctly pexpect.expect('(psh)%:' ) in test runner will match the first one instead of second one.

niewim19: Good point. Added buffer flush before newline. I think it should be added to every runner (and firstly to _common_ module). I will take care of that

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # openocd prints important information (counterintuitively) on stderr, not stdout. However pipe both for user
+        openocd_process = subprocess.Popen(
+            openocd_cmd,
+            stdout=sys.stdout.fileno(),
+            stderr=sys.stdout.fileno()
+            )
+
+        if openocd_process.wait() != 0:
+            print(Color.colorify("OpenOCD error: cannot flash target\n", Color.BOLD))
```
nalajcie, line 61: `\n` is not needed

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,95 @@
...
+import pexpect.fdpexpect
+import serial
+from pexpect import TIMEOUT
+
+from .common import DeviceRunner, Color
+from .common import _BOOT_DIR
+
+
+class STM32L4Runner(DeviceRunner):
+    """ This class provides interface to run test case on STM32L4 target using RaspberryPi.
```
nalajcie, line 24: Note for the future: when writing docstring it's not needed to start with "This class" as I have the class name just above :).

Also, my personal preference is to format it like:

```python
def fun():
    """ single line docstring """

def fun3():
    """ Single-line summary
    
    Multiline detailed description
    with possible examples
    and description of input parameters.
    """

def fun3():
    """
    Multiline docstring.
    With many newlines, and possible indentations, eg. use examples:
        if fun3():
            return fun()
    """
```
(no extra indentation on multiline docstrings)

PEP-257 says these variants are equal so it boils down to the possible increased readability only.

See sample docstrings in bigger python projects, eg.: https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html#docstrings

nalajcie (commented):
Just some small formatting problems, otherwise LGTM :+1:.

I assume we might encounter some problems when enabling particular tests to run on `stm32l4` platform and we will solve them then :)

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,95 @@
...
+        The ST-Link programming board should be connected to USB port
+        and a target should be connected to ST-Link and powered"""
+
+    class oneByOne_fdspawn(pexpect.fdpexpect.fdspawn):
+        """ Workaround for Phoenix-RTOS on stm32l4 targets not processing characters fast enough
+            This class inherits and is passed to harness instead of pexpect.fdpexpect.fdspawn
+            It redefines send() with addition of timeout after each letter sent. Timeout set to 30ms per character
```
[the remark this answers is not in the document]
niewim19: Changed:
 - _timeout_ -> _delay_
 - _character/letter_ -> _byte_

trunner/runners/STM32L4Runner.py
niewim19: newline character removed

trunner/runners/STM32L4Runner.py
niewim19: Thank you for providing example :) 

Docstrings changed to conform with convention

trunner/runners/STM32L4Runner.py
```diff
@@ -0,0 +1,95 @@
...
+        # openocd prints important information (counterintuitively) on stderr, not stdout. However pipe both for user
+        openocd_process = subprocess.Popen(
+            openocd_cmd,
+            stdout=sys.stdout.fileno(),
+            stderr=sys.stdout.fileno()
+            )
+
+        if openocd_process.wait() != 0:
+            print(Color.colorify("OpenOCD error: cannot flash target\n", Color.BOLD))
+            exit(1)
```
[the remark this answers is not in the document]
niewim19: Changed to sys.exit(), thanks!

niewim19:
Made final editorial touches proposed by @nalajcie

nalajcie (approved):
I think we can merge this now, any problems with existing tests will be fixed in consecutive PRs.

PR #65, opened 2021-11-10
runner: add imxrt117x runner

damianloew:
## Description

- runner: add working imxrt117x runner (runner isn't available in default targets for now)
- runner: add ARMV7M7 class, common for imxrt106x and imxrt117x targets
- runner: move led method to device runner class
- runner: add status method for Runner class and restructure led control
- runner: adjust imxrt117x runner to MIMXRT1170 EVK
- tests: exclude running specific tests on imxrt117x target

## Motivation and Context

It's needed for imxrt117x rack testing unit.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt117x EVK, tested 100 times with restart by jtag and 100 times with restart by power off).

## Checklist:

## Special treatment

### DIFF
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -30,22 +30,24 @@ PHRTOS_TEST_DIR = PHRTOS_PROJECT_DIR / 'phoenix-rtos-tests'
 PYEXPECT_TIMEOUT = 8
 
 # Available targets for test runner.
-ALL_TARGETS = ['ia32-generic', 'host-pc', 'armv7m7-imxrt106x', 'armv7m4-stm32l4']
+ALL_TARGETS = ['ia32-generic', 'host-pc', 'armv7m7-imxrt106x', 'armv7m7-imxrt117x', 'armv7m4-stm32l4']
 
 # Default targets used by parser if 'target' value is absent
 DEFAULT_TARGETS = [target for target in ALL_TARGETS
                    if target not in ('host-pc', 'armv7m4-stm32l4')]
 
-DEVICE_TARGETS = ['armv7m7-imxrt106x']
+DEVICE_TARGETS = ['armv7m7-imxrt106x', 'armv7m7-imxrt117x']
 
 CURRENT_TARGET = None
 
-# Port to communicate with hardware board
+# Port to communicate with hardware boards
 DEVICE_SERIAL_PORT = "/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.1"
 DEVICE_SERIAL_BAUDRATE = 115200
 
 # DEVICE_SERIAL USB port address
 DEVICE_SERIAL_USB = "1-1.4"
+
+PHOENIXD_PORT = '/dev/serial/by-id/usb-Phoenix_Systems_plo_CDC_ACM-if00'
 
 
 class ParserError(Exception):
--- a/trunner/runners/ARMV7M7Runner.py
+++ b/trunner/runners/ARMV7M7Runner.py
@@ -1,0 +1,167 @@
+#
+# Phoenix-RTOS test runner
+#
+# ARMV7M7 runner
+#
+# Copyright 2021 Phoenix SYstems
+# Authors: Jakub Sarzyński, Damian Loewnau
+#
+
+import logging
+import time
+import sys
+import select
+
+from pexpect.exceptions import TIMEOUT, EOF
+from trunner.tools.color import Color
+from .common import Psu, Phoenixd, PhoenixdError, PloError, PloTalker, DeviceRunner, Runner
+from .common import GPIO, phd_error_msg, rootfs
+
+
+def hostpc_reboot(serial_downloader=False):
+    """Reboots a tested device by a human."""
+    if serial_downloader:
+        print("\n\tPlease change the board's boot mode to serial download, reset the board and press enter")
+    else:
+        print("\n\tPlease change the board's boot mode to internal/external flash if the current mode is different.\n\
+        Next, reset the board and press enter")
+    print("\tNote that You should press the enter key just after reset button")
+
+    if sys.stdin in select.select([sys.stdin], [], [], 30)[0]:
+        sys.stdin.readline()
+    else:
+        raise RebootError('It took too long to wait for a key pressing')
+
+
+class RebootError(Exception):
+    def __init__(self, message):
+        msg = Color.colorify("REBOOT ERROR:\n", Color.BOLD)
+        msg += str(message) + '\n'
+        super().__init__(msg)
+
+
+class ARMV7M7Runner(DeviceRunner):
+    """This class provides interface to run tests on targets with armv7m7 architecture"""
+
+    status_color = {Runner.BUSY: 'blue', Runner.SUCCESS: 'green', Runner.FAIL: 'red'}
+
+    # redefined by target runners
+    SDP = None
+    IMAGE = None
+
+    def __init__(self, serial, is_rpi_host=True):
+        # has to be defined before super, because Runner constructor calls set_status, where it's used
+        self.is_rpi_host = is_rpi_host
+        if self.is_rpi_host:
+            self.reset_gpio = GPIO(17)
+            self.reset_gpio.high()
+            self.power_gpio = GPIO(2)
+            self.power_gpio.high()
+            self.boot_gpio = GPIO(4)
+            self.leds = {'red': GPIO(13), 'green': GPIO(18), 'blue': GPIO(12)}
+
+        super().__init__(serial)
+        # default values, redefined by specified target runners
+        self.phoenixd_port = None
+        self.is_cut_power_used = False
+        self.flash_memory = 0
+
+    def _restart_by_poweroff(self):
+        pass
+
+    def _restart_by_jtag(self):
+        self.reset_gpio.low()
+        time.sleep(0.050)
+        self.reset_gpio.high()
+
+    def rpi_reboot(self, serial_downloader=False, cut_power=False):
+        """Reboots a tested device using Raspberry Pi as host-pc."""
+        if serial_downloader:
+            self.boot_gpio.low()
+        else:
+            self.boot_gpio.high()
+
+        if cut_power:
+            self._restart_by_poweroff()
+        else:
+            self._restart_by_jtag()
+
+    def reboot(self, serial_downloader=False, cut_power=False):
+        """Reboots a tested device."""
+        if self.is_rpi_host:
+            self.rpi_reboot(serial_downloader, cut_power)
+        else:
+            hostpc_reboot(serial_downloader)
+
+    def flash(self):
+        phd = None
+        try:
+            self.reboot(serial_downloader=True, cut_power=self.is_cut_power_used)
+            with PloTalker(self.serial_port) as plo:
+                Psu(script=self.SDP).run()
+                plo.wait_prompt()
+                with Phoenixd(self.phoenixd_port) as phd:
+                    plo.copy_file2mem(
+                        src='usb0',
+                        file=self.IMAGE,
+                        dst=f'flash{self.flash_memory}',
+                    )
+            self.reboot(cut_power=self.is_cut_power_used)
+        except (TIMEOUT, EOF, PloError, PhoenixdError, RebootError) as exc:
+            exception = f'{exc}\n'
+            if phd and not isinstance(exc, RebootError):
+                exception = phd_error_msg(exception, phd.output())
+
+            logging.info(exception)
+            sys.exit(1)
+
+    def load(self, test):
+        """Loads test ELF into syspage using plo"""
+        phd = None
+        load_dir = str(rootfs(test.target) / 'bin')
+        try:
+            self.reboot(cut_power=self.is_cut_power_used)
+            with PloTalker(self.serial_port) as plo:
+                plo.wait_prompt()
+
+                if not test.exec_cmd:
+                    # We got plo prompt, we are ready for sending the "go!" command.
+                    return True
+
+                with Phoenixd(self.phoenixd_port, dir=load_dir) as phd:
+                    plo.app('usb0', test.exec_cmd[0], 'ocram2', 'ocram2')
+        except (TIMEOUT, EOF, PloError, PhoenixdError, RebootError) as exc:
+            if isinstance(exc, TIMEOUT) or isinstance(exc, EOF):
+                test.exception = Color.colorify('EXCEPTION PLO\n', Color.BOLD)
+                test.handle_pyexpect_error(plo.plo, exc)
+            else:  # RebootError, PloError or PhoenixdError
+                test.exception = str(exc)
+                test.fail()
+                if phd:
+                    test.exception = phd_error_msg(test.exception, phd.output())
+            return False
+
+        return True
+
+    def set_status(self, status):
+        super().set_status(status)
+        if self.is_rpi_host:
+            if status in self.status_color:
+                self.set_led(self.status_color[status])
+
+    def set_led(self, color):
+        """Turn on the specified RGB LED's color."""
+        if color in self.leds:
+            for led_gpio in self.leds.values():
+                led_gpio.low()
+            self.leds[color].high()
+        else:
+            print(f'There is no specified led color: {color}')
+
+    def run(self, test):
+        if test.skipped():
+            return
+
+        if not self.load(test):
+            return
+        super().run(test)
--- a/trunner/runners/IMXRT106xRunner.py
+++ b/trunner/runners/IMXRT106xRunner.py
@@ -7,60 +7,36 @@
 # Authors: Jakub Sarzyński, Mateusz Niewiadomski, Damian Loewnau
 #
 
+import logging
 import time
 import sys
-from pexpect.exceptions import TIMEOUT, EOF
 
-from .common import DeviceRunner, PloTalker, PloError, Psu, Phoenixd, PhoenixdError, GPIO, Color, logging
-from .common import unbind_rpi_usb, power_usb_ports, wait_for_dev, phd_error_msg, rootfs
+from .ARMV7M7Runner import ARMV7M7Runner
+from .common import unbind_rpi_usb, power_usb_ports, wait_for_dev
 
 
-class IMXRT106xRunner(DeviceRunner):
+class IMXRT106xRunner(ARMV7M7Runner):
     """This class provides interface to run test case on IMXRT106x using RaspberryPi.
-       GPIO 17 must be connected to the JTAG_nSRST (j21-15) (using an additional resistor 1,5k).
-       GPIO 4 must be connected to the SW7-3 (using a resistor 4,3k).
-       GPIO 2 must be connected to an appropriate IN pin in relay module"""
+
+    Default configuration for running on Rpi:
+
+    GPIO 17 must be connected to the JTAG_nSRST (j21-15) (using an additional resistor 1,5k).
+    GPIO 4 must be connected to the SW7-3 (using a resistor 4,3k).
+    GPIO 2 must be connected to an appropriate IN pin in relay module
+    GPIO 12 must be connected to a blue pin of an RGB LED
+    GPIO 13 must be connected to a red pin of an RGB LED
+    GPIO 18 must be connected to a green pin of an RGB LED"""
 
     SDP = 'plo-ram-armv7m7-imxrt106x.sdp'
     IMAGE = 'phoenix-armv7m7-imxrt106x.disk'
 
-    def __init__(
-        self,
-        port,
-        phoenixd_port='/dev/serial/by-id/usb-Phoenix_Systems_plo_CDC_ACM-if00',
-        is_cut_power_used=False
-    ):
-        super().__init__(port[0])
+    def __init__(self, port, phoenixd_port, is_rpi_host=True):
+        super().__init__(port[0], is_rpi_host)
         self.port_usb = port[1]
         self.phoenixd_port = phoenixd_port
-        self.is_cut_power_used = is_cut_power_used
-        self.reset_gpio = GPIO(17)
-        self.reset_gpio.high()
-        self.power_gpio = GPIO(2)
-        self.power_gpio.high()
-        self.boot_gpio = GPIO(4)
-        self.ledr_gpio = GPIO(13)
-        self.ledg_gpio = GPIO(18)
-        self.ledb_gpio = GPIO(12)
-        self.ledb_gpio.high()
-
-    def led(self, color, state="on"):
-        if state == "on" or state == "off":
-            self.ledr_gpio.low()
-            self.ledg_gpio.low()
-            self.ledb_gpio.low()
-        if state == "on":
-            if color == "red":
-                self.ledr_gpio.high()
-            if color == "green":
-                self.ledg_gpio.high()
-            if color == "blue":
-                self.ledb_gpio.high()
-
-    def _restart_by_jtag(self):
-        self.reset_gpio.low()
-        time.sleep(0.050)
-        self.reset_gpio.high()
+        # restart by power off is temporarily disabled for this target
+        self.is_cut_power_used = False
+        self.flash_memory = 1
 
     def _restart_by_poweroff(self):
         unbind_rpi_usb(self.port_usb)
@@ -77,84 +53,3 @@ class IMXRT106xRunner(DeviceRunner):
         except TimeoutError:
             logging.error('Serial port not found!\n')
             sys.exit(1)
-
-    def reboot(self, serial_downloader=False, cut_power=False):
-        if serial_downloader:
-            self.boot_gpio.low()
-        else:
-            self.boot_gpio.high()
-
-        if cut_power:
-            self._restart_by_poweroff()
-        else:
-            self._restart_by_jtag()
-
-    def flash(self):
-        self.reboot(serial_downloader=True, cut_power=self.is_cut_power_used)
-
-        phd = None
-        try:
-            with PloTalker(self.serial_port) as plo:
-                Psu(script=self.SDP).run()
-                plo.wait_prompt()
-                with Phoenixd(self.phoenixd_port) as phd:
-                    plo.copy_file2mem(
-                        src='usb0',
-                        file=self.IMAGE,
-                        dst='flash1',
-                        off=0
-                    )
-        except (TIMEOUT, EOF, PloError, PhoenixdError) as exc:
-            exception = f'{exc}\n'
-            if phd:
-                exception = phd_error_msg(exception, phd.output())
-
-            logging.info(exception)
-            sys.exit(1)
-
-        self.reboot(cut_power=self.is_cut_power_used)
-
-    def load(self, test):
-        """Loads test ELF into syspage using plo"""
-
-        phd = None
-        load_dir = str(rootfs(test.target) / 'bin')
-        self.reboot(cut_power=self.is_cut_power_used)
-        try:
-            with PloTalker(self.serial_port) as plo:
-                # Because of powering all Rpi ports after powering the board,
-                # there is need to second reboot (without cut power) in order to capture all data
-                # (when using _restart_by_poweroff)
-                if self.is_cut_power_used:
-                    self.reboot(cut_power=False)
-                plo.wait_prompt()
-
-                if not test.exec_cmd:
-                    # We got plo prompt, we are ready for sending the "go!" command.
-                    return True
-
-                with Phoenixd(self.phoenixd_port, dir=load_dir) as phd:
-                    plo.app('usb0', test.exec_cmd[0], 'ocram2', 'ocram2')
-        except (TIMEOUT, EOF, PloError, PhoenixdError) as exc:
-            if isinstance(exc, PloError) or isinstance(exc, PhoenixdError):
-                test.exception = str(exc)
-                test.fail()
-            else:  # TIMEOUT or EOF
-                test.exception = Color.colorify('EXCEPTION PLO\n', Color.BOLD)
-                test.handle_pyexpect_error(plo.plo, exc)
-
-            if phd:
-                test.exception = phd_error_msg(test.exception, phd.output())
-
-            return False
-
-        return True
-
-    def run(self, test):
-        if test.skipped():
-            return
-
-        if not self.load(test):
-            return
-
-        super().run(test)
--- a/trunner/runners/IMXRT117xRunner.py
+++ b/trunner/runners/IMXRT117xRunner.py
@@ -1,0 +1,40 @@
+#
+# Phoenix-RTOS test runner
+#
+# IMXRT117x runner
+#
+# Copyright 2021 Phoenix Systems
+# Authors: Damian Loewnau, Jakub Sarzyński
+#
+
+import time
+
+from .ARMV7M7Runner import ARMV7M7Runner
+
+
+class IMXRT117xRunner(ARMV7M7Runner):
+    """This class provides interface to run test case on IMXRT117x using RaspberryPi.
+
+    Default configuration for running on Rpi:
+
+    GPIO 17 must be connected to the JTAG_nSRST (j21-15) (using an additional resistor 1,5k).
+    GPIO 4 must be connected to the pin 4 in P1 Header (using a resistor 3,3k).
+    GPIO 2 must be connected to an appropriate IN pin in relay module
+    GPIO 12 must be connected to a blue pin of an RGB LED
+    GPIO 13 must be connected to a red pin of an RGB LED
+    GPIO 18 must be connected to a green pin of an RGB LED"""
+
+    SDP = 'plo-ram-armv7m7-imxrt117x.sdp'
+    IMAGE = 'phoenix-armv7m7-imxrt117x.disk'
+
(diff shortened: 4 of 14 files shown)


trunner/runners/IMXRT106xRunner.py
```diff
@@ -89,72 +72,5 @@ def reboot(self, serial_downloader=False, cut_power=False):
...
-        return True
-
     def run(self, test):
-        if test.skipped():
-            return
-
-        if not self.load(test):
-            return
-
         super().run(test)
```
niewim19, line 76: If all that `run()` does is running `super().run()` then it is not unnecessary redefinition. You can omit it

trunner/runners/common.py
```diff
@@ -343,6 +378,84 @@ def run(self, test):
             self.serial.close()
 
 
+class ARMV7M7Runner(DeviceRunner):
+    """This class provides interface to run tests on targets with armv7m7 processor"""
```
niewim19, line 382:
```suggestion
    """This class provides interface to run tests on targets with armv7m7 architecture"""
```

trunner/test_runner.py
```diff
@@ -56,6 +56,7 @@ def run(self):
         for target, tests in self.tests_per_target.items():
             for test_case in tests:
                 test_case.log_test_started()
+                self.runners[target].status('busy')
```
niewim19, line 59: Maybe a dictionary of states somewhere in `config.py` that can. In future we want to add some state there may by less work

Or even better - an Enum class with statuses. Then we can simply use status.BUSY, status.FAILED etc

trunner/runners/IMXRT117xRunner.py
```diff
@@ -0,0 +1,52 @@
...
+       GPIO 4 must be connected to the pin 4 in P1 Header (using a resistor 3,3k).
+       GPIO 2 must be connected to an appropriate IN pin in relay module"""
+
+    SDP = 'plo-ram-armv7m7-imxrt117x.sdp'
+    IMAGE = 'phoenix-armv7m7-imxrt117x.disk'
+
+    def __init__(
+        self,
+        port,
+        phoenixd_port='/dev/serial/by-id/usb-Phoenix_Systems_plo_CDC_ACM-if00',
```
niewim19, line 26: Maybe this should be passed from `RunnerFactory` not hardcoded here?

niewim19 (commented):
I hope that in future we can exchange `ARMV7M7Runner` abstraction for something more generic like `NXPRunner` or something.

trunner/config.py
```diff
@@ -30,19 +30,25 @@ def resolve_phrtos_dir() -> Path:
...
-                   if target not in ('host-pc')]
+                   if target not in ('host-pc', 'armv7m7-imxrt117x')]
 
-DEVICE_TARGETS = ['armv7m7-imxrt106x']
+DEVICE_TARGETS = ['armv7m7-imxrt106x', 'armv7m7-imxrt117x']
 
-# Port to communicate with hardware board
-DEVICE_SERIAL = "/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.1"
+# Ports to communicate with hardware boards
+SERIAL_DEVICES = {
```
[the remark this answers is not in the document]
damianloew: I would prefer as it is, but if there are more similar suggestions, I will change it.

trunner/runners/IMXRT106xRunner.py
damianloew: That's right, thanks

trunner/runners/IMXRT117xRunner.py
```diff
@@ -0,0 +1,52 @@
...
+        super().__init__(port)
+        self.phoenixd_port = phoenixd_port
+        self.is_cut_power_used = True
+        self.flash_memory = 0
+        self.power_gpio = GPIO(2)
+        self.power_gpio.high()
+        self.boot_gpio = GPIO(4)
+
+    def _restart_by_poweroff(self):
+
```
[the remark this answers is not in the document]
damianloew: I liked it, but maybe that's right

trunner/test_runner.py
damianloew: I would prefer simpler solution for now. If we add other states, we need to change it here anyway

trunner/test_runner.py
```diff
@@ -67,4 +68,9 @@ def run(self):
                 failed += int(test.failed())
                 skipped += int(test.skipped())
 
+            if failed > 0:
+                self.runners[target].status('failed')
+            else:
+                self.runners[target].status('succeeded')
```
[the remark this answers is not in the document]
damianloew: yeah, maybe that's better

trunner/runners/IMXRT117xRunner.py
damianloew: That's right, I created `PHOENIXD_PORT` in config and use it.

trunner/runners/IMXRT106xRunner.py
```diff
@@ -27,35 +26,19 @@ class IMXRT106xRunner(DeviceRunner):
     def __init__(
         self,
         port,
-        phoenixd_port='/dev/serial/by-id/usb-Phoenix_Systems_plo_CDC_ACM-if00',
-        is_cut_power_used=False
+        phoenixd_port,
+        is_cut_power_used=False,
     ):
```
kemonats, line 31: Why not in one line?
```
    def __init__(self, port, phoenixd_port, is_cut_power_used=False):
```

trunner/runners/IMXRT117xRunner.py
```diff
@@ -0,0 +1,47 @@
...
+       GPIO 2 must be connected to an appropriate IN pin in relay module"""
+
+    SDP = 'plo-ram-armv7m7-imxrt117x.sdp'
+    IMAGE = 'phoenix-armv7m7-imxrt117x.disk'
+
+    def __init__(
+        self,
+        port,
+        phoenixd_port
+    ):
```
kemonats, line 27: Why not in one line.
```suggestion
    def __init__(self, port, phoenixd_port):
```

trunner/runners/common.py
```diff
@@ -308,21 +310,58 @@ def go(self):
...
+        if color in colors:
+            self.ledr_gpio.low()
+            self.ledg_gpio.low()
+            self.ledb_gpio.low()
+            if color == 'red':
+                self.ledr_gpio.high()
+            if color == 'green':
+                self.ledg_gpio.high()
+            if color == 'blue':
+                self.ledb_gpio.high()
```
kemonats, line 364: More `python` style:
```suggestion
        self.leds = {'red': GPIO(13), 'green': GPIO(18), 'blue': GPIO(12)}
        self.status_color = {'BUSY': 'blue', 'SUCCESS': 'green', 'FAIL': 'red'}

        self.leds['blue'].high()

    def status(self, status):
        if status in self.status_color:
            self.set_led(self.status_color[status])

    def set_led(self, color):
        if color in self.leds:
            for v in self.leds.values():
                v.low()
            self.leds[color].high()
```

damianloew: Thanks, changes applied

nalajcie:
nah, don't add merge commits to PR.

Please remove them and rebase the PR source branch (`damianloew/add_imxrt117x_runner`) on top of current master.

niewim19:
Yes, I have made a mistake, and I am correcting it right now

trunner/runners/common.py
```diff
@@ -308,21 +310,46 @@ def go(self):
 class Runner:
     """Common interface for test runners"""
 
+    @abstractmethod
```
nalajcie, line 313: You may use `@abstractmethod` only when the class is derived from `ABC`, see: https://docs.python.org/3/library/abc.html#abc.abstractmethod

trunner/runners/common.py
```diff
@@ -308,21 +310,46 @@ def go(self):
...
+    def status(self, status):
```
nalajcie, line 314: For `status` please use class consts instead of strings You can easily mistype (check how it's done in `TestCase` class).

I'm missing "idle" status + initialization to IDLE in the constructor here.

If we use `@abstractmethod` properly  - qemu runners will fail as this method is not overriden there (so think carefully if it really should be abstract here)

trunner/runners/common.py
```diff
@@ -308,21 +310,46 @@ def go(self):
...
+    """This class provides interface to run tests on hardware targets using serial port
+       Hardware targets are run on Rpi
+       GPIO 13 must be connected to a red LED
+       GPIO 18 must be connected to a green LED
+       GPIO 12 must be connected to a blue LED"""
 
     def __init__(self, port):
         self.port = port
         self.serial = None
+        self.leds = {'red': GPIO(13), 'green': GPIO(18), 'blue': GPIO(12)}
```
nalajcie, line 339: seems like class attribute, not instance one - but we might keep it as is if You think we will soon read leds configuration from external file dynamically when instantiating the class)

trunner/runners/common.py
```diff
@@ -308,21 +310,46 @@ def go(self):
...
+        self.status_color = {'BUSY': 'blue', 'SUCCESS': 'green', 'FAIL': 'red'}
```
nalajcie, line 340: definitely should be a class attribute, not instance one

trunner/runners/common.py
```diff
@@ -308,21 +310,46 @@ def go(self):
...
+
+        self.leds['blue'].high()
```
nalajcie, line 342: no, use self.status(SOMETHING)

trunner/test_runner.py
```diff
@@ -57,6 +57,7 @@ def run(self):
         for target, tests in self.tests_per_target.items():
             for test_case in tests:
                 test_case.log_test_started()
+                self.runners[target].status('BUSY')
```
nalajcie, line 60: probably should be in the outer `for`, no need to change state `BUSY -> BUSY` with each new test

trunner/runners/common.py
```diff
@@ -343,6 +370,97 @@ def run(self, test):
...
+        self,
+        port,
+    ):
+        # default values, redefined by specified target runners
+        self.is_cut_power_used = False
+        self.flash_memory = 0
+        super().__init__(port)
+
+    def _restart_by_jtag(self):
+        self.reset_gpio.low()
```
nalajcie, line 386: we're using instance attribute not defined in the same class, only in the childs.

Please define it in this class (to be overriden in subclasses if needed).

niewim19:
Sorry for temporary mess in tree. 

This branch is now correctly squashed and rebased after `STM32L4Runner` implementation with all conflicts resolved.

trunner/runners/common.py
```diff
@@ -343,6 +370,97 @@ def run(self, test):
...
+        time.sleep(0.050)
+        self.reset_gpio.high()
+
+    def reboot(self, serial_downloader=False, cut_power=False):
+        if serial_downloader:
+            self.boot_gpio.low()
```
nalajcie, line 392: same as above

trunner/runners/common.py
```diff
@@ -343,6 +370,97 @@ def run(self, test):
             self.serial.close()
 
 
+class ARMV7M7Runner(DeviceRunner, ABC):
+    """This class provides interface to run tests on targets with armv7m7 architecture"""
```
nalajcie, line 374: I think we should put this class in separate file, hopefully it will replace custom runners for RT106x and RT117x in the futre.

nalajcie (changes requested):
pointed out some remarks ;)

nalajcie:
> This branch is now correctly squashed and rebased after `STM32L4Runner` implementation with all conflicts resolved.

Please note that squash was not needed (but it might have been desirable if there were too many commits).

damianloew:
> Sorry for temporary mess in tree.
> 
> This branch is now correctly squashed and rebased after `STM32L4Runner` implementation with all conflicts resolved.
> 
> Awaiting @damianloew confirmation about whether it works correctly on hardware

Yes, it's working, thanks

trunner/runners/common.py
```diff
@@ -308,21 +310,46 @@ def go(self):
...
+    @abstractmethod
     def run(self, test):
         """Method used for running a single test case which is represented by TestCase class."""
         pass
 
 
 class DeviceRunner(Runner):
-    """This class provides interface to run test case using serial port"""
+    """This class provides interface to run tests on hardware targets using serial port
+       Hardware targets are run on Rpi
```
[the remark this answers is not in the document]
damianloew: I planned to make running on host-pc possible in next pr's, but that's right it's not the point to make it harder. So for now, I've added the `is_rpi_host` argument to make it easier. I think the next steps will be only adding argument to runner about that (I thought about separate pr)

trunner/runners/common.py
damianloew: Thanks, changed to consts

Because of using only 3 led collors, there are 3 states: busy, pass, fail. The idle state is pass or fail (depend on last runner run). In the future there can be more statuses

I've put it here, because I was afraid that without this declaration runners without status method can't be called from self.runners[target].status(). But that's right, if it's abstract, they should have it defined. Can I leave it here without @abstractmethod or add empty definitions for Qemu, Host and STM Runner?

trunner/runners/common.py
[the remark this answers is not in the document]
nalajcie: we would probably like to read information about test environment from YAML file in the future (with some sane defaults if no file is present - probably a subset working on laptop).

It should have all necessary information regarding connected devices - like serial console port paths, reboot strategy, existence of status leds, etc.

(the YAML idea is not yet precise enough, should we have 2 separate configs - one for TestEnv (which should abstract things like LEDs as it's a part of test environment - not the device itself) and second for DeviceRunner? Or just one? To be designed in the future.)

trunner/runners/common.py
damianloew: I hope it will be soon, so I will keep it

trunner/runners/common.py
nalajcie: - I proposed "idle" as it's a good value for initialization. In this particular case maybe "busy" is a good init value, but we should write in comment why is that so (as it seems counter-intuitive that the runner status is busy when it's not running any test, this is actually the "DUT status" (device under test)
- maybe the default implementation shouldn't be empty, but should at least set `self.status` to be available for getting in the future?
- the mapping 1-1 status to gpio is the most basic one (but well enough for now), I can think of having one RGB led where You can mix any color You want using 3 PWM channels in the future :)

trunner/runners/common.py
```diff
@@ -308,21 +310,46 @@ def go(self):
...
+        self.leds = {'red': GPIO(13), 'green': GPIO(18), 'blue': GPIO(12)}
+        self.status_color = {'BUSY': 'blue', 'SUCCESS': 'green', 'FAIL': 'red'}
+
+        self.leds['blue'].high()
+
+    def status(self, status):
+        if status in self.status_color:
+            self.set_led(self.status_color[status])
+
+    def set_led(self, color):
```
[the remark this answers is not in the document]
damianloew: I would keep it, because of `is_rpi_host`

trunner/test_runner.py
damianloew: That's right, thanks

trunner/runners/common.py
```diff
@@ -308,21 +310,46 @@ def go(self):
...
+        if color in self.leds:
```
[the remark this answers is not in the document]
damianloew: Nothing happens, I've added printing a statement about it

trunner/runners/common.py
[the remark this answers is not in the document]
damianloew: I think, that's good idea, so I tried to make it easier (running device runners on host pc)

damianloew:
> pointed out some remarks ;)

Thanks, I've applied that

trunner/runners/common.py
damianloew: Sorry, I missed that, so I needed another force push to apply it

- You are right, I've added additional comment

- I agree that's better, applied

- Yes, that is an RGB diode and mixing colors using PWM was the plan, but because it's not so important there is simpler solution now

trunner/runners/common.py
```diff
@@ -305,25 +306,43 @@ def go(self):
...
 
     def __init__(self, port):
         self.port = port
         self.serial = None
 
+    @abstractmethod
+    def set_status(self, status):
+        pass
+
+    @abstractmethod
```
nalajcie, line 342: was not abstract in base class (`Runner`). What is happening? :)

trunner/runners/IMXRT117xRunner.py
```diff
@@ -0,0 +1,45 @@
...
+from .ARMV7M7Runner import ARMV7M7Runner
+
+
+class IMXRT117xRunner(ARMV7M7Runner):
+    """This class provides interface to run test case on IMXRT117x using RaspberryPi.
+
+    Default configuration for running on Rpi:
+
+    GPIO 17 must be connected to the JTAG_nSRST (j21-15) (using an additional resistor 1,5k).
+
```
nalajcie, line 21: I think we don't need double-newline with each new sentence.

trunner/runners/__init__.py
```diff
@@ -9,9 +9,11 @@
...
 
+from .ARMV7M7Runner import ARMV7M7Runner
 from .HostRunner import HostRunner
 from .QemuRunner import QemuRunner
 from .IMXRT106xRunner import IMXRT106xRunner
+from .IMXRT117xRunner import IMXRT117xRunner
 from .STM32L4Runner import STM32L4Runner
 
-__all__ = ['HostRunner', 'QemuRunner', 'IMXRT106xRunner', 'STM32L4Runner']
+__all__ = ['ARMV7M7Runner', 'HostRunner', 'QemuRunner', 'IMXRT106xRunner', 'IMXRT117xRunner', 'STM32L4Runner']
```
nalajcie, line 19: I think `ARMV7M7Runner` is internal class and should not be exported by `runners` module by default

trunner/runners/common.py
```diff
@@ -305,25 +306,43 @@ def go(self):
...
-class DeviceRunner(Runner):
-    """This class provides interface to run test case using serial port"""
+class DeviceRunner(Runner, ABC):
+    """This class provides interface to run tests on hardware targets using serial port"""
 
     def __init__(self, port):
         self.port = port
         self.serial = None
 
+    @abstractmethod
```
nalajcie, line 338: no, You already have implementation in base class

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if self.is_rpi_host:
+            self.reset_gpio = GPIO(17)
+            self.reset_gpio.high()
+            self.power_gpio = GPIO(2)
+            self.power_gpio.high()
+            self.boot_gpio = GPIO(4)
+            self.leds = {'red': GPIO(13), 'green': GPIO(18), 'blue': GPIO(12)}
+
+            self.set_status(Runner.BUSY)
```
nalajcie, line 47: this should be probably done in constructor of `Runner` class regardless or `is_rpi_host` as otherwise status is undefined

trunner/test_runner.py
nalajcie: not needed at all if setting in constructor

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,161 @@
...
+    @abstractmethod
+    def _restart_by_poweroff(self):
+        pass
+
+    def _restart_by_jtag(self):
+        self.reset_gpio.low()
+        time.sleep(0.050)
+        self.reset_gpio.high()
+
+    def hostpc_reboot(self, serial_downloader=False):
```
nalajcie, line 58: docstring

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,161 @@
...
+        signal.alarm(30)
+        if serial_downloader:
+            print("\n\tPlease change the board's boot mode to serial download, reset the board and press enter")
+        else:
+            print("\n\tPlease change the board's boot mode to internal/external flash if the current mode is different.\n \
+                    \t Next, reset the board and press enter")
+        input("Note that, You should press the enter key just after reset button")
+        signal.alarm(0)
+
+    def rpi_reboot(self, serial_downloader=False, cut_power=False):
```
nalajcie, line 69: docstring

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,161 @@
...
+            self.boot_gpio.low()
+        else:
+            self.boot_gpio.high()
+
+        if cut_power:
+            self._restart_by_poweroff()
+        else:
+            self._restart_by_jtag()
+
+    def reboot(self, serial_downloader=False, cut_power=False):
```
nalajcie, line 80: docstring

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,161 @@
...
+                test.exception = phd_error_msg(test.exception, phd.output())
+            return False
+
+        return True
+
+    def set_status(self, status):
+        if self.is_rpi_host:
+            if status in self.status_color:
+                self.set_led(self.status_color[status])
+        else:
```
nalajcie, line 144: unnecessary else

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,161 @@
...
+from pexpect.exceptions import TIMEOUT, EOF
+from abc import ABC, abstractmethod
+
+
+def handler(signum, frame):
+    logging.error("\nThe boot mode's change wasn't validated by pressing enter!\n")
+    sys.exit(1)
+
+
+class ARMV7M7Runner(DeviceRunner, ABC):
```
nalajcie, line 27: pylint errors:
```
trunner/runners/ARMV7M7Runner.py:22:12: W0613: Unused argument 'signum' (unused-argument)
trunner/runners/ARMV7M7Runner.py:22:20: W0613: Unused argument 'frame' (unused-argument)
trunner/runners/ARMV7M7Runner.py:58:4: R0201: Method could be a function (no-self-use)
trunner/runners/ARMV7M7Runner.py:92:27: E1101: Instance of 'ARMV7M7Runner' has no 'SDP' member (no-member)
trunner/runners/ARMV7M7Runner.py:94:30: E1101: Instance of 'ARMV7M7Runner' has no 'phoenixd_port' member (no-member)
trunner/runners/ARMV7M7Runner.py:97:29: E1101: Instance of 'ARMV7M7Runner' has no 'IMAGE' member (no-member)
trunner/runners/ARMV7M7Runner.py:124:30: E1101: Instance of 'ARMV7M7Runner' has no 'phoenixd_port' member (no-member)
trunner/runners/ARMV7M7Runner.py:17:0: C0411: first party import "from trunner.tools.color import Color" should be placed before "from .common import Psu, Phoenixd, PhoenixdError, PloError, PloTalker, Runner, DeviceRunner" (wrong-import-order)
trunner/runners/ARMV7M7Runner.py:18:0: C0411: third party import "from pexpect.exceptions import TIMEOUT, EOF" should be placed before "from trunner.tools.color import Color" (wrong-import-order)
trunner/runners/ARMV7M7Runner.py:19:0: C0411: standard import "from abc import ABC, abstractmethod" should be placed before "from pexpect.exceptions import TIMEOUT, EOF" (wrong-import-order)
```

ABC probably not needed

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,161 @@
...
+from .common import Psu, Phoenixd, PhoenixdError, PloError, PloTalker, Runner, DeviceRunner
+from .common import GPIO, logging, phd_error_msg, rootfs
+from trunner.tools.color import Color
+from pexpect.exceptions import TIMEOUT, EOF
+from abc import ABC, abstractmethod
+
+
+def handler(signum, frame):
+    logging.error("\nThe boot mode's change wasn't validated by pressing enter!\n")
+    sys.exit(1)
```
nalajcie, line 24: I don't like exiting from trunner in some error and potentially leaving test setup in invalid state (but that is a generic problem I've pointed out in @niewim19 PR recently).

Timeout should result in exception, thus failed testcase (or whole testing depending on when it happened) and should be propagated to the upper layer.

Also: using signals might not be the best way to set timeout for user input - if we would ever think of making this toolkit multi-threaded. I'm not focusing on reset-by-human use case as it's not the main point of this PR

nalajcie (changes requested):
code looks better now, I've pointed some small issues which still need to be fixed

trunner/runners/common.py
```diff
@@ -305,25 +306,43 @@ def go(self):
...
+    BUSY = 'BUSY'
+    SUCCESS = 'SUCCESS'
+    FAIL = 'FAIL'
+
+    def set_status(self, status):
+        """Method for sygnalising a current runner status: busy/failed/succeeded"""
+        # for now, not used in all target runners
+        self.status = status
+
     def flash(self):
```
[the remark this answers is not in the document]
damianloew: I think it can't be abstract, because there is no flash() method in QemuRunner and HostRunner

trunner/runners/common.py
```diff
@@ -305,25 +306,43 @@ def go(self):
...
 
+    @abstractmethod
     def run(self, test):
         """Method used for running a single test case which is represented by TestCase class."""
         pass
 
 
-class DeviceRunner(Runner):
-    """This class provides interface to run test case using serial port"""
+class DeviceRunner(Runner, ABC):
```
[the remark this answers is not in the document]
damianloew: That's right, sorry

trunner/runners/common.py
damianloew: I thought like that, every DeviceRunner has to have flash() and set_status() methods defined (so they are abstract methods and there will be error, if child class doesn't implement them ), but not every Runner. I am not sure what is a correct approach here.

trunner/runners/IMXRT117xRunner.py
damianloew: I thought it looks better in a preview, like here. But I think without it it looks better in code, changed

trunner/runners/common.py
damianloew: That's right, it can't be abstract, because not all of the device runners implement it. (for example stm runner)

trunner/runners/ARMV7M7Runner.py
damianloew: Right, the set_status isn't needed here, when the status is set to `BUSY` in the Runner init

trunner/runners/ARMV7M7Runner.py
damianloew: I've tested a few solutions and the last one is my proposal.

If host_pc reboot fail is when flashing, whole testing is stopped, but when loading tests, the specific test case fails.
Screenshot from 2021-11-26 18-09-33
Screenshot from 2021-11-26 18-10-23

nalajcie (commented):
left some comments

trunner/runners/ARMV7M7Runner.py
nalajcie: still, You need to initialize status with a known value in constructor, as it should always have one of the defined values. 

In child class You can always call set_status second time after initializing GPIOS if needed (with some comment), but You may keep also changing state to BUSY in `test_runner.py`.

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,169 @@
...
+    def __init__(self, message):
+        msg = Color.colorify("REBOOT ERROR:\n", Color.BOLD)
+        msg += str(message) + '\n'
+        super().__init__(msg)
+
+
+class ARMV7M7Runner(DeviceRunner):
+    """This class provides interface to run tests on targets with armv7m7 architecture"""
+
+    status_color = {'BUSY': 'blue', 'SUCCESS': 'green', 'FAIL': 'red'}
```
nalajcie, line 45: use Runner.BUSY instead of direct strings

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,169 @@
...
+        self.is_cut_power_used = False
+        self.flash_memory = 0
+
+        if self.is_rpi_host:
+            self.reset_gpio = GPIO(17)
+            self.reset_gpio.high()
+            self.power_gpio = GPIO(2)
+            self.power_gpio.high()
+            self.boot_gpio = GPIO(4)
+            # without this init gpio values an rgb led is turned off for a while
```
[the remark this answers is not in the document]
damianloew: There was a problem, because, the led gpio init was before setting Runner state to `BUSY` (in runner.flash()) (the first screen). So I moved the setting state to `BUSY` to Runner class constructor. But I had to call super().__init__ after gpio init in ARMV7M7 class (the second screen).
Screenshot from 2021-12-01 12-11-00

Screenshot from 2021-12-01 12-16-56

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,169 @@
...
+            if isinstance(exc, PloError) or isinstance(exc, PhoenixdError):
+                test.exception = str(exc)
+                test.fail()
+            elif isinstance(exc, TIMEOUT) or isinstance(exc, EOF):
+                test.exception = Color.colorify('EXCEPTION PLO\n', Color.BOLD)
+                test.handle_pyexpect_error(plo.plo, exc)
+            else:  # RebootError
+                test.exception = str(exc)
+                test.fail()
+            if phd:
```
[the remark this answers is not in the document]
damianloew: > left some comments

Thanks, fixes applied

damianloew:
> 

Thanks, fixes applied

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,167 @@
...
+#
+
+import time
+import sys
+import select
+
+from pexpect.exceptions import TIMEOUT, EOF
+from trunner.tools.color import Color
+from .common import Psu, Phoenixd, PhoenixdError, PloError, PloTalker, DeviceRunner, Runner
+from .common import GPIO, logging, phd_error_msg, rootfs
```
jsarzy, line 17: logging is not the part of common, just `import logging`

trunner/runners/ARMV7M7Runner.py
```diff
@@ -0,0 +1,167 @@
...
+            self.reboot(serial_downloader=True, cut_power=self.is_cut_power_used)
+            with PloTalker(self.port) as plo:
+                Psu(script=self.SDP).run()
+                plo.wait_prompt()
+                with Phoenixd(self.phoenixd_port) as phd:
+                    plo.copy_file2mem(
+                        src='usb0',
+                        file=self.IMAGE,
+                        dst=f'flash{self.flash_memory}',
+                        off=0
```
jsarzy, line 107: nit: off is by default 0

trunner/runners/IMXRT106xRunner.py
```diff
@@ -9,58 +9,33 @@
 
 import time
 import sys
-from pexpect.exceptions import TIMEOUT, EOF
 
-from .common import DeviceRunner, PloTalker, PloError, Psu, Phoenixd, PhoenixdError, GPIO, Color, logging
-from .common import unbind_rpi_usb, power_usb_ports, wait_for_dev, phd_error_msg, rootfs
+from .ARMV7M7Runner import ARMV7M7Runner
+from .common import unbind_rpi_usb, power_usb_ports, wait_for_dev, logging
```
jsarzy, line 14: logging is not part of common

trunner/test_runner.py
```diff
@@ -55,6 +56,7 @@ def run(self):
                 runner.flash()
 
         for target, tests in self.tests_per_target.items():
+            # self.runners[target].set_status(Runner.BUSY)
```
jsarzy, line 59: commented code

jsarzy (commented):
few nits, it would be nice to keep imports in alphabetical order

trunner/runners/common.py
nalajcie: remove abstract implementation from here, it's already abstract in `Runner` class

trunner/runners/ARMV7M7Runner.py
damianloew: thanks, removed

damianloew:
> few nits, it would be nice to keep imports in alphabetical order

Earlier, @nalajcie wrote about pylint warnings, where the target order was like that, so I left it. Thanks for review

damianloew:
The newest commit rebased. @niewim19 please confirm that, the STMRunner works fine with this commit.

trunner/runners/common.py
```diff
@@ -305,26 +305,46 @@ def go(self):
...
-    """This class provides interface to run test case using serial port"""
+    """This class provides interface to run tests on hardware targets using serial port"""
 
     def __init__(self, serial):
+        super().__init__()
         self.serial_port = serial[0]
         self.serial_baudrate = serial[1]
         self.serial = None
 
+    def set_status(self, status):
```
nalajcie, line 345: not needed (function of the parent class will be called if it's not redefined in the child class)

nalajcie (commented):
small styling comments, otherwise lgtm.

nalajcie (approved):

PR #66, opened 2021-11-26
runner: `--baudrate` option and `psh` yaml property

niewim19:
JIRA: SKAL-193

## Description
Added two new features to runner:
 - `--baudrate` commandline option specifying the serial communication speed
 - `psh` yaml property which if runner should expect psh to appear on serial connection

## Motivation and Context

 - `--baudrate`:
   - Some future test targets can have multiple serial speed connections, so this may be handy for development purposes.
   - For full configuration there should be options for parity bits/stop bits etc. but this should be done when those config options will be moved to some _yaml_ config file
   - this change should not be necessary for full CI, but rather for development of new runners, together with project specific runners
 - `psh`:
   - some tests or test targets may not run through `psh` shell so it should be possible to specify it

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (stm32l4).

## Checklist:

## Special treatment

### DIFF
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -45,9 +45,6 @@ DEVICE_SERIAL_BAUDRATE = 115200
 # DEVICE_SERIAL USB port address
 DEVICE_SERIAL_USB = "1-1.4"
 
-# Default psh prompt
-DEFAULT_PROMPT = '(psh)% '
-
 
 class ParserError(Exception):
     pass
@@ -112,7 +109,6 @@ class Config(dict):
     def setdefaults(self) -> None:
         self.setdefault('ignore', False)
         self.setdefault('type', 'unit')
-        self.setdefault('prompt', '(psh)% ')
         self.setdefault('psh', True)
         self.setdefault('timeout', PYEXPECT_TIMEOUT)
         self.setdefault_targets()
@@ -144,7 +140,7 @@ class TestConfig(Config):
 
 
 class ConfigParser:
-    KEYWORDS: Tuple[str, ...] = ('exec', 'harness', 'ignore', 'name', 'targets', 'psh', 'prompt', 'timeout', 'type')
+    KEYWORDS: Tuple[str, ...] = ('exec', 'harness', 'ignore', 'name', 'targets', 'psh', 'timeout', 'type')
     TEST_TYPES: Tuple[str, ...] = ('unit', 'harness')
 
     def parse_keywords(self, config: Config) -> None:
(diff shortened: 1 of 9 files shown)


damianloew (commented):
The additional `baudrate` argument may be helpful for now, but when there will be a config yaml file I would like to set it only in yaml with other additional parameters, as You said. For me looks good, but I don't think I can approve it by myself

trunner/config.py
```diff
@@ -108,6 +112,8 @@ def setdefault_targets(self) -> None:
     def setdefaults(self) -> None:
         self.setdefault('ignore', False)
         self.setdefault('type', 'unit')
+        self.setdefault('prompt', '(psh)% ')
```
jsarzy, line 115:
```suggestion
        self.setdefault('prompt', 'DEFAULT_PROMPT ')
```

trunner/config.py
```diff
@@ -214,6 +220,23 @@ def parse_exec(self, config: Config) -> None:
...
+        psh = config.get('psh')
+        if not psh:
+            return
+        psh = str(psh).lower()
+        if psh == 'true':
+            config['psh'] = True
+        elif psh == 'false':
+            config['psh'] = False
+        else:
+            raise ParserError(f'psh value of {psh} is not "True" or "False"')
```
jsarzy, line 238: There is no need for pasing true/false, yaml package do it for us. Look how `ignore` keyword is parsed.

trunner/config.py
```diff
@@ -214,6 +220,23 @@ def parse_exec(self, config: Config) -> None:
 
         config['exec'] = shlex.split(exec_cmd)
 
+    def parse_prompt(self, config: Config) -> None:
+        prompt = config.get('prompt')
+        if not prompt:
+            return
```
jsarzy, line 226: No need for this if, it do nothing.

trunner/testcase.py
```diff
@@ -157,6 +161,10 @@ def handle(self, proc, psh=True):
 
         self.status = TestCase.PASSED
 
+        # override psh parameter if set to False
+        if self.psh is False:
+            psh = False
```
jsarzy, line 166: Remove psh argument and use self.psh instead.

jsarzy (changes requested):
As we talk earlier, there is no need for `prompt` keyword for now.
Things to do:
- Remove `psh` argument in `handle` function, instead use `self.psh`. It should require few changes (`test/test_exceptions.py`, runners)
- Host pc tests use `psh` argument. We would like to use `self.psh` here and set it always to false. 
- Config keywords have its own tests in the `test` directory. I would like to see test for `psh` keyword, it should be similar to `ignore` keyword :). You can look into ci-runner.yaml to see how test are run for trunner.

niewim19:
### Cumulative answer after @jsarzy review:
 - `psh` argument usage in `handle()` function changed
 - tests of `trunner` were updated for:
   - updated for changes in its api (`psh` argument usage)
   - updaed with tests for `psh` keyword
 - care taken of `psh` usage in `HostRunner`

trunner/config.py
```diff
@@ -39,11 +39,15 @@ def resolve_phrtos_dir() -> Path:
...
-DEVICE_SERIAL = "/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.1"
+DEVICE_SERIAL_PORT = "/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.1"
+DEVICE_SERIAL_BAUDRATE = 115200
 
 # DEVICE_SERIAL USB port address
 DEVICE_SERIAL_USB = "1-1.4"
 
+# Default psh prompt
+DEFAULT_PROMPT = '(psh)% '
+
```
jsarzy, line 50: Do we need this var? When `prompt` keyword was implemented it was necessary but now is only used in the `handle` function.

jsarzy (dismissed):
lgtm, one small remark

trunner/config.py
niewim19: @jsarzy your remark was implemented

jsarzy (approved):

PR #69, opened 2021-12-06
workflows: add getting hardware test targets from one file

damianloew:
JIRA: CI-97

## Description

workflows: add getting hardware test targets from one file

- add getting targets for test-hw job from file in phoenix-rtos-tests (temporarily from this branch, after merge the file will be in master branch and I will update it)
- add file hw_targets.json file in .github directory

## Motivation and Context

This change will allow adding and removing hardware targets for testing simply by editing one file (instead of all ci.yaml files in various repos).
 
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/.github/hw_targets.json
+++ b/.github/hw_targets.json
@@ -1,0 +1,1 @@
+["armv7m7-imxrt106x"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,18 @@ on:
 
 # jobs
 jobs:
+  # getting hardware test targets from the file placed in phoenix-rtos-tests repository
+  get-hw-targets:
+    name: get hardware targets for testing
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get hardware test targets
+        id: get_file
+        run: |
+          echo "::set-output name=matrix::$(curl https://raw.githubusercontent.com/phoenix-rtos/phoenix-rtos-tests/damianloew/get_hw_targets_from_file/.github/hw_targets.json)"
+    outputs:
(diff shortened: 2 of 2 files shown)


nalajcie:
I would prefer the reusable workflows approach (keeping the actions specifications in one single location, not spread over many files/repositories). Maybe in the future we might want to dynamically gather supported platforms from test runner, but IMHO it would never be needed (static targets declaration approach is less prone to errors and strange cases).

damianloew:
> I would prefer the reusable workflows approach (keeping the actions specifications in one single location, not spread over many files/repositories). Maybe in the future we might want to dynamically gather supported platforms from test runner, but IMHO it would never be needed (static targets declaration approach is less prone to errors and strange cases).

Yeah, reusable workflows seems to be a better solution. I've tested it and it's working, but there is one problem I can't solve. PR is waiting for build checks, which already have been done. In the result checks are still in progress. I couldn't find where this window is implemented. 
Screenshot from 2022-01-14 14-00-20

I attached link to the pr, where I've tested it (reusable workflow is on my branch in phoenix-rtos-tests):
https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/107

Moreover, workflows in phoenix-rtos-project are little different than in other repositories. So with reusable workflows there will be two places, where for example hardware targets should be changed.

Finally, I propose the solution with a file, and eventually remove it when reusable workflows will work properly (then it will only be 2 ci.yml files to edit - one for phoenix-rtos-project and second for other repositories).

damianloew:
Probably won't be introduced

PR #85, opened 2022-01-24
posix: unix socket tests (peer socket close, poll)

ziemleszcz:
JIRA: DTR-240, DTR-241

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imx6ull).

## Checklist:

## Special treatment


socket/test_socket.c
```diff
@@ -779,6 +793,236 @@ TEST(test_unix_socket, unix_transfer)
...
+	TEST_ASSERT(n == -1);
+	TEST_ASSERT(errno == err);
+
+	got_epipe = 0;
+	n = send(fd[0], data, sizeof(data), 0);
+	TEST_ASSERT(got_epipe == epipe);
+	TEST_ASSERT(n == -1);
+	TEST_ASSERT(errno == err);
+
+	close(fd[0]);
```
agkaminski, line 852: I don't know if we care or how to solve this, but won't this fd leak if some `TEST_ASSERT` fail?

ziemleszcz: I assumed that leaking few descriptors is not a big deal - automatic tests run on a fresh system. I you really don't like it please add a new issue and I will fix later.

agkaminski (dismissed):

damianloew (commented):
Just rebased. I see that these changes were approved, @agkaminski @ziemleszcz can we merge this?

libc/Makefile
```diff
@@ -17,6 +17,19 @@ LOCAL_LDFLAGS := -z stack-size=12288 $(2)
...
+NAME := test-libc-$(2)
+LIBC_UNIT_TESTS += test-libc-$(2)
+SRCS := $(addprefix $(MY_LOCAL_DIR)$(1)/, $(CUSTOM_SRCS)) $(COMMON_SRCS)
+DEP_LIBS := unity
+LOCAL_CFLAGS := -I$(MY_LOCAL_DIR) $(4)
+LOCAL_LDFLAGS := -z stack-size=12288 $(3)
+
+include $(binary.mk)
+endef
+
```
mateusz-bloch, line 32: I don't like using almost the same macro again. I get that you want to make two different tests from one folder because not all tests will work on every target. But, can't we just use an if statement in the macro for setting SRCS, or split the tests into two folders instead?

libc/socket/inet-socket.c
```diff
@@ -0,0 +1,172 @@
...
+#include <string.h>
+#include <errno.h>
+#include <sys/uio.h>
+#include <sys/wait.h>
+
+#include "common.h"
+#include "unity_fixture.h"
+
+char data[DATA_SIZE];
+char buf[DATA_SIZE];
```
mateusz-bloch, line 32: Is this used anywhere?
```suggestion
char buf[DATA_SIZE];
```

libc/socket/inet-socket.c
```diff
@@ -0,0 +1,172 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include <sys/uio.h>
+#include <sys/wait.h>
+
+#include "common.h"
+#include "unity_fixture.h"
+
+char data[DATA_SIZE];
```
mateusz-bloch, line 31: Is it necessary for this to be declared globally?

adamdebek: It will be used much more in all future test cases and it is initialized by test setup.

libc/Makefile
```diff
@@ -17,6 +17,19 @@ LOCAL_LDFLAGS := -z stack-size=12288 $(2)
 include $(binary.mk)
 endef
 
+# Add libc test using custom sources from subdir
+# $(eval $(call add_libc_test,SUBDIR_NAME,NAME[,LOCAL_LDFLAGS][,LOCAL_CFLAGS]))
+define add_test_libc_custom
+NAME := test-libc-$(2)
+LIBC_UNIT_TESTS += test-libc-$(2)
+SRCS := $(addprefix $(MY_LOCAL_DIR)$(1)/, $(CUSTOM_SRCS)) $(COMMON_SRCS)
```
nalajcie, line 25: - using external variable in macro instead of param is a way to introduce obscure errors
- IMHO you should either write it without a macro or specify a macro that just takes the list of c files to compile
- why not a single binary for socket tests?

libc/socket/inet-socket.c
nalajcie: this is defined as extern in common.h so should probably be placed in common.c, please remove it from here.

Still, the `data` is being used only once in `common.c` so it probably doensn't need to be global (extern), just local (static) for the test.

libc/Makefile
adamdebek: I create 2 binaries because `unix-sockets` test is run on all targets and `inet-sockets` test is run on targets with lwip
image

nalajcie: Still could be a single binary - with different test groups and using -g param in runtime to split the inet/unix tests - if it's too cumbersome to split them into different binaries (I assume all tests should compile regardless of the target architecture).

Side note: please don't add extra tests in this PR, we should aim for minimal changes that make them work and open new PR to bring new tests/functionality.

damianloew (changes requested):
CI fails - you probably want to run this test only for nightly campaigns

damianloew (commented):
The `TEST(test_unix_socket, unix_msg_data_only)` test case takes too much time IMHO - ~30s, when on imxrt106x it takes ~1s, it has to be investigated. For now, let's disable unix-socket this target.

libc/test.yaml
```diff
@@ -62,3 +62,14 @@ test:
       execute: test-libc-time
       targets:
         include: [host-generic-pc]
+
+    - name: unix-socket
+      execute: test-libc-unix-socket
+      targets:
+        exclude: [armv7m4-stm32l4x6-nucleo]
```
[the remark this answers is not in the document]
damianloew: Good point, but I think it will be better to fix this issue (if there is some) and then enable back the whole test or if the stm is just slower enable this test as nightly one.

damianloew (approved):

PR #87, opened 2022-02-01
runner: catch unicode error

damianloew:
## Description

- add handling UnicodeDecodeError exception
- trunner: change codec from utf-8 to ascii
- trunner: combine two try blocks into one in test execution

## Motivation and Context

For the problem known from the following issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/295 instead of Unicode Decode Error and aborting test runner, the output is printed.

Before change (test-helloworld prints here the following file content: ab�c):
Screenshot from 2022-02-14 10-47-06

After change:
Screenshot from 2022-03-01 14-55-35

https://github.com/phoenix-rtos/phoenix-rtos-project/issues/175

JIRA: PD-238

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic).

## Checklist:

## Special treatment

### DIFF
--- a/trunner/testcase.py
+++ b/trunner/testcase.py
@@ -95,6 +95,9 @@ class TestCase:
             self.exception = Color.colorify(msg, Color.BOLD)
             self.handle_pyexpect_error(proc, exc)
             return
+        except UnicodeDecodeError as exc:
+            self.handle_unicode_error(proc, exc)
+            return
 
         if self.exec_cmd:
             try:
@@ -106,6 +109,9 @@ class TestCase:
                 msg = f'Executing {self.exec_cmd} failed!\n'
                 self.exception = Color.colorify(msg, Color.BOLD)
                 self.handle_pyexpect_error(proc, exc)
+                return
+            except UnicodeDecodeError:
+                self.handle_unicode_error(proc)
                 return
 
     def handle_pyexpect_error(self, proc, exc):
@@ -124,6 +130,17 @@ class TestCase:
         got = str(proc.before)
         got = got.replace("\r", "")
         self.exception += Color.colorify('GOT:\n', Color.BOLD)
+        self.exception += f'{got}\n'
+
+    def handle_unicode_error(self, proc, exc):
+        self.status = TestCase.FAILED
+        msg = 'Unicode Decode Error detected!\n'
+        self.exception = Color.colorify(msg, Color.BOLD)
+        self.exception += Color.colorify('STRING WITH NON-ASCII CHARACTER:\n', Color.BOLD)
+        self.exception += f'{exc.object}\n'
+        self.exception += Color.colorify('OUTPUT CAUGHT BEFORE EXCEPTION:\n', Color.BOLD)
+        got = str(proc.before)
+        got = got.replace("\r", "")
         self.exception += f'{got}\n'
 
     def handle_assertion(self, proc, exc):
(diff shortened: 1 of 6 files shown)


damianloew:
I've also tested a solution with `proc.read()`, which could get a whole output at once, but it doesn't work (encounters timeout error, because it can't find eof). That's why there is reading 1 byte each.

xvuko:
Should we use UTF-8 to decode process output? Why ASCII codec is not used?

damianloew:
> Should we use UTF-8 to decode process output? Why ASCII codec is not used?

That's right, ASCII codec could be better. I will change it in next commit here.

trunner/testcase.py
```diff
@@ -171,6 +188,11 @@ def handle(self, proc):
             msg = 'EXCEPTION ' + ('EOF' if isinstance(exc, EOF) else 'TIMEOUT') + '\n'
             self.exception += Color.colorify(msg, Color.BOLD)
             self.handle_pyexpect_error(proc, exc)
+        except UnicodeDecodeError:
+            msg = 'Unicode Decode Error detected!\n'
+            self.exception = Color.colorify(msg, Color.BOLD)
+            self.handle_unicode_error(proc)
+            return
```
xvuko, line 195: I see a lot of code duplication in exception handling here. Those exact 4 lines appear 3 times. Could this try block be extended to unify catching errors from both harness and exec test?

trunner/testcase.py
```diff
@@ -126,6 +136,13 @@ def handle_pyexpect_error(self, proc, exc):
         self.exception += Color.colorify('GOT:\n', Color.BOLD)
         self.exception += f'{got}\n'
 
+    def handle_unicode_error(self, proc):
```
xvuko, line 139: I see no separation between private and public class members here. Just pointing that as I hardly see any underscore prefixed functions in trunner so this is out of scope of this review.

trunner/testcase.py
damianloew: Yeah, I've checked this and it can be done in one try block here. I would like to leave it as it is in handle function.

trunner/testcase.py
damianloew: I've created issue for this: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/331. I will think about it in meantime.

trunner/testcase.py
```diff
@@ -107,6 +112,11 @@ def exec_test(self, proc):
                 self.exception = Color.colorify(msg, Color.BOLD)
                 self.handle_pyexpect_error(proc, exc)
                 return
+            except UnicodeDecodeError:
+                msg = 'Unicode Decode Error detected!\n'
+                self.exception = Color.colorify(msg, Color.BOLD)
```
xvuko: Can be moved into handle unicode error.  Additionaly passing original exception into handler would allow to print original message that contains information about problematic character.

xvuko (changes requested):

trunner/testcase.py
```diff
@@ -90,23 +90,21 @@ def exec_test(self, proc):
         try:
             # Wait for a prompt
             proc.expect_exact('(psh)% ')
+            if self.exec_cmd:
+                if self.use_sysexec:
+                    self.sysexec(proc)
+                else:
+                    self.exec(proc)
         except (TIMEOUT, EOF) as exc:
             msg = 'Waiting for psh prompt failed!\n'
```
jsarzy, line 99: Now if execution fails we will get misleading error message. Change it to "exec failed" or similar.

damianloew: Thanks, I've missed it.

xvuko (approved):

PR #90, opened 2022-02-09
proc: add threads priority inversion test

Maxez:
## Description

Added test for checking threads priority inversion deadlock.
I'm going to implement priority inheritance in the kernel which should allow for the test to pass.

## Motivation and Context

[JIRA: DTR-269](https://jira.phoenix-rtos.com/browse/DTR-269)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic.

## Checklist:

## Special treatment


nalajcie:
not doing a review, just an idea: maybe we could make this test finish by itself by using `alarm()` (so, killing ourselfs within eg. 3s since start)? (I know this won't work on RT platforms right now due to broken signals :/). That way we might be able to make it an automatic test in the near future (or even now, just not unity-based). Please note that `mutexLock` should probably follow `pthread_mutex_lock` specification and never exit with `EINTR` (but `SIGALRM` default action is to terminate the process, so it should work anyway - and we could determine the test result just by looking at the exit code)

proc/test_priority.c
```diff
@@ -0,0 +1,158 @@
...
+#include <stdlib.h>
+#include <unistd.h>
+
+#include <sys/threads.h>
+
+
+typedef struct {
+	handle_t lock;
+	handle_t cond;
+	int done;
```
agkaminski, line 23:
```suggestion
	volatile int done;
```

proc/test_priority.c
```diff
@@ -0,0 +1,158 @@
...
+	mutexUnlock(priority_common.comp.lock);
+
+	endthread();
+}
+
+
+static void test_priority_mthr(void *arg)
+{
+	/* Busy waiting for test completion (starves the low priority thread) */
+	while (!(*(volatile int *)&priority_common.comp.done))
```
agkaminski, line 94:
```suggestion
	while (!priority_common.comp.done)
```

proc/test_priority.c
```diff
@@ -0,0 +1,158 @@
...
+		;
+
+	endthread();
+}
+
+
+static void test_priority_hthr(void *arg)
+{
+	/* Finish test completion */
+	/* Will not be able to take the test completion lock unless priority inversion is implemented */
```
agkaminski, line 104:
```suggestion
	/* Will not be able to take the test completion lock unless priority inversion fix is implemented */
```

proc/test_priority.c
```diff
@@ -0,0 +1,158 @@
...
+
+static void test_priority_lthr(void *arg)
+{
+	mutexLock(priority_common.comp.lock);
+
+	completion_finish(&priority_common.lcomp);
+	completion_wait(&priority_common.scomp);
+
+	/* Return the test completion lock */
+	/* Will not get here unless priority inversion is implemented */
```
agkaminski, line 83:
```suggestion
	/* Will not get here unless priority inversion fix is implemented */
```

proc/test_priority.c
```diff
@@ -0,0 +1,158 @@
...
+
+static void test_priority_inversion(void *arg)
+{
+	/* Init completions */
+	completion_init(&priority_common.comp);
+	completion_init(&priority_common.lcomp);
+	completion_init(&priority_common.scomp);
+
+	/* Run low priority thread */
+	beginthread(test_priority_lthr, 7, priority_common.stack[0], sizeof(priority_common.stack[0]), NULL);
```
agkaminski, line 119: 7 is reserved for idle kthread
```suggestion
	beginthread(test_priority_lthr, 6, priority_common.stack[0], sizeof(priority_common.stack[0]), NULL);
```

Maxez:
I've added running the test in a separate thread with and `sleep(2)` in the main thread. After 2s we assume deadlock and fail the test.

agkaminski (commented):
Test idea seems good to me

proc/test_priority.c
Maxez: Done.

agkaminski (approved):
LGTM

proc/test_priority.c
Maxez: Done.

proc/test_priority.c
Maxez: Done.

proc/test_priority.c
Maxez: Done.

proc/test_priority.c
Maxez: Done.

nalajcie:
IMHO we could plug it into CI in the current form, maybe it could be done by @niewim19 or @damianloew?

nalajcie:
> @nalajcie Good idea, maybe after this problem is solved?

Ideally I would like to have PR prepared showcasing the problem in CI (test failing) - and after the fix will be merged and upstreamed, we would re-run the actions in PR showing that the problem indeed existed and now is fixed.

But I'm not insisting if we don't have the manpower to do it ;).

damianloew:
> IMHO we could plug it into CI in the current form, maybe it could be done by @niewim19 or @damianloew?

There is my proposal. Temporary solution, when reading return code in psh is't available.
https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/91

PR #55, opened 2021-09-29
psh: add kill test and fix reading current target in psh test module

damianloew:
JIRA: PD-77

## Description

- psh: add kill test
- There is also helper executable, which creates 3 while test processes in background (to kill) 
## Motivation and Context

-  test to cover psh kill fixes in phoenix-rtos-kernel (https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/239)
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## How Has This Been Tested?

- [ ] Already covered by automatic testing.
- [ ] New test added: (add PR link here).
- [x] Tested by hand on: (ia32-generic - qemu).

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing linter checks and tests passed.
- [x] My changes generate no new compilation warnings for any of the targets.

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/239).
- [ ] I will merge this PR by myself when appropriate.

I will merge this PR after phoenix-rtos-kernel fixes implementation

PR #91, opened 2022-02-09
add running test_priority in CI process and syspage yaml property to trunner

damianloew:
DONE: CI-109

## Description

- execute and assert test_priority in a python harness test
- trunner: add syspage yaml property to allow loading binaries without executing it

## Motivation and Context

Simple solution that plugs test_priority into CI. For now there is no possibility to get return code from executed binary, so there is a temporary alternative option.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/proc/test.yaml
+++ b/proc/test.yaml
@@ -3,3 +3,6 @@ test:
         - name: test_thread_rand
           exec: test_thread_rand
           type: unit
+        - name: test_priority
+          exec: test_priority
+          harness: test_priority.py
--- a/proc/test.yaml
+++ b/proc/test.yaml
@@ -4,5 +4,5 @@ test:
           exec: test_thread_rand
           type: unit
         - name: test_priority
-          exec: test_priority
+          syspage: test_priority
           harness: test_priority.py
--- a/proc/test_priority.py
+++ b/proc/test_priority.py
@@ -1,0 +1,22 @@
+# Phoenix-RTOS
+#
+# phoenix-rtos-tests
+#
+# simple harness test asserting test_priority executable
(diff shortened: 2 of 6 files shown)


proc/test.yaml
```diff
@@ -3,3 +3,7 @@ test:
         - name: test_thread_rand
           exec: test_thread_rand
           type: unit
+        - name: test_priority
+          harness: test_priority.py
+          targets:
+            exclude: [armv7m7-imxrt106x, armv7m7-imxrt117x]
```
agkaminski, line 9: We don't need to exclude targets, this test should work on all targets

agkaminski (commented):
Efficient approach 👍

proc/test.yaml
damianloew: Oh, sorry I didn't see that `alarm()` wasn't used. I will enable it, but have to wait for https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/55 merge (it fixes `psh.assert_exec()` on IMXRT platforms).

agkaminski (commented):
Why we need to execute this one specifically from the syspage?

damianloew:
> Why we need to execute this one specifically from the syspage?

Now there is yaml property called exec. Thanks to it additional binaries can be loaded using plo in test runner, but they are executed at once, before executing code from harness test. I could  only check last status in test, but I think executing it in test is more readable. It can also be helpful for future tests, where executing some binaries more times may be needed.

jsarzy (approved):

PR #92, opened 2022-02-11
runner: add armv7a9-zynq7000-emu implementation and enable test-mkdir for imxrt targets

damianloew:
## Description

- runner: add armv7a9-zynq7000-emu implementation
- psh: disable asserting ownership in mkdir test and enable the test on dummyfs targets
- psh: test-mkdir: divide fname with all chars

## Motivation and Context

JIRA: CI-110
JIRA: CI-120

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt eval kits, zynq7000 qemu).

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-project/pull/339).

### DIFF
--- a/psh/test.yaml
+++ b/psh/test.yaml
@@ -1,6 +1,9 @@
 test:
-    tests:
+    targets:
+      #TODO: remove when zynq7000-qemu target will stop being experimental
+      include: [armv7a9-zynq7000-qemu]
 
+    tests:
         - name: auth
           harness: test-auth.py
 
@@ -41,6 +44,7 @@ test:
               exclude:
                   - armv7m7-imxrt106x
                   - armv7m7-imxrt117x
+                  - armv7a9-zynq7000-qemu
 
         - name: ls
           harness: test_ls.py
--- a/psh/test.yaml
+++ b/psh/test.yaml
@@ -40,11 +40,6 @@ test:
 
         - name: mkdir
           harness: test-mkdir.py
-          targets:
-              exclude:
-                  - armv7m7-imxrt106x
-                  - armv7m7-imxrt117x
-                  - armv7a9-zynq7000-qemu
 
         - name: ls
           harness: test_ls.py
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -30,13 +30,21 @@ PHRTOS_TEST_DIR = PHRTOS_PROJECT_DIR / 'phoenix-rtos-tests'
 PYEXPECT_TIMEOUT = 8
 
 # Available targets for test runner.
-ALL_TARGETS = ['ia32-generic', 'host-pc', 'armv7m7-imxrt106x', 'armv7m7-imxrt117x', 'armv7m4-stm32l4']
+ALL_TARGETS = ['ia32-generic',
+               'armv7a9-zynq7000-qemu',
+               'host-pc',
+               'armv7m7-imxrt106x',
+               'armv7m7-imxrt117x',
+               'armv7m4-stm32l4']
 
+EXPERIMENTAL_TARGETS = ['armv7m4-stm32l4', 'armv7a9-zynq7000-qemu']
 # Default targets used by parser if 'target' value is absent
 DEFAULT_TARGETS = [target for target in ALL_TARGETS
-                   if target not in ('host-pc', 'armv7m4-stm32l4')]
+                   if target not in EXPERIMENTAL_TARGETS + ['host-pc']]
 
-DEVICE_TARGETS = ['armv7m7-imxrt106x', 'armv7m7-imxrt117x']
+SYSEXEC_TARGETS = ['armv7m7-imxrt106x', 'armv7m7-imxrt117x', 'armv7a9-zynq7000-qemu']
+
+QEMU_TARGETS = ['ia32-generic', 'armv7a9-zynq7000-qemu']
 
 CURRENT_TARGET = None
 
--- a/trunner/device.py
+++ b/trunner/device.py
@@ -1,19 +1,14 @@
 # Imports of available runners
 from .runners import HostRunner, QemuRunner, IMXRT106xRunner, IMXRT117xRunner, STM32L4Runner
 
-from .config import PHRTOS_PROJECT_DIR, DEVICE_SERIAL_USB, PHOENIXD_PORT
-
-
-QEMU_CMD = {
-    'ia32-generic': f'{PHRTOS_PROJECT_DIR}/scripts/ia32-generic-test.sh'
-}
+from .config import DEVICE_SERIAL_USB, PHOENIXD_PORT, QEMU_TARGETS
 
 
 class RunnerFactory:
     @staticmethod
     def create(target, serial, log=False):
-        if target == 'ia32-generic':
-            return QemuRunner(QEMU_CMD[target], log=log)
+        if target in QEMU_TARGETS:
+            return QemuRunner(target, log=log)
         if target == 'host-pc':
             return HostRunner(log=log)
         if target == 'armv7m7-imxrt106x':
--- a/psh/test-mkdir.py
+++ b/psh/test-mkdir.py
@@ -58,7 +58,8 @@ def mkdir(p, dir):
 def assert_dir_present(dir, files):
     for file in files:
         if dir == file.name:
-            assert file.owner == 'root', f'{dir} not owned by root'
+            # TODO: enable after adding multiple user support
+            # assert file.owner == 'root', f'{dir} not owned by root'
             assert file.is_dir, f'{dir} is not directory'
             break
     else:
--- a/psh/test-mkdir.py
+++ b/psh/test-mkdir.py
@@ -112,7 +112,9 @@ def harness(p):
     assert files == ls(p, 'example_dir/..')
 
     assert_dir_created(p, 'example_dir/another_dir')
-    assert_dir_created(p, ''.join(chars))
+    # there are some targets, where max fname length equals 64
+    assert_dir_created(p, ''.join(chars[:50]))
+    assert_dir_created(p, ''.join(chars[50:]))
 
     assert_random_dirs(p, chars)
 
--- a/psh/tools/psh.py
+++ b/psh/tools/psh.py
@@ -84,7 +84,7 @@ def assert_exec(pexpect_proc, prog, expected='', msg=''):
     ''' Executes specified program and asserts that it's displayed correctly
     with optional expected output and next prompt'''
 
-    if config.CURRENT_TARGET in config.DEVICE_TARGETS:
+    if config.CURRENT_TARGET in config.SYSEXEC_TARGETS:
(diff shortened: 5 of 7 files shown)


trunner/config.py
```diff
@@ -30,13 +30,18 @@ def resolve_phrtos_dir() -> Path:
 PYEXPECT_TIMEOUT = 8
 
 # Available targets for test runner.
-ALL_TARGETS = ['ia32-generic', 'host-pc', 'armv7m7-imxrt106x', 'armv7m7-imxrt117x', 'armv7m4-stm32l4']
+ALL_TARGETS = ['ia32-generic',
+               'armv7a9-zynq7000-emu',
```
nalajcie, line 34: I'm against adding here targets not defined in `phoenix-rtos-project`. It will:
 - increase confusion regarding the test names
 - break the build functionality of the test runner
 - the third part of the target is for the `project` part of the target

I assume the binaries/images for emulator and target HW might need to be different - but if that would be the case - we need to introduce different targets in `phoenix-rtos-project` repo first.

sample/test/fails/test.yaml
```diff
@@ -9,19 +9,29 @@ test:
         - name: fail-output-1
           exec: test-fail-output-1
           harness: test-fail-output-1.py
+          targets:
+              exclude: [armv7a9-zynq7000-emu]
```
nalajcie, line 13: If You need to exclude the new target in almost all tests **it shouldn't be put in DEFAULT_TARGETS** for now.

.github/workflows/ci.yml
```diff
@@ -110,6 +110,7 @@ jobs:
         with:
           target: ${{ matrix.target }}
 
+
```
nalajcie, line 151: unrelated change in this commit

trunner/config.py
```diff
@@ -30,7 +30,7 @@ def resolve_phrtos_dir() -> Path:
 PYEXPECT_TIMEOUT = 8
 
 # Available targets for test runner.
-ALL_TARGETS = ['ia32-generic',
+ALL_TARGETS = ['ia32-generic-emu',
```
nalajcie, line 33: nah, don't do it (see my reasoning below)

trunner/device.py
```diff
@@ -12,14 +12,25 @@
...
+        ]
     )
 }
 
 
 class RunnerFactory:
     @staticmethod
     def create(target, serial, log=False):
-        if target == 'ia32-generic':
+        if target == 'ia32-generic' or target == 'armv7a9-zynq7000-emu':
```
nalajcie:
```python
if target in [ 'ia32-generic', 'armv7a9-zynq7000']:
```
or even specify `QEMU_TARGETS = [..]`

Should probably be upgraded soon to device-configs.

runner.py
```diff
@@ -88,6 +88,9 @@ def parse_args():
     if not args.target:
         # Run on all available targets
         args.target = config.ALL_TARGETS
+    # for now there is only test runner for ia32-generic emulator
```
nalajcie, line 91: should not be needed

.github/workflows/ci.yml
```diff
@@ -110,6 +110,44 @@ jobs:
         with:
           target: ${{ matrix.target }}
 
+  test-emu:
```
nalajcie, line 113: nah, don't ^C^V, don't jungle target names in CI, thus would introduce only more confusion

.github/workflows/ci.yml
```diff
@@ -71,15 +71,15 @@ jobs:
             _boot
             rootfs-${{ matrix.target }}.tar
 
-  test-emu:
+  test-host:
     needs: build
-    name: run tests on emulators
+    name: run tests on ubuntu host
```
nalajcie, line 76: coincidentally true, but not because there is ubuntu specified in the next line (but because phoenixrtos/devel:latest docker image uses ubuntu:20.04 as a base image)

nalajcie (changes requested):

sample/test/fails/test.yaml
damianloew: That's right, including it in psh for psh tests would be better

.github/workflows/ci.yml
damianloew: Thanks for review, this and the remaining changes above won't be necessary, if there are two separate targets in phoenix-rtos-poject. I've adjusted testing process for that, but I have to talk about it with Hubert. Now there is a test pr with changed zynq target name and this branch to run ci properly (https://github.com/phoenix-rtos/phoenix-rtos-project/pull/381).

trunner/config.py
damianloew: I propose `armv7a9-zynq7000emu` and `armv7a9-zynq7000hw`.

nalajcie: nah, the TARGET value should be `${TARGET_FAMILY}-${TARGET_SUBFAMILY}-${PROJECT}` - where `TARGET_SUBFAMILY` usually should mean target CPU. I would probably go for `armv7a9-zynq7000-qemu` and `armv7a9-zynq7000-zedboard` or similar (not sure if we will have multiple qemu targets for `zynq7000` or if `zedboard` is unique enough. Please consult with @HBuczynski regarding how to approach it.

damianloew: Ok, I will consult it and then make changes.

HBuczynski: I agree with @nalajcie, `armv7a9-zynq7000-qemu` and `armv7a9-zynq7000-zedboard` is fine. There is a one target for `qemu` and the `zedboard` is a unique name for the development board from `digilent`.

jsarzy (dismissed):
lgtm

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       build_result: ${{ steps.build.outcome }}
     strategy:
       matrix:
-        target: ['armv7a7-imx6ull', 'armv7a9-zynq7000', 'armv7m7-imxrt106x', 'armv7m7-imxrt117x', 'armv7m4-stm32l4x6', 'host-pc', 'ia32-generic', 'riscv64-spike', 'riscv64-virt']
+        target: ['armv7a7-imx6ull', 'armv7a9-zynq7000-zedboard', 'armv7a9-zynq7000-qemu', 'armv7m7-imxrt106x', 'armv7m7-imxrt117x', 'armv7m4-stm32l4x6', 'host-pc', 'ia32-generic', 'riscv64-spike', 'riscv64-virt']
```
nalajcie, line 24: this needs to be moved to `phoenix-rtos-project` due to reusable workflows introduction

psh/test.yaml
```diff
@@ -41,6 +43,7 @@ test:
               exclude:
                   - armv7m7-imxrt106x
                   - armv7m7-imxrt117x
+                  - armv7a9-zynq7000-qemu
```
nalajcie, line 46: @HBuczynski I assume it's not yet possible to create dirs on zynq-qemu?

trunner/device.py
```diff
@@ -1,18 +1,19 @@
 # Imports of available runners
 from .runners import HostRunner, QemuRunner, IMXRT106xRunner, IMXRT117xRunner, STM32L4Runner
 
-from .config import PHRTOS_PROJECT_DIR, DEVICE_SERIAL_USB, PHOENIXD_PORT
+from .config import PHRTOS_PROJECT_DIR, DEVICE_SERIAL_USB, PHOENIXD_PORT, QEMU_TARGETS
 
 
 QEMU_CMD = {
```
nalajcie, line 7: not directly related to this PR but probably should be (sooner or later) moved to `QemuRunner` class

nalajcie (commented):
quite nice, but we need to move CI changes to `phoenix-rtos-project`. Left some other small remarks which might be implemented.

psh/test.yaml
damianloew: Creating dirs works fine (test ls uses this functionality), but there may be other problems, like those related to ownership.
Screenshot from 2022-04-25 19-01-53
:

trunner/device.py
damianloew: Do You mean `QEMU_CMD` should be moved?

trunner/config.py
```diff
@@ -30,13 +30,20 @@ def resolve_phrtos_dir() -> Path:
...
+               'armv7a9-zynq7000-qemu',
+               'host-pc',
+               'armv7m7-imxrt106x',
+               'armv7m7-imxrt117x',
+               'armv7m4-stm32l4']
 
 # Default targets used by parser if 'target' value is absent
 DEFAULT_TARGETS = [target for target in ALL_TARGETS
-                   if target not in ('host-pc', 'armv7m4-stm32l4')]
+                   if target not in ('host-pc', 'armv7m4-stm32l4', 'armv7a9-zynq7000-qemu')]
```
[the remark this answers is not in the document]
damianloew: Yeah, I think it can be more readable, applied.

damianloew:
> quite nice, but we need to move CI changes to `phoenix-rtos-project`. Left some other small remarks which might be implemented.

Thanks, I've applied those changes. I will add CI related things in `phoenix-rtos-project` PR after merge. I know that branch name is not valid now, but I hope it isn't a problem.

psh/test.yaml
HBuczynski: I confirm that `mkdir` works fine on `zynq-qemu`.

psh/test.yaml
```diff
@@ -1,6 +1,8 @@
 test:
-    tests:
+    targets:
+      include: [armv7a9-zynq7000-qemu]
```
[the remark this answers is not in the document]
HBuczynski: Why should the `armv7a9-zynq7000-qemu` be set as experimental?

trunner/device.py
nalajcie: It should probably internal to `QemuRunner` class (so it can be instantiated by the target directly

psh/test.yaml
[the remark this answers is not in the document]
nalajcie: My understanding is that most of the tests don't succeed on `zynq7000-qemu` - based on previous iteration of this PR where zynq was explicitly excluded from almost all tests. So for now we mark `zynq` as exprimental and opt for explicitly include it in tests that do pass.

psh/test.yaml
nalajcie: ok, so should pass but doesn't - seems like a bug? (either in the target implementation or in test) - we should fix it in instead of not running the test :).

trunner/device.py
damianloew: Do You mean sth I've done now, yes?

psh/test.yaml
damianloew: Apart from disabling checking ownership in the test (not needed for now, no multi-user support), I had to make one file name shorter to enable the test for all targets.

nalajcie (approved):

PR #96, opened 2022-02-23
tests: auth timing loophole fix

niewim19:
DONE: CI-114

## Description

Fix waiting for psh prompt in login test

## Motivation and Context

Fixture:
-`auth` applet, after passing empty password shall wait 1 second and then close, effectively going back to psh (showing psh prompt)
- `assert_login_fail()` uses `assert_prompt_fail()` which guarantees that for given time (1 second here) psh prompt **doesn`t** appear on screen
- after that the new test commences where `auth` is written into console

If psh prompt appears a liitle bit late than 1s, the test can already assert its absence and immediately write `auth` to console which will be written into _nothing_ as the command interpreter is not yet present. Thus it fails next test case which thinks it properly called `auth` command when it did not. 

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/psh/test-auth.py
+++ b/psh/test-auth.py
@@ -54,18 +54,18 @@ def harness(p):
 
     # Wrong credentials
     assert_auth(p)
-    logintools.assert_login_fail(p, cred_ok.user, cred_bad.passwd)
+    logintools.assert_login_fail(p, cred_ok.user, cred_bad.passwd, expect_psh_afterwards=True)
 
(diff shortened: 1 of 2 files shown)


damianloew (changes requested):
Small typo: exect -> expect

damianloew (approved):
Looks good to me, I think it will solve the problem.

PR #104, opened 2022-03-25
use ia32 generic testing script and handle programs with arguments in test-ps

damianloew:
## Description

- trunner: add using testing script for launching ia32-generic emualtor
- psh: handle programs with arguments in test-ps

## Motivation and Context

JIRA: CI-116

Now the command for running qemu (ia32-generic target) is now placed strictly in test runner:
Screenshot from 2022-03-25 12-30-04

It may be better to create the proper script in `scripts` directory that could be used by a test runner. If there is a need for change in qemu running it will be done only in the `scripts` directory (it's easier to remember about changing similar script in the same directory than changing some file in tests repository).

After enabling testing script the lwip with arguments will be present in ps output and `ValueError` will occur. The second commit fixes it.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-project/pull/367).

### DIFF
--- a/psh/test-ps.py
+++ b/psh/test-ps.py
@@ -35,7 +35,10 @@ def harness(p):
             try:
                 pid, ppid, pr, state, cpu, wait, time, vmem, thr, task = line.split()
             except ValueError:
-                assert False, f'wrong ps output: {line}'
+                try:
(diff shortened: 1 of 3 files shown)


niewim19 (approved):
LGTM

PR #110, opened 2022-05-16
adapt test runner and ls tests to a new armv7a9-zynq7000

damianloew:
## Description

- psh: ls: move size test case to test-ls-rootfs
- psh: ls: make -f arg case more flexible
- trunner: adapt timeout and encoding to a new armv7a9-zynq7000

## Motivation and Context

- Default timeout for pyexpect was too low for jffs init on armv7a9-zynq7000-qemu
- Because of non-ascii character in init logs (©) switch back to utf-8 encoding in psh
- -f argument test case assumed that not sorted files are listed by creation time, which is wrong for jffs. Checking files order in this case has been disabled.
- The correction of size sorting was checked by listing an empty file and an empty directory.
    Assumption that directory has larger size than a file is wrong for jffs.
    Now size size sorting assertion is done by comparing empty file with psh binary in /bin.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (armv7a9-zynq7000-qemu with newest mtd changes).

## Checklist:

## Special treatment

### DIFF
--- a/psh/test-ls-rootfs.py
+++ b/psh/test-ls-rootfs.py
@@ -65,6 +65,21 @@ def assert_ls_t(p):
     psh.assert_cmd(p, f'ls -t {ROOT_TEST_DIR}', expected_pattern, msg, is_regex=True)
 
 
+def assert_ls_S(p):
+    expected_pattern = r'.*?psh' + SEPARATOR_PATTERN + r'.*?empty_file' + SEPARATOR_PATTERN + r'.*?'
+    expected_pattern = expected_pattern + psh.EOL
+
+    psh.assert_cmd(p, 'touch /bin/empty_file', '', 'Wrong output when creating empty file!')
+
+    # TODO: use only newly created files, when it will be possible to write content to a file
+    psh.assert_cmd(
+        pexpect_proc=p,
+        cmd='ls -1S /bin',
+        expected=expected_pattern,
+        msg='Wrong output, when calling `ls -1S`',
+        is_regex=True)
+
+
 def assert_multi(p):
     files = []
     for i in range(20):
@@ -101,6 +116,6 @@ def harness(p):
     assert_extralong(p)
 
     psh_cmds = psh.get_commands(p)
-    if 'date' in psh_cmds:
-        assert_ls_t(p)
+    assert_ls_t(p)
+    assert_ls_S(p)
     assert_ls_pshcmds(p, psh_cmds)
--- a/psh/test_ls.py
+++ b/psh/test_ls.py
@@ -39,15 +39,30 @@ def prepare_testdir(p, files):
     psh.assert_cmd(p, f'mkdir {TEST_DIR_BASIC}/dir', '', msg)
 
 
-def create_pattern(files, separate_lines=False):
+def create_unordered_pattern(files):
+    """ Create regex pattern for ls output
+    containing all files from the `files` list in any order """
+
+    filesCnt = len(files)
+    files = '|'.join(files)
+    pattern = f'({OPTIONAL_CONTROL_CODE}({files}){SEPARATOR_PATTERN})'
+    pattern += rf'{{{filesCnt}}}'
+
+    # ( optional CC + (file1|file2|filen) + ((wss) + CC)|(CC + wss)|(wss)) ){n}
+    # wss - white spaces, CC - control code
+    return pattern
+
+
+def create_ordered_pattern(files, separate_lines=False):
+    """ Create regex pattern for ls output
+    containing all files from the `files` list in the same order """
+
     newline_separator = OPTIONAL_CONTROL_CODE + r'(\r+\n)'
-    pattern = r''
-    for file in files:
-        pattern += OPTIONAL_CONTROL_CODE + f'{file}'
-        if separate_lines:
-            pattern += newline_separator
-        else:
-            pattern += SEPARATOR_PATTERN
+    sep = newline_separator if separate_lines else SEPARATOR_PATTERN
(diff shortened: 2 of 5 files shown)


psh/test_ls.py
```diff
@@ -39,7 +39,23 @@ def prepare_testdir(p, files):
     psh.assert_cmd(p, f'mkdir {TEST_DIR_BASIC}/dir', '', msg)
 
 
-def create_pattern(files, separate_lines=False):
+def create_soft_pattern(files):
```
niewim19, line 42: `create_soft_pattern()` and `create_strict_pattern()` can be joined into one function with one parameter like 'ordering=strict/soft`. Please add more info in docstring about this option purpose.

trunner/testcase.py
```diff
@@ -89,7 +89,8 @@ def sysexec(self, proc):
     def exec_test(self, proc):
         try:
             # Wait for a prompt
-            proc.expect_exact('(psh)% ')
+            # On armv7a9-zynq7000-qemu jffs initialization may take to ~20s, that's why it's set to 25
+            proc.expect_exact('(psh)% ', timeout=25)
```
niewim19, line 93: If this currently happen only on `armv7a9-zynq7000-qemu` maybe increase timeout only for this target? under self there should be a target.
```suggestion
            # comment why we need longer wait for armv7a9-zynq7000-qemu
            prompt_timeout = 25 if self.target == "armv7a9-zynq7000-qemu" else 8
            proc.expect_exact('(psh)% ', timeout=prompt_timeout)
```
This may need additional parametrization in the future as there may be a need for different waiting times for different targets.

psh/test-ls-rootfs.py
```diff
@@ -65,6 +65,17 @@ def assert_ls_t(p):
     psh.assert_cmd(p, f'ls -t {ROOT_TEST_DIR}', expected_pattern, msg, is_regex=True)
 
 
+def assert_ls_S(p):
```
niewim19, line 68: Idea of different tests for rootfs and non-rootfs targets is all we can do for now, but as it is a common difference between targets maybe some parametrization should be introduced at the .yaml/test_runner level for differentiating between these two.

I bet that some of such splitted tests will be skipped when targets without rootfs will get them (at some point in the future) because it will require for test developer to remember that when the transition occurs such tests should be merged/turned on. And as it is done with `include` in yaml file, each splitted test will require to add new target parameter.

psh/test-ls-rootfs.py
```diff
@@ -65,6 +65,17 @@ def assert_ls_t(p):
...
+    expected_pattern = r'.*?psh' + SEPARATOR_PATTERN + r'.*?empty_file' + SEPARATOR_PATTERN + r'.*?'
+    expected_pattern = expected_pattern + psh.EOL
+
+    psh.assert_cmd(p, 'touch /bin/empty_file', '', 'Wrong output when creating empty file!')
+
+    msg = 'Wrong output, when calling `ls -1S`'
+    # TODO: use only newly created files, when it will be possible to write content to a file
+    psh.assert_cmd(p, 'ls -1S /bin', expected_pattern, msg, is_regex=True)
```
niewim19, line 76:
```suggestion
    # TODO: use only newly created files, when it will be possible to write content to a file
    psh.assert_cmd(
        pexpect_proc=p, 
        cmd='ls -1S /bin', 
        expected=expected_pattern, 
        msg='Wrong output, when calling `ls -1S`',
        is_regex=True)
```

does this fit under "120 char max python line` rule?

niewim19 (changes requested):
Added some remarks. 

Two biggest are about parametrization of prompt waiting time and rootfs/non-rootfs targets tests splitting.

Apart from that minor code tweaks. Please add little bit more comments about regexp creation techniques for strict/soft ordering as regexes may be sometimes difficult to understand when presented in final form, not mentioning dynamically creating them. If we change something in `ls` and these regexes will need to change a good documantation of _how its done_ will surely help :)

psh/test_ls.py
damianloew: I would like to leave these 2 functions here. I tried to do it in `create_pattern()`, but the result was `spaghetti code` with nested if's. I think that used approach is much more readable.

trunner/testcase.py
damianloew: Thanks, I think it may be a better solution.

psh/test-ls-rootfs.py
damianloew: Yes, maybe it's also a better solution, thanks.

psh/test-ls-rootfs.py
damianloew: Yes, I agree that it could be done automatically in the future.

psh/test_ls.py
```diff
@@ -39,7 +39,27 @@ def prepare_testdir(p, files):
...
+    """ Create regex pattern for ls output
+    containing all files from the `files` list in any order """
+    pattern = r''
+    pattern += '(' + OPTIONAL_CONTROL_CODE + '('
+    for file in files:
+        pattern += file
+        if file == files[-1]:
+            pattern += ')'
+        else:
+            pattern += '|'
```
niewim19, line 52: One can use .join function

https://www.w3schools.com/python/ref_string_join.asp

niewim19: the same applies for ordered function

niewim19 (changes requested):

psh/test_ls.py
damianloew: Thanks, applied.

niewim19 (approved):
LGTM

PR #119, opened 2022-07-07
trunner: add long test interface and support running mbedtls tests

damianloew:
## Description

 - trunner: add the interface for adding long tests
 - mbedtls: add support for mbedtls long tests

## Motivation and Context

- make adding new long tests easier
- JIRA: PD-292

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (`ia32-generic-qemu`: running both `busybox` and `mbedtls` test suites).

## Checklist:

## Special treatment


trunner/long_test.py
```diff
@@ -0,0 +1,140 @@
+#
```
jsarzy, line 1: Please do not name it as `long_test.py`. Maybe create directory named `harnesses` and there we can have `base.py` for base classes, `unity.py` for Unity, `busybox.py` for Busybox and `mbedtls.py` for Mbedtls tests.

trunner/long_test.py
```diff
@@ -0,0 +1,140 @@
...
+from .tools.color import Color
+
+
+LONG_TESTS = ['busybox', 'mbedtls']
+
+
+class LongTestHarnessFactory:
+    @staticmethod
+    def create(ltest_type):
+        if (ltest_type == 'busybox'):
```
jsarzy, line 19: no need for parenthesis

trunner/long_test.py
```diff
@@ -0,0 +1,140 @@
...
+            return BusyboxTestHarness.harness
+        if (ltest_type == 'mbedtls'):
```
jsarzy, line 21: no need for parenthesis

trunner/long_test.py
```diff
@@ -0,0 +1,140 @@
...
+# Authors: Damian Loewnau
+#
+
+from .tools.color import Color
+
+
+LONG_TESTS = ['busybox', 'mbedtls']
+
+
+class LongTestHarnessFactory:
```
jsarzy, line 16: We do not need this class only for "long tests". We could create factory for "3rd party" harnesses (unity, busybox, embetls) and reduce duplicated code using inheritance

trunner/long_test.py
```diff
@@ -0,0 +1,140 @@
...
+                # We ended processing test result and message
+                if msg and test['status'] == 'FAIL':
+                    test['msg'] = msg
+                    msg = ''
+
+                test['name'] = test['name'].replace('.', '')
+                test_results.append(LongTestResult(**test))
+            else:
+                line = groups[0]
+                msg += '\t\t' + line + '\n'
```
jsarzy, line 94: It works ok? In the first loop iteration `test` is `None` and (as I suppose) `idx != 2` so `groups[0]` is not line output. Why it won't work the same as in busybox harness where we have `elif idx == 2`?

trunner/testcase.py
```diff
@@ -261,11 +262,12 @@ def handle(self, proc):
         return res
 
 
-class TestCaseBusybox(TestCase):
-    """The test case representing Busybox test suite."""
+class TestCaseLong(TestCase):
+    """The test case representing Mbedtls test suite."""
```
jsarzy, line 266: `Mbetls` or `Long test`?

trunner/long_test.py
```diff
@@ -0,0 +1,140 @@
...
+
+class MbedtlsTestHarness:
+    """Class providing harness for parsing output of Mbedtls tests"""
+
+    # common result line: 'Hash: RIPEMD160 ................................................... PASS'
+    RESULT = r"(?P<test_name>.{67})\s(?P<status>(PASS|----|FAILED))\r+\n"
+    # after failed test case there is some additional output
+    MSG_LINE = r"  ([^\r\n]+?)\r+\n"
+    # common final line: 'FAILED (16 / 21 tests (6 skipped))'
+    FINAL = r"(?P<status>FAILED|PASSED)\s\((?P<nr>\d+)\s/\s(?P<total_nr>\d+)\stests\s\(\d+\sskipped\)\)"
```
jsarzy, line 65: Please add test cases to `trunner/test` that covers `MbedtlsTestHarness.harness` method. Now the long test are not running up on PRs so I can't even check how the output of Mbedtls tests looks like without setting up the zynq target with ports. This is also a good way to document the cases of the Mbedtls tests output (e.g. passed tests, failed tests with msg, failed tests with multiline message, skipped tests)

trunner/testcase.py
```diff
@@ -261,11 +262,12 @@ def handle(self, proc):
         return res
 
 
-class TestCaseBusybox(TestCase):
-    """The test case representing Busybox test suite."""
+class TestCaseLong(TestCase):
```
jsarzy, line 265: If I'm looking right, `TestCaseLong` and `TestCaseUnit` are the same, maybe we should share to code by a common class?

mbedtls/test.yaml
```diff
@@ -0,0 +1,201 @@
...
+        - name: test_suite_mdx
+          exec: test_suite_mdx /mbedtls_test_configs/test_suite_mdx.datax
+        - name: test_suite_memory_buffer_alloc
+          exec: test_suite_memory_buffer_alloc /mbedtls_test_configs/test_suite_memory_buffer_alloc.datax
+        - name: test_suite_mpi
+          exec: test_suite_mpi /mbedtls_test_configs/test_suite_mpi.datax
+        - name: test_suite_mps
+          exec: test_suite_mps /mbedtls_test_configs/test_suite_mps.datax
+        - name: test_suite_net
+          ignore: True #issue #408
```
jsarzy, line 123: maybe add a link to issue

mbedtls/test.yaml
```diff
@@ -0,0 +1,201 @@
...
+        - name: test_suite_psa_crypto_slot_management
+          exec: test_suite_psa_crypto_slot_management /mbedtls_test_configs/test_suite_psa_crypto_slot_management.datax
+        - name: test_suite_psa_crypto_storage_format.current
+          exec: test_suite_psa_crypto_storage_format.current /mbedtls_test_configs/test_suite_psa_crypto_storage_format.current.datax
+        - name: test_suite_psa_crypto_storage_format.misc
+          exec: test_suite_psa_crypto_storage_format.misc /mbedtls_test_configs/test_suite_psa_crypto_storage_format.misc.datax
+        - name: test_suite_psa_crypto_storage_format.v0
+          exec: test_suite_psa_crypto_storage_format.v0 /mbedtls_test_configs/test_suite_psa_crypto_storage_format.v0.datax
+        - name: test_suite_psa_its
+          ignore: True #issue #409
```
jsarzy, line 182: maybe add a link to issue

jsarzy (changes requested):
Some thoughts about the code structure, I think we should discuss it on a call. I will write you pm :)

trunner/testcase.py
damianloew: Yeah, `Long test`

trunner/long_test.py
damianloew: I will prefer to treat `busybox`, `mbedtls` and other tests of our ports as `long_tests` and place their harnesses in one file, but I think the idea of creating `harnesses` directory could be a good solution.

trunner/testcase.py
damianloew: After switching to `3rdpartyTestHarnessFactory` it can be combined into one.

trunner/long_test.py
damianloew: The goal here was to make possible adding new test suite by changing possibly one file. 

EDIT: I am not sure about splitting it into files, like `unity.py`, `busybox.py`, because the only thing in those files will be one class with one harness method.

damianloew:
> Some thoughts about the code structure, I think we should discuss it on a call. I will write you pm :)

I've restructured the code. In the last commit is my proposal. I will also add the tests to cover MbedtlsTestHarness.harness

trunner/testcase.py
```diff
@@ -235,54 +238,14 @@ def __init__(
...
-        self.harness = BusyboxTestHarness.harness
+        self.harness = TestHarnessFactory.create(type)
         self.test_results = []
 
     def log_test_status(self):
         super().log_test_status()
 
         for test in self.test_results:
-            if test.status == BusyboxTestResult.FAIL:
+            if test.status == TestResult.FAIL:
```
jsarzy, line 248:
```suggestion
            if test.status == self.FAIL:
```

trunner/testcase.py
```diff
@@ -294,7 +257,7 @@ def handle(self, proc):
             self.test_results = res
 
         for test in self.test_results:
-            if test.status == BusyboxTestResult.FAIL:
+            if test.status == TestResult.FAIL:
```
jsarzy, line 260:
```suggestion
            if test.status == self.FAIL:
```

trunner/test/test_harness.py
```diff
@@ -0,0 +1,199 @@
...
+
+    return testcases_info
+
+
+def assert_result(result, state=TestResult.PASS, names=[], fails_info=[]):
+
+    color = {
+        TestResult.PASS: Color.OK,
+        TestResult.IGNORE: Color.SKIP,
+        TestResult.FAIL: Color.FAIL}
```
jsarzy, line 37: nit: new line after Color.FAIL

trunner/test/test_harness.py
```diff
@@ -0,0 +1,199 @@
...
+
+    FIFO = "./test_harness.fifo"
+
+    @pytest.fixture(scope="class")
+    def fifo(self):
+        os.mkfifo(self.FIFO)
+        yield
+        os.unlink(self.FIFO)
+
+    def get_result(self, writer_thr):
```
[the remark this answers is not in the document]
damianloew: Yeah, it may be better

trunner/test/test_harness.py
```diff
@@ -0,0 +1,199 @@
...
+import pytest
+import pexpect.fdpexpect
+
+from trunner.harnesses.mbedtls import MbedtlsTestHarness
+from trunner.harnesses.common import TestResult
+
+from trunner.tools.color import Color
+
+
+def prepare_info(testcases_info=[]):
```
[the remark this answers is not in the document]
damianloew: I prefer the first solution to show how a sample output looks like.

trunner/test/test_harness.py
```diff
@@ -0,0 +1,199 @@
...
+    print('prepare')
+    for i, info in enumerate(testcases_info):
+        print(i)
+        testcases_info[i] = testcases_info[i].replace('  ', '\t\t')
+        testcases_info[i] = testcases_info[i].replace('\r', '')
+
+    return testcases_info
+
+
+def assert_result(result, state=TestResult.PASS, names=[], fails_info=[]):
```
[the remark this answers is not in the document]
damianloew: I see, this is much better

trunner/test/test_harness.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # Release lock for writer to close fifo
+        lock.release()
+        thread.join()
+        fifo_r.close()
+
+        return result
+
+    def test_pass(self, fifo):
+        names = ['Hash: RIPEMD160',
```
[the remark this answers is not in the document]
damianloew: It was little bit harder to build input from expected (`TestResult` type). In the next commit will be the list of dictionaries to build `input` and `expected`.

trunner/test/test_harness.py
```diff
@@ -0,0 +1,199 @@
...
+
+    for i in range(len(names)):
+        testcase_result = f'{result[i]}'
+        expected = f'{status}: {names[i]}'
+        if state == TestResult.FAIL:
+            expected += '\n\t\tOutput of the failed test case:\n\t\t---\n'
+            expected += f'{fails_info[i]}'
+            expected += '\t\t---'
+
+        assert testcase_result == expected
```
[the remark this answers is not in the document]
damianloew: There was some problem with using `itertools`, so I've decided to use `for i in range(len(expected)):` and `result[i].status`.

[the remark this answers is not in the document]
damianloew: > Testcases should be `TestResult` type, not some `names` and `testcase_info`, we also do not need to copy `__str__` method from `TestResult`, just compare two instance of `TestResult`
> 
> We did not cover case when embtls outputs something that we do not expect.
> 
> Overall looks good, nice to have harness test :)

Thanks, I've applied changes. `TestResult` is now used, please check the latest commit.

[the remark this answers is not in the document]
jsarzy: > There was some problem with using `itertools`, so I've decided to use `for i in range(len(expected)):` and `result[i].status`.

Yeap, I meant `zip(result, expected)

trunner/test/test_harness.py
```diff
@@ -0,0 +1,240 @@
...
+                 {'name': 'Overwrite 3 -> 0',
+                  'status': TestResult.FAIL,
+                  'msg': testcase_msg},
+                 {'name': 'NIST KWP unwrap AES-128 CAVS 21.4 PLAINTEXT LENGTH = 4096 count 0',
+                  'status': TestResult.IGNORE,
+                  'msg': ''})
+
+        expected = []
+        for case in cases:
+            expected.append(TestResult(case['name'], case['status'], get_prepared_msg(case['msg'])))
```
jsarzy, line 215:
```suggestion
         expected = [
            TestResult(
                 'XTEA Decrypt_ecb #1', 
                 TestResult.PASS,
             ),
             TestResult(
                 'Overwrite 3 -> 0',
                 TestResult.FAIL,
                 msg=('\t\t( psa_its_set_wrap( uid, data1->len, data1->x, flags1 ) ) == PSA_SUCCESS
                      '\t\tat line 122, suites/test_suite_psa_its.function\n'
                      '\t\tlhs = 0xffffffffffffff6e = -146\n'
                      '\t\trhs = 0x0000000000000000 = 0\n')
             ),
             TestResult(
                 'NIST KWP unwrap AES-128 CAVS 21.4 PLAINTEXT LENGTH = 4096 count 0',
                  TestResult.IGNORE,
             ),
         ]
```
I suggest using one array with ready `TestResult`. Now we can build `output` just by applying `get_prepared_msg` in an inverse manner or we can just leave `output` as a raw plain text (so `output` is just copy-paste of the embetls test result). I think that the latter option is better. Thanks to that we see how do the raw output of the embedtls test look like. And thanks to that we do not need to use `get_prepare_msg` function

trunner/test/test_harness.py
```diff
@@ -0,0 +1,240 @@
...
+        thread.start()
+        fifo_r = open(self.FIFO, 'r')
+
+        if not assert_wrong:
+            proc = pexpect.fdpexpect.fdspawn(fifo_r, encoding='ascii', timeout=5)
+            result = MbedtlsTestHarness.harness(proc)
+        else:
+            proc = pexpect.fdpexpect.fdspawn(fifo_r, encoding='ascii', timeout=1)
+            testcase.harness = MbedtlsTestHarness.harness
+            testcase.handle(proc)
```
jsarzy, line 83:
```suggestion
        proc = pexpect.fdpexpect.fdspawn(fifo_r, encoding='ascii', timeout=5)
        try:
            result = MbedtlsTestHarness.harness(proc)
        finally:
```
We misunderstood here. We would like to check if `MbetlsTestHarness.harness` throws exception - now we test if `TestCase` handles exception properly.

trunner/test/test_harness.py
```diff
@@ -0,0 +1,240 @@
...
+    def test_wrong(self, fifo, testcase):
+        def writer(lock):
+            output = "Wrong mbedtls test suite output"
+
+            write_output(output, self.FIFO, lock)
+
+        testcase = self.run_harness(writer, assert_wrong=True, testcase=testcase)
+
+        assert "EXCEPTION TIMEOUT" in testcase.exception
+        assert "Wrong mbedtls test suite output" in testcase.exception
```
jsarzy, line 240:
```suggestion
        with pytest.raises(pexpect.TIMEOUT):
            testcase = self.run_harness(writer)
```
That is enough

trunner/testcase.py
jsarzy: I meant `test.FAIL`, sorry

trunner/test/test_harness.py
```diff
@@ -0,0 +1,240 @@
...
+                   + f"{cases[2]['name']} ............................................... PASS\r\n" \
+                   + f"{cases[3]['name']} ...................... PASS\r\n" \
+                   + f"{cases[4]['name']} ..................... PASS\r\n" \
+                   + "PASSED (5 / 5 tests (0 skipped))\r\n"
+
+            write_output(output, self.FIFO, lock)
+
+        result = self.run_harness(writer)
+
+        print(f'type(result) = {type(result)}, type(expected) = {type(expected)}')
```
jsarzy, line 120: please don't leave debug prints in code

trunner/test/test_harness.py
```diff
@@ -0,0 +1,240 @@
+#
```
jsarzy, line 1: I suggest renaming this file to `harness/test_embdtls.py` or `test_harness_embdtls.py`

jsarzy (changes requested):

trunner/test/test_harness.py
damianloew: Yeah, I'd also prefer choosing the second solution

damianloew: I've left only names as variables.

trunner/test/test_harness.py
```diff
@@ -0,0 +1,240 @@
...
+                     + "----------------------------------------------------------------------------\r\n" \
+                     + "PASSED (5 / 5 tests (5 skipped))"
+
+            write_output(output, self.FIFO, lock)
+
+        result = self.run_harness(writer)
+
+        assert_results(result, expected)
+
+    def test_mix(self, fifo):
```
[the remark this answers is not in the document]
damianloew: Thanks,corrections applied in the last force push

jsarzy (approved):

PR #123, opened 2022-07-25
Micropython test suite

PNieck:
## Description

- adding logic for micropython tests
- adding two types of harnesses for micopython tests - one for testing scripts and one for micropython REPL
- enables executing 838 official micropython tests on our system

## Motivation and Context

This change allows to perform micropython tests during executing long tests on ia32-generic-qemu.

We are using official tests form micropython GitHub.

JIRA: CI-134

## Types of changes

## How Has This Been Tested?

`ia32-generic-qemu`

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/52


micropython/test_micropython.c
```diff
@@ -0,0 +1,126 @@
...
+		if (!result)
+			exit_with_error("There is no enough memory to run this test");
+
+		strcpy(result, &line[opt_iden_len]);
+
+		/* Changing the last character from '\n' to ' ' */
+		result[strlen(result) - 1] = ' ';
+	}
+	else
+		result = NULL;
```
niewim19, line 66: We can skip this `else`, if `result` would be declared with NULL value, and then simply returned as it is now

micropython/test_micropython.c
```diff
@@ -0,0 +1,126 @@
...
+			exit_with_error("There is no enough memory to run this test");
+		}
+
+		free(tmp);
+	}
+	else {
+		cmd = cat(MICROPYTHON, argv[1]);
+		if (!cmd)
+			exit_with_error("There is no enough memory to run this test");
+	}
```
niewim19, line 112: In both outcomes of `if` condition, in some way, you are concatenating `MICROPYTHON` and `argv[1]`. Maybe restrict condition just to the `options` part. 

Note: `exit_with_error` usage without proper `free()` of allocated memory.

micropython/test_micropython.c
```diff
@@ -0,0 +1,126 @@
...
+
+	if ((t = system(cmd)) == 1) {
```
niewim19, line 114: We are trying to avoid assignments in conditions. Just for clarity please separate `system()` call and output check. And please use more meaningful name for that output instead of `t` :)

micropython/test_micropython.c
```diff
@@ -0,0 +1,126 @@
...
+		free(cmd);
+		exit_with_error("There was an error during execution micropython test");
+	}
+	else if (t) {
+		free(cmd);
+		exit_with_error("There was an error caused by function system() (not micropython test)");
+	}
```
niewim19, line 121: Maybe here we can drop the `exit_with_error` function and just use `printf()`. No matter if an error occurs we call `free(cmd)`, so that just adds boilerplate code.

trunner/config.py
```diff
@@ -50,7 +50,7 @@ def resolve_phrtos_dir() -> Path:
 QEMU_TARGETS = ['armv7a9-zynq7000-qemu', 'ia32-generic-qemu', 'riscv64-generic-qemu']
 
 # Tests intended to run in long test campaigns
-LONG_TESTS = ['busybox', 'mbedtls']
+LONG_TESTS = ['busybox', 'mbedtls', 'standard_micropython', 'repl_micropython']
```
niewim19, line 53: maybe just `micropython` and `micropython_repl`?

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,221 @@
...
+
+        if re.match(expected_output, test_result):
+            tests_results.append(TestResult(test_path, TestResult.PASS))
+        else:
+            tests_results.append(TestResult(test_path, TestResult.FAIL, error_msg))
+
+        return tests_results
+
+
+class Tools:
```
niewim19, line 182: I think that this construct of `Tools` class works, but should be changed. You use methods of that class in harnesses, but they are not defined as static, and no instance of this object is created. The pythonc way would be to either:
 - make this tools class into methods and store them in `tools.py`, or `tools/upythTools.py` module
 - move this class to the beginning of this file and:
   - make an instance of that object in each harness
   - make those functions static

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,221 @@
...
+        exp_result = exp_file_cat[index+1:]
+
+        return exp_result
+
+    @staticmethod
+    def harness(p):
+        tests_results = []
+
+        try:
+            (test_path, test_result) = MicropythonStandardHarness.get_test_path_and_result(p)
```
niewim19, line 78: You are using methods from the same class, so you do not need to call them by class specificator. You can use:
```suggestion
            (test_path, test_result) = self.get_test_path_and_result(p)
```

That way, if you change the name of that class, you do not need to change all calls :)

**Note:"** applies in more places across both harnesses.

niewim19 (changes requested):
This PR is still a draft, but I left some remarks. Summarizing:
 - **.c part:**
   - please apply coding convention remarks
   - `exit_with_error` approach of error information needs to change as in many places it causes memory leaks in forms like open descriptors or allocated memory not `free`d
 - **.py part:**
   - I just highlighted some coding remarks about usage of methods in classes
   - please think about moving `Tools` methods to other directory

micropython/repl_tests/test.yaml
```diff
@@ -0,0 +1,9 @@
+test:
+    type: repl_micropython
+    targets:
+      include: [ia32-generic-qemu, armv7m7-imxrt106x-evk]
+    tests:
+        - name: cmdline.repl_basic
+          exec: echo usr/test/micropython/cmdline/repl_basic.py
+        - name: cmdline.repl_emacs_keys
+          exec: echo usr/test/micropython/cmdline/repl_emacs_keys.py
```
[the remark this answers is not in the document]
PNieck: OK, I see

micropython/test_micropython.c
```diff
@@ -0,0 +1,126 @@
...
+
+	free(cmd);
+
+	return 0;
+}
```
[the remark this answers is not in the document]
PNieck: I am going to fix this

micropython/test_micropython.c
```diff
@@ -0,0 +1,126 @@
...
+	if (!f)
+		exit_with_error("There was a problem with opening file with a script");
+
+	if (getline(&line, &len, f) < 0)
+		exit_with_error("There was a problem with reading script to execute");
+
+	len = strlen(line) + 1;
+
+	if (len > opt_iden_len &&
+		strncmp(line, CMDLINE_OPTION_IDEN, opt_iden_len) == 0) {
```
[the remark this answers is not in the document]
PNieck: I thought that this this if statement will be too long, but as it turn out it's not bad.

trunner/harnesses/micropython.py
PNieck: I cannot do that, because `harness` is a static method

niewim19: Yeah, i missed that. Sorry. So maybe, to be more pythonic mace those functions as nested functions in harness method? 

I don`t really like the overall idea that harness factory does not return the object, but only static method of some class, but that's whole another topic :)

trunner/harnesses/micropython.py
PNieck: I am going to move this class to the beginning and make those functions static

micropython/test_micropython.c
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	if (access(argv[1], F_OK) != 0) {
+		upyth_errMsg("There is no such a micropython test to run, build project with \"LONG_TEST=y\"");
+		return 1;
+	}
+
+	printf("Running test: %s%s\n", PATH_TO_TESTS, argv[1]);
+
+	tmp = MICROPYTHON;
```
niewim19, line 120: Maybe change `concat` to:
```c
char *upyth_concat(char *str1, const char *str2)
```
and use `realloc` inside to lower memory fragmentation.

micropython/test_micropython.c
```diff
@@ -0,0 +1,157 @@
...
+#define TRUE  1
+#define FALSE 0
+
+static char *PROG_NAME = "";
+
+void upyth_errMsg(char *msg)
+{
+	fprintf(stderr, "Error: %s - %s\n", PROG_NAME, msg);
+}
+
```
niewim19, line 34: Please use double newline between functions

micropython/test_micropython.c
```diff
@@ -0,0 +1,157 @@
...
+			free(line);
+			upyth_errMsg("Malloc error");
+			Error = TRUE;
+			return NULL;
+		}
+
+		strcpy(result, &line[OPTION_IDEN_LEN]);
+
+		/* Changing the last character from '\n' to ' ' */
+		result[strlen(result) - 1] = ' ';
```
niewim19, line 87: If you read only one line you can lose some data

micropython/test_micropython.c
```diff
@@ -0,0 +1,157 @@
...
+int main(int argc, char **argv)
+{
+	char *cmd, *options, *tmp;
+	int upythProgRes, err = FALSE;
+
+	PROG_NAME = argv[0];
+
+	if (argc != 2) {
+		upyth_errMsg("There is no required argument - please execute this function with path to the micropython tests");
+		return 1;
```
niewim19, line 105: You can use `EXIT_FAILURE` for better readibility. Applies elsewhere

niewim19 (changes requested):
Small remarks, as discussed :)

PNieck:
Recent pushes were made in order to adjust test suite to new trunner 
https://github.com/phoenix-rtos/phoenix-rtos-tests/commit/3860a610f7d98d241dc4540924d36fd53fe9c9e2

damianloew (changes requested):
Do all tests (without ignored) pass?

micropython/test_micropython.c
```diff
@@ -0,0 +1,171 @@
...
+char *upyth_concat(char *str1, const char *str2)
+{
+	int len;
+	char *res;
+
+	len = strlen(str1) + strlen(str2) + 1;
+	res = (char *)realloc(str1, sizeof(char) * len);
+
+	if (res != NULL) {
+		res[0] = '\0';
```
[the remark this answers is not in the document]
PNieck: First 'strcat' needs to know where to start overwriting the 'res' string.

micropython/test_micropython.c
```diff
@@ -0,0 +1,171 @@
...
+
+static char *PROG_NAME;
+
+void upyth_errMsg(char *msg)
+{
+	fprintf(stderr, "Error: %s - %s\n", PROG_NAME, msg);
+}
+
+
+char *upyth_concat(char *str1, const char *str2)
```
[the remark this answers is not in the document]
PNieck: 'strcat' do not allocate new memory, you need to have enough space for both strings

micropython/test_micropython.c
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if (access(argv[1], F_OK) != 0) {
+		upyth_errMsg("There is no such a micropython test to run, build project with \"LONG_TEST=y\"");
+		return 1;
+	}
+
+	printf("Running test: %s%s\n", PATH_TO_TESTS, argv[1]);
+
+	cmd = upyth_concat(NULL, MICROPYTHON);
```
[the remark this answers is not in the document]
PNieck: `cmd` need to be allocated dynamically - `realloc` function will be called on this pointer

trunner/config.py
```diff
@@ -50,7 +50,7 @@ def resolve_phrtos_dir() -> Path:
 QEMU_TARGETS = ['armv7a9-zynq7000-qemu', 'ia32-generic-qemu', 'riscv64-generic-qemu']
 
 # Tests intended to run in long test campaigns
-LONG_TESTS = ['busybox', 'mbedtls']
+LONG_TESTS = ['busybox', 'mbedtls', 'micropython', 'micropython_repl']
```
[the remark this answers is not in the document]
PNieck: Originally it was called `standard_micropython`, but @niewim19 suggested to change for `micropython`.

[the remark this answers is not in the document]
PNieck: So maybe `micropython_std` will be an optimal name?

micropython/test_micropython.c
```diff
@@ -0,0 +1,171 @@
...
+#define MICROPYTHON   "/bin/micropython "
+#define PATH_TO_TESTS "/usr/test/micropython/"
+
+#define CMDLINE_OPTION_IDEN "# cmdline: "
+#define OPTION_IDEN_LEN     strlen(CMDLINE_OPTION_IDEN)
+
+#define TRUE  1
+#define FALSE 0
+
+static char *PROG_NAME;
```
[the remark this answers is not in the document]
PNieck: Ok, but in this way I have to pass `argv[0]` to every function which uses `upyth_errMsg(0` - not very elegant in my opinion.
(In this situation it is only `upyth_optionsGet`)

trunner/config.py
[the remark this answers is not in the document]
niewim19: Yea, thats better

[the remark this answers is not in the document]
damianloew: Yes, it would be ok

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,243 @@
...
+        assert index is not None, f"Can't find path to the test\nProgram output:\n{program_output}"
+        index = index.span()
+
+        test_path = program_output[index[0]+1:index[1]]
+
+        # After the path there is a program result
+        index = program_output.find('\n')
+        test_result = program_output[index+1:]
+
+        return (test_path, test_result)
```
jsarzy, line 126: nit: no need for parentheses but of course is a matter of taste :P

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,243 @@
...
+        expected_output_path = test_path + ".exp"
+
+        proc.sendline("cat %s" % expected_output_path)
+        Tools.expect(PROMPT, proc)
+
+        exp_file_cat = Tools.get_before_output(proc)
+
+        # exp_file_cat have a header "cat (path to the expected outpue)"
+        # removing the header
+        index = exp_file_cat.find('\n')
```
jsarzy, line 138: Just send command end expect it afterwards. I mean `proc.expect(rf"cat {expected_output_path}{EOF}")` (it may not work in this form, look at some examples in `psh` directory)

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,243 @@
...
+
+            if (re.match(f'SKIP{EOL}', test_result)):
+                tests_results.append(TestResult(test_path, TestResult.IGNORE,
+                                     "Test uses modules, which are disabled in our implementation"))
+                return tests_results
+
+            exp_result = self.get_exp_output(test_path, p)
+        except AssertionError as e:
+            tests_results.append(TestResult("Unknown micropython test", TestResult.FAIL, str(e)))
+            return tests_results
```
jsarzy, line 157: We do not need to catch assertion, they are handled in `TestCase` class

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,243 @@
...
+        index = exp_file_cat.find('\n')
+        exp_result = exp_file_cat[index+1:]
+
+        return exp_result
+
+    def harness(self, p):
+        tests_results = []
+
+        try:
+            (test_path, test_result) = self.get_test_path_and_result(p)
```
jsarzy, line 147: rename `test_result` to `test_output` or just `output`

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,243 @@
...
+
+            if (re.match(f'SKIP{EOL}', test_result)):
```
jsarzy, line 149: remove parentheses

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,243 @@
...
+                tests_results.append(TestResult(test_path, TestResult.IGNORE,
+                                     "Test uses modules, which are disabled in our implementation"))
+                return tests_results
+
+            exp_result = self.get_exp_output(test_path, p)
```
jsarzy, line 154: rename to `expected_output`

jsarzy (changes requested):
It's nice to have more tests :+1: 
Here is the part of the review. I'm going to finish tomorrow, in meantime you can try to fix few things that I listed

micropython/test_micropython.c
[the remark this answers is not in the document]
damianloew: Yeah, that's right.

micropython/test_micropython.c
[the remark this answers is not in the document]
damianloew: Ok, we could use a bigger buffer for a new string, but maybe `realloc` is a better solution.

micropython/test_micropython.c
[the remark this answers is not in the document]
damianloew: If we decide to use `realloc`, You are right, it can't be used this way.

micropython/test_micropython.c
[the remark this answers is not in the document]
damianloew: Ok, I think You can leave it as it was or avoid using `upyth_errMsg()`

jsarzy (changes requested):
Additional comments

micropython/test_micropython.c
```diff
@@ -0,0 +1,171 @@
...
+
+		strcpy(result, &line[OPTION_IDEN_LEN]);
+
+		/* Removing new line character for result */
+		newLine = strchr(result, '\n');
+		if (newLine != NULL) {
+			*newLine = ' ';
+		}
+
+		/* Changing the last character to ' ' */
```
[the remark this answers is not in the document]
PNieck: I have added a new comment

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,243 @@
...
+# Copyright 2022 Phoenix Systems
+# Authors: Piotr Nieciecki
+#
+
+import re
+import pathlib
+import pexpect
+from .common import TestResult
+
+PROMPT = r'((\r+)\x1b\[0J)?' + r'\n?' + r'\(psh\)% '
```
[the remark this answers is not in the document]
PNieck: It's for harness tests - there prompt is just `\n(psh)% `

micropython/test_micropython.c
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if (getline(&line, &len, f) < 0) {
+		fclose(f);
+		upyth_errMsg("There was a problem with reading script to execute");
+		error = true;
+		return NULL;
+	}
+
+	len = strlen(line) + 1;
```
kemonats, line 74: The `getline()` function returns the number of characters in the buffer or `-1` in the case of the error.
Using `strlen()` in this case is redundant.

micropython/test_micropython.c
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if (access(argv[1], F_OK) != 0) {
+		upyth_errMsg("There is no such a micropython test to run, build project with \"LONG_TEST=y\"");
+		return 1;
+	}
+
+	printf("Running test: %s%s\n", PATH_TO_TESTS, argv[1]);
+
+	cmd = upyth_concat(NULL, MICROPYTHON_BIN);
```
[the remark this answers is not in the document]
kemonats: Maybe
```suggestion
	cmd = strdup(MICROPYTHON_BIN);
```

kemonats, line 130: Maybe
```suggestion
	cmd = strdup(MICROPYTHON_BIN);
```

trunner/harnesses/micropython.py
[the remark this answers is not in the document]
PNieck: I have changed this for standard regex for PROMPT

PNieck:
I have updated this branch after considering all suggestions.
I have left some too long lines, but in my option fixing them won't look good.

trunner/test/harnesses/test_micropython.py
```diff
@@ -0,0 +1,503 @@
...
+        test_path = "/usr/test/micropython/basics/test.py"
+
+        program_name = "/bin/test_haness"
+        error_desc = "Some kind of error"
+        error = f"Error: {program_name} - {error_desc}"
+
+        program_output = f"Running test: {test_path}\r\n{error}"
+
+        with pytest.raises(AssertionError,
+                           match=f"There was an error {error_desc} during program {program_name} execution:\n{program_output}\n"):
```
damianloew, line 305: Just to avoid `line too long` warning.
```suggestion
        msg = f"There was an error {error_desc} during program {program_name} execution:\n{program_output}\n"
        with pytest.raises(AssertionError, match=msg):
```

micropython/standard_tests/test.yaml
```diff
@@ -0,0 +1,1845 @@
...
+        - name: basics.string_slice
+          exec: test_micropython basics/string_slice.py
+        - name: basics.builtin_divmod_intbig
+          exec: test_micropython basics/builtin_divmod_intbig.py
+        - name: basics.boundmeth1
+          exec: test_micropython basics/boundmeth1.py
+        - name: basics.iter2
+          exec: test_micropython basics/iter2.py
+
+          # Tests with names like "viper_ ..." or "native_ ..." are ignored in MicroPython tests
```
damianloew, line 999: Why are they ignored? Please add the explanation

damianloew (changes requested):
We prefer to solve flake8 warnings if possible.

damianloew (commented):
I've run all tests and have 4 fails (ia32-generic-qemu, building with micropython and busybox ports). Please investigate it and solve issues/skip them (with appropriate description).
Screenshot from 2022-08-18 14-48-00
Screenshot from 2022-08-18 14-48-14
a36ba5547.png)
Screenshot from 2022-08-18 14-48-38

PNieck:
@damianloew Have you used mine script form `phoenix-rtos-ports`? https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/52
Where I am giving more memory for MicroPython

damianloew:
> @damianloew Have you used mine script form `phoenix-rtos-ports`? [phoenix-rtos/phoenix-rtos-ports#52](https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/52) Where I am giving more memory for MicroPython

I see now that I've built micropython without your changes in ports, then have done checkout and build (without clean). I think that was a reason, now they pass :)

micropython/standard_tests/test.yaml
```diff
@@ -0,0 +1,1732 @@
...
+          exec: test_micropython micropython/heap_lock.py
+        - name: micropython.viper_binop_comp
+          exec: test_micropython micropython/viper_binop_comp.py
+        - name: micropython.viper_ptr32_store
+          exec: test_micropython micropython/viper_ptr32_store.py
+        - name: micropython.native_try
+          exec: test_micropython micropython/native_try.py
+        - name: micropython.viper_binop_multi_comp
+          exec: test_micropython micropython/viper_binop_multi_comp.py
+          ignore: True
```
damianloew, line 1058: No comment about the reason of skipping

damianloew (approved):
Please add the comment, apart from that LGTM

trunner/test/harnesses/test_micropython.py
```diff
@@ -0,0 +1,503 @@
...
+        program_output = args[1]
+        exp_output = args[2]
+
+        send_through_socket(socket, program_output)
+        # b = program_output.encode("utf-8")
+        # socket.send(b)
+
+        if exp_output is not None:
+            recv_exp_through_socket(socket, f"cat {test_path}.exp\n")
+            send_through_socket(socket, exp_output)
```
jsarzy, line 260:
```suggestion
    def harness_callback(mock, *args):
        program_output = args[0]
        mock.send(program_output)

        if len(args) > 1:
            test_path = args[1]
            exp_output = args[2]
            mock.recv_exp(f"cat {test_path}.exp\n")
            mock.send(exp_output)
```
suggestion how it can looks with "mock" object that I mentioned before

trunner/test/harnesses/test_micropython.py
```diff
@@ -0,0 +1,503 @@
...
+        for cmd in program_commands:
+            element = cmd
+
+            if not cmd == program_commands[-1]:
+                element += '\r\n'
+
+            commands += element
+
+        commands += PROMPT
+        commands = "".join(commands)
```
jsarzy, line 415:
```suggestion
        commands = "\r\n".join(program_commands)
        commands += PROMPT
        # or
        commands = f"{'\r\n'.join(program_commands)}{PROMPT}"
```
There is no need to build the entire string from array

trunner/test/harnesses/test_micropython.py
```diff
@@ -0,0 +1,503 @@
...
+    proc = pexpect.fdpexpect.fdspawn(conn.fileno(), encoding="utf-8")
+    result = harness(proc)
+
+    conn.close()
+    program.join()
+
+    try:
+        os.remove(SOCKET_FILE)
+    except OSError:
+        pass
```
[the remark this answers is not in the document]
PNieck: I agree, this is not the best behavior

micropython/test_micropython.c
```diff
@@ -0,0 +1,180 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+
+#define MICROPYTHON_BIN "/bin/micropython "
+#define PATH_TO_TESTS   "/usr/test/micropython/"
+
+#define CMDLINE_PREFIX     "# cmdline: "
+#define CMDLINE_PREFIX_LEN strlen(CMDLINE_PREFIX)
```
jsarzy, line 23:
```suggestion
#define CMDLINE_PREFIX_LEN (sizeof(CMDLINE_PREFIX) - 1)
```
As we know the CMDLINE_PREFIX at the compile time there is no need to call strlen

trunner/harnesses/micropython.py
```diff
@@ -0,0 +1,183 @@
...
+            return test_output
+
+        exp_output = self.get_exp_output(test_path, p)
+
+        test_passed = False
+        if self.is_test_with_regex_output(test_path):
+            expected_output_re = create_regex(exp_output)
+            if re.match(expected_output_re, test_result):
+                test_passed = True
+
```
jsarzy, line 124: remove this new line

trunner/test/harnesses/test_micropython.py
```diff
@@ -0,0 +1,473 @@
...
+
+        s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        if os.path.exists(self.SOCKET_FILE):
+            os.remove(self.SOCKET_FILE)
+        s.bind(self.SOCKET_FILE)
+
+        lock = threading.Lock()
+        lock.acquire()
+
+        thread = threading.Thread(target=self.__start_callback, args=(lock, *callback_args))
```
jsarzy, line 76: pass `callback` directly to the __start_callback, not by setting it as an attribute

trunner/test/harnesses/test_micropython.py
```diff
@@ -0,0 +1,473 @@
...
+    if exp_result.msg is not None:
+        assert result.msg == exp_result.msg
+
+
+class MicropythonMock:
+
+    SOCKET_FILE = "/tmp/test_micropython"
+
+    def __init__(self):
+        self.success = True
```
jsarzy, line 51:
```suggestion
        self.socket = None
        self.success = True
        self.error_msg  = None
```

trunner/test/harnesses/test_micropython.py
```diff
@@ -0,0 +1,473 @@
...
+
+    def send(self, text):
+        self.socket.send(text.encode("ascii"))
+
+    def recv_exp(self, exp_output):
+        res = self.socket.recv(len(exp_output))
+        res = res.decode("ascii")
+        if res != exp_output:
+            self.success = False
+            self.problem_msg = f"Error - callback didn't received \"{exp_output}\", but got \"{res}\""
```
jsarzy, line 61: Maybe raise an exception here and catch it in `__start_callback`?

jsarzy, line 61: I think simple `error_msg` would fit better

trunner/test/harnesses/test_micropython.py
```diff
@@ -0,0 +1,473 @@
...
+        s.close()
+
+        proc = pexpect.fdpexpect.fdspawn(conn.fileno(), encoding="ascii")
+
+        try:
+            result = harness(proc)
+        finally:
+            conn.close()
+            thread.join()
+
```
jsarzy, line 91: nit: no need for new line

trunner/test/harnesses/test_micropython.py
```diff
@@ -0,0 +1,473 @@
...
+    def test_standard_regex_escape(self):
+
+        def get_outputs(inputs):
+            outputs = []
+
+            for input in inputs:
+                outputs.append(standard_regex_escape(input))
+
+            return outputs
```
jsarzy, line 125:
```suggestion
        get_outputs = lambda inputs: [standard_regex_escape(input) for input in inputs]
```
Maybe use list comprehension here? I think both versions are clear and the latter is shorter.

jsarzy (changes requested):
I left a few comments, most of them nitpicking

I don't like that we call exec echo name_of_test.py in repl_tests but I know this is in part due to limitations in the runner. I will keep in mind to fix this :)

trunner/test/harnesses/test_micropython.py
PNieck: Why I have to catch it? Let the pytest catch it

jsarzy: Child thread exception is not propagated to the main thread, so pytest can't catch this exception (at least I think it can't, you may check it)

trunner/test/harnesses/test_micropython.py
PNieck: Unfortunately our review dog don't like this idea

PNieck: But I propose this:
```python
def get_outputs(inputs):
       return [standard_regex_escape(input) for input in inputs]
```

jsarzy (dismissed):
lgtm

trunner/test/harnesses/test_micropython.py
PNieck: > Child thread exception is not propagated to the main thread, so pytest can't catch this exception (at least I think it can't, you may check it)

Pytest return a warning, that thread gives had an uncought exception, but test passes

damianloew (approved):
My last force push was only resolving small conflict after my recently merged change. I've run all tests and everything seems to work fine.

niewim19 (approved):
LGTM

PR #126, opened 2022-08-24
Add posix tests (waiting for lsb_vsx port merge)

adamdebek:
## Description

Add program to execute posix test suite
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/lsb_vsx/lsb_vsx.py
+++ b/lsb_vsx/lsb_vsx.py
@@ -1,0 +1,56 @@
+from typing import Optional
+
+from trunner.dut import Dut
+from trunner.ctx import TestContext
+from trunner.types import Status, TestResult
+
+
+def harness(dut: Dut, ctx: TestContext, result: TestResult) -> Optional[TestResult]:
+    tc_start = r"(.+?\|){2}TP Start\r+\n"
+    tc_end = r"(.+?\|){2}IC End\r+\n"
+    results = r"(.+?\|){2}(?P<status>PASS|FAIL|UNRESOLVED|UNSUPPORTED|NOTINUSE|UNTESTED|UNINITIATED|NORESULT|INVALID RESULT)\r+\n"  # noqa: E501
+    final = r"(.+?\|){2}TC End.+?\r+\n"
+    msg_line = r"(.+?\|){2}(?P<msg_line>.+?)\r+\n"
+
+    stats = {"FAIL": 0, "IGNORE": 0, "PASS": 0, "UNTESTED": 0}
+    get_msg, msg = False, ""
+    tc_num = 0
+
+    dut.expect("Config End\r+\n")
+
+    while True:
+        idx = dut.expect([tc_start, results, tc_end, final, msg_line], timeout=2000)
+        parsed = dut.match.groupdict()
+
+        if idx == 0:
+            get_msg = True
+
+        elif idx == 1:
+            if parsed["status"] in ("UNSUPPORTED", "NOTINUSE"):
+                parsed["status"] = "IGNORE"
+            elif parsed["status"] in ("UNRESOLVED", "UNINITIATED", "NORESULT"):
+                # TODO: research pseudo-languages
+                if msg and "pseudo language" not in msg:
+                    parsed["status"] = "FAIL"
+                else:
+                    parsed["status"] = "IGNORE"
+            elif parsed["status"] in ("INVALID RESULT"):
+                parsed["status"] = "FAIL"
+
+            status = Status.from_str(parsed["status"])
+            stats[parsed["status"]] += 1
+
+        elif idx == 2:
+            tc_num += 1
+            subname = f"testcase {tc_num}"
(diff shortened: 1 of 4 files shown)


mateusz-bloch (commented):
I just took a look at the Python code, not a full review.

lsb_vsx/lsb_vsx.py
```diff
@@ -0,0 +1,52 @@
...
+        parsed = dut.match.groupdict()
+
+        if idx == 0:
+            get_msg = True
+        elif idx == 1:
+            break
+        elif idx == 2:
+            # Right now, not supported statuses treat as PASS
+            if parsed["status"] in ("UNRESOLVED", "UNSUPPORTED", "NOTINUSE", "UNTESTED"):
+                parsed["status"] = "PASS"
```
mateusz-bloch, line 34: Shouldn't this be more like SKIP? By the way, we have catching UNTESTED .
image

lsb_vsx/lsb_vsx.py
```diff
@@ -0,0 +1,52 @@
...
+
+            status = Status.from_str(parsed["status"])
+            tc_num += 1
+            subname = f"testcase {tc_num}"
+            if msg_lines:
+                msg += msg_lines[0]
+                for line in msg_lines[1:]:
+                    msg += "\n" + " " * 26 + line.lstrip()
```
mateusz-bloch, line 42: It should work, but it’s up to you
```suggestion
                    msg += "\n\t\t" + line.lstrip()
```

adamdebek: I think trunner should be able to format msg with new lines.

lsb_vsx/run-test.c
```diff
@@ -0,0 +1,68 @@
...
+
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <stdlib.h>
+
+
+int main(int argc, char **argv)
+{
+	const char *cmd = "/usr/bin/tcc -j - -e -l ";
```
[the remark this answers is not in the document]
adamdebek: Hyphen as option argument is widely used to indicate standard output or input. In this case it mean journal is writen to standard output, which is needed by harness.

Comments are used to describe why you do something not to describe what it does, every one can read manual about tcc, links are in README.

I think there is no such option to silence this output.

lsb_vsx/test.yaml
```diff
@@ -0,0 +1,546 @@
...
+        - name: T.isxdigit
+          execute: run-test /tset/ANSI.os/charhandle/isxdigit/T.isxdigit
+        - name: T.tolower
+          execute: run-test /tset/ANSI.os/charhandle/tolower/T.tolower
+        - name: T.toupper
+          execute: run-test /tset/ANSI.os/charhandle/toupper/T.toupper
+        - name: T.assert
+          execute: run-test /tset/ANSI.os/diagnostics/assert/T.assert
+        - name: T.abort
+          ignore: True
```
damianloew, line 37: why is it skipped? Lack of link to the issue. There are a few more cases like this.

lsb_vsx/lsb_vsx.py
```diff
@@ -0,0 +1,56 @@
...
+
+from trunner.dut import Dut
+from trunner.ctx import TestContext
+from trunner.types import Status, TestResult
+
+
+def harness(dut: Dut, ctx: TestContext, result: TestResult) -> Optional[TestResult]:
+    tc_start = r"(.+?\|){2}TP Start\r+\n"
+    tc_end = r"(.+?\|){2}IC End\r+\n"
+    results = r"(.+?\|){2}(?P<status>PASS|FAIL|UNRESOLVED|UNSUPPORTED|NOTINUSE|UNTESTED|UNINITIATED|NORESULT|INVALID RESULT)\r+\n"  # noqa: E501
```
damianloew, line 11:
```
# noqa: E501
```

What is this comment about?

damianloew (changes requested):

lsb_vsx/lsb_vsx.py
adamdebek: This line is longer than 120 chars and this comment supress this rule for that line

lsb_vsx/test.yaml
adamdebek: It is skipped because this test failed to build, because lack of some system functions definition

damianloew: Then please leave it in a comment, like:
```
# skipped because the test wasn't built as the implementation of fun1, fun2, fun3 is missing
```

adamdebek: Then I would need to put 100 comments

damianloew: so please leave it at the first case like this and note that it's applicable for the rest of skips without comment.

damianloew (approved):

PR #128, opened 2022-08-25
psh: add history test

damianloew:
## Description

- psh: add history test

## Motivation and Context

- JIRA: PD-77

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/psh/test-history.py
+++ b/psh/test-history.py
@@ -1,0 +1,113 @@
+import string
+
+import psh.tools.psh as psh
+from psh.tools.common import CHARS, get_rand_strings
+
+
+def assert_sent(p, cmds, msg='', repeat=False):
+    i = 2
+    expected = ['   1  history']
+    for cmd in cmds:
+        for _ in range(3 if repeat else 1):
+            p.sendline(cmd)
+            psh.assert_prompt(p, msg=f'No prompt seen after passing the following command: {cmd}')
+
+        line = f"{i:>4}  {cmd}"
+        expected.append(line)
+        i += 1
+    expected.append(f'  {i}  history')
+    psh.assert_cmd(p, 'history', expected, msg)
+
+
+def assert_help(p):
+    assert_msg = "Prompt hasn't been displayed properly, when calling with -h argument"
+    psh.assert_prompt_after_cmd(p, 'history -h', assert_msg)
+    psh.assert_cmd_successed(p)
+
+    assert_msg = "Prompt hasn't been displayed properly, when calling with wrong arguments"
+    psh.assert_prompt_after_cmd(p, 'history -z', assert_msg)
+    psh.assert_cmd_failed(p)
+    # Temporary disabled, because of the following issue:
+    # https://github.com/phoenix-rtos/phoenix-rtos-project/issues/521
+    # psh.assert_prompt_after_cmd(p, 'history -xx', assert_msg)
+    # psh.assert_cmd_failed(p)
+    psh.assert_prompt_after_cmd(p, 'history @!$ qrw $*%', assert_msg)
+    psh.assert_cmd_failed(p)
+    # the order in CHARS list can be random
+    # temporarily because of #521 issue we use sorted version
+    psh.assert_prompt_after_cmd(p, 'history ' + ''.join(sorted(CHARS)), assert_msg)
+    psh.assert_cmd_failed(p)
+
+
+def assert_clear(p):
+    psh.assert_cmd(p, 'history -c', msg='Unexpected output after history -c command')
+    psh.assert_cmd(p, 'history', '  1  history', msg='Wrong history output after clear')
+
+
+def assert_rand_cmds(p, psh_cmds):
+    cmds = get_rand_strings(CHARS, 50)
+
+    cmds = list(set(cmds) - set(psh_cmds))
+    msg = 'Wrong history output after running random unknown commands'
+    assert_sent(p, cmds, msg)
+
+
+def assert_psh_cmds(p, psh_cmds, with_repeats=False):
+    # perf, pm disabled because of issues, double history is printed as one
+    omitted = {'edit', 'exit', 'perf', 'pm', 'top', 'reboot', 'history'}
+
+    remaining = list(set(psh_cmds) - omitted)
+
+    if with_repeats:
+        msg = 'Wrong history output after running available psh commands multiple time'
+        assert_sent(p, remaining, msg, repeat=True)
+    else:
+        msg = 'Wrong history output after running available psh commands'
+        assert_sent(p, remaining, msg)
+
+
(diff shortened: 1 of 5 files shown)


psh/test-history.py
```diff
@@ -0,0 +1,106 @@
...
+    expected.append(f'  {i}  history')
+    psh.assert_cmd(p, 'history', expected, msg)
+
+
+def assert_help(p):
+    assert_msg = "Prompt hasn't been displayed properly, when calling with -h argument"
+    psh.assert_prompt_after_cmd(p, 'history -h', assert_msg)
+
+    assert_msg = "Prompt hasn't been displayed properly, when calling with wrong arguments"
+    psh.assert_prompt_after_cmd(p, 'history -z', assert_msg)
```
jsarzy, line 28: I checked and `history -z` prints usage output, but `history --z` prints nothing. Do you know why?

psh/test-history.py
```diff
@@ -0,0 +1,106 @@
...
+def assert_multicall_random(p):
+    cmds = get_rand_names(CHARS, 20)
+    msg = 'Wrong history output after multiple time running same commands'
+    assert_sent(p, cmds, msg, multi=True)
+
+
+def assert_empty(p):
+    assert_clear(p)
+    msg = 'Wrong history output after sending multiple whitespaces'
+    p.sendline('')
```
jsarzy, line 79:
```suggestion
    p.sendline()
```

psh/tools/common.py
```diff
@@ -20,6 +23,14 @@
 # acceptable separators: white spaces (wss) + CC, CC + wss, wss
 SEPARATOR_PATTERN = r'(?:' + CONTROL_CODE + r'|\s)+'
 
+CHARS = list(set(string.printable) - set(string.whitespace) - set('/'))
+
+
+def get_rand_names(pool, count, min_chars=8, max_chars=16):
+    ''' Returns random names (with length between min_chars and max_chars) from characters pool'''
+    names = [''.join(random.choices(pool, k=random.randint(min_chars, max_chars))) for _ in range(count)]
+    return names
```
jsarzy, line 32: just return it, there is no need to assing it to variable

psh/tools/common.py
```diff
@@ -20,6 +23,14 @@
 # acceptable separators: white spaces (wss) + CC, CC + wss, wss
 SEPARATOR_PATTERN = r'(?:' + CONTROL_CODE + r'|\s)+'
 
+CHARS = list(set(string.printable) - set(string.whitespace) - set('/'))
+
+
+def get_rand_names(pool, count, min_chars=8, max_chars=16):
```
jsarzy, line 29: maybe get_rand_strings

psh/test-history.py
```diff
@@ -0,0 +1,106 @@
...
+
+def assert_rand_cmds(p, psh_cmds):
+    cmds = get_rand_names(CHARS, 50)
+
+    cmds = list(set(cmds) - set(psh_cmds))
+    msg = 'Wrong history output after running random unknown commands'
+    assert_sent(p, cmds, msg)
+
+
+def assert_psh_cmds(p, psh_cmds, double=False):
```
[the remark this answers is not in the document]
damianloew: Yeah, it would be better

psh/test-history.py
```diff
@@ -0,0 +1,106 @@
...
+            line = f'   {i}  ' + cmd
+        expected.append(line)
+        i = i + 1
+    expected.append(f'  {i}  history')
+    psh.assert_cmd(p, 'history', expected, msg)
+
+
+def assert_help(p):
+    assert_msg = "Prompt hasn't been displayed properly, when calling with -h argument"
+    psh.assert_prompt_after_cmd(p, 'history -h', assert_msg)
```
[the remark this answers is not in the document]
damianloew: Yes, that's right. I will add checking return code in next commit.

psh/tools/common.py
```diff
@@ -20,6 +23,14 @@
...
+    ''' Returns random names (with length between min_chars and max_chars) from characters pool'''
+    names = [''.join(random.choices(pool, k=random.randint(min_chars, max_chars))) for _ in range(count)]
+    return names
+
```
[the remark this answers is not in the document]
damianloew: Yeah, I will restructure test-mkdir when adding test-touch, which would be very similar

psh/test-history.py
damianloew: It's the matter of current implementation. getopt() returns `-1`, when `--` is pased and optind is incremented, so nothing happens:
https://github.com/phoenix-rtos/phoenix-rtos-utils/blob/94a09ff0f89b2feda06ac9a9719ea4c949fc71e3/psh/pshapp/pshapp.c#L930-L952

jsarzy (approved):

psh/tools/common.py
```diff
@@ -20,6 +23,14 @@
 # acceptable separators: white spaces (wss) + CC, CC + wss, wss
 SEPARATOR_PATTERN = r'(?:' + CONTROL_CODE + r'|\s)+'
 
+# When subtracting sets the order is mixed, we want to avoid this
+CHARS = sorted(list(set(string.printable) - set(string.whitespace) - set('/')))
```
jsarzy, line 27: Why do we care about the order?
I think `CHARS` might be too general for this set of chars (I would expect that chars are from range 0-255). Maybe we do not need to define it here, just put it directly in the tests

damianloew: I am not sure whether randomness is a good thing here. There were random cases (1~50), which revealed the https://github.com/phoenix-rtos/phoenix-rtos-project/issues/521 issue, but I think it's better to have separate test case for checking that.

About name: I don't have an idea for some better common name here.

It will be used in more tests, so I think it's better to place it here.

jsarzy: I still don't understand why we need to sort the `CHARS` variable. As I look at the tests, we use `CHARS` in two cases:
1. In the line 36 as an argument to the `history` call
2. As an argument for `get_rand_strings`

In the second case, the order of the pool doesn't matter. Since we are making the random selection of elements from the pool we don't care about the order (we treat the pool as a box with a hole from which we pull out a ball with some letter written on it and then put the ball back in the box).
In the first case if the order is important we should just pass `sorted(CHARS)` there and leave a comment why we are doing this.

If random testing reveals any problems then we're happy and leave it with the hope that this test case might help us find some other weird bugs in the future. (But maybe I did not understand your reply correctly. If there is something I don't see, please correct me)

About the variable name: OK, you can leave it as is. I also do not have the good name for this variable. I made the name `CHARS` in the `mkdir` test because it was defined locally in the module.

jsarzy (approved):

PR #129, opened 2022-08-26
Add libcache tests

malwro:
## Description

Add tests of libcache library

## Motivation and Context

JIRA: RTOS-204

## Types of changes

## How Has This Been Tested?

## Checklist:

- [x] I have updated the documentation accordingly.

## Special treatment

https://github.com/phoenix-rtos/phoenix-rtos-corelibs/pull/21

### DIFF
--- a/libcache/libcache_utils.c
+++ b/libcache/libcache_utils.c
@@ -36,20 +36,24 @@ int test_genCharFile(void)
 	if (access("/var/libcache_test_char.txt", F_OK) != 0) {
 		file = fopen("/var/libcache_test_char.txt", "w");
 		if (!file) {
-			perror("Unable to open /var/libcache_test_char.txt\n");
+			fprintf(stderr, "Unable to open /var/libcache_test_char.txt\n");
 			ret = -1;
 		}
 		else {
-			fseek(file, 0, SEEK_SET);
-
-			for (i = 0; i < LIBCACHE_SRC_MEM_SIZE; ++i) {
-				num = rand();
-				fwrite_unlocked(&num, sizeof(char), 1, file);
+			if (fseek(file, 0, SEEK_SET) != 0) {
+				fprintf(stderr, "fseek() fail in %s\n", __FILE__);
+				fclose(file);
+				ret = -1;
 			}
+			else {
+				for (i = 0; i < LIBCACHE_SRC_MEM_SIZE; ++i) {
+					num = rand();
+					fwrite_unlocked(&num, sizeof(char), 1, file);
+				}
+				fclose(file);
 
-			fclose(file);
-
-			ret = EOK;
+				ret = EOK;
+			}
 		}
 	}
 
@@ -69,20 +73,25 @@ int test_genIntFile(void)
 	if (access("/var/libcache_test_int.txt", F_OK) != 0) {
 		file = fopen("/var/libcache_test_int.txt", "w");
 		if (!file) {
-			perror("Unable to open /var/libcache_test_char.txt\n");
+			fprintf(stderr, "Unable to open /var/libcache_test_char.txt\n");
 			ret = -1;
 		}
 		else {
-			fseek(file, 0, SEEK_SET);
-
-			for (i = 0; i < LIBCACHE_SRC_MEM_SIZE / 4; ++i) {
-				num = rand();
-				fwrite_unlocked(&num, sizeof(int), 1, file);
+			if (fseek(file, 0, SEEK_SET) != 0) {
+				fprintf(stderr, "fseek() fail in %s\n", __FILE__);
+				fclose(file);
+				ret = -1;
 			}
+			else {
+				for (i = 0; i < LIBCACHE_SRC_MEM_SIZE / 4; ++i) {
+					num = rand();
+					fwrite_unlocked(&num, sizeof(int), 1, file);
+				}
 
-			fclose(file);
+				fclose(file);
 
-			ret = EOK;
+				ret = EOK;
+			}
 		}
 	}
 
--- a/libcache/libcache_utils.c
+++ b/libcache/libcache_utils.c
@@ -89,22 +89,22 @@ ssize_t test_writeCbErr(uint64_t offset, const void *buffer, size_t count, cache
 
 void *test_cache_write(void *args)
 {
-	ssize_t written = -1;
+	ssize_t *ret;
 	test_write_args_t *arguments = args;
 
-	written = cache_write(arguments->cache, arguments->addr, arguments->buffer, arguments->count, arguments->policy);
-	TEST_ASSERT_EQUAL_INT(arguments->count, written);
+	ret = malloc(sizeof(ssize_t));
+	*ret = cache_write(arguments->cache, arguments->addr, arguments->buffer, arguments->count, arguments->policy);
 
-	return NULL;
+	return (void *)ret;
 }
 
 void *test_cache_read(void *args)
 {
-	ssize_t read = -1;
+	ssize_t *ret;
 	test_read_args_t *arguments = args;
 
-	read = cache_read(arguments->cache, arguments->addr, arguments->buffer, arguments->count);
-	TEST_ASSERT_EQUAL_INT(arguments->count, read);
+	ret = malloc(sizeof(ssize_t));
+	*ret = cache_read(arguments->cache, arguments->addr, arguments->buffer, arguments->count);
 
-	return NULL;
+	return (void *)ret;
 }
--- a/libcache/libcache_utils.c
+++ b/libcache/libcache_utils.c
@@ -130,22 +130,18 @@ ssize_t test_writeCbErr(uint64_t offset, const void *buffer, size_t count, cache
 
 void *test_cache_write(void *args)
 {
-	ssize_t *ret;
 	test_write_args_t *arguments = args;
 
-	ret = malloc(sizeof(ssize_t));
-	*ret = cache_write(arguments->cache, arguments->addr, arguments->buffer, arguments->count, arguments->policy);
+	arguments->actualCount = cache_write(arguments->cache, arguments->addr, arguments->buffer, arguments->count, arguments->policy);
 
-	return (void *)ret;
+	return NULL;
 }
 
 void *test_cache_read(void *args)
 {
-	ssize_t *ret;
 	test_read_args_t *arguments = args;
 
-	ret = malloc(sizeof(ssize_t));
-	*ret = cache_read(arguments->cache, arguments->addr, arguments->buffer, arguments->count);
+	arguments->actualCount = cache_read(arguments->cache, arguments->addr, arguments->buffer, arguments->count);
 
-	return (void *)ret;
+	return NULL;
 }
--- a/libcache/libcache_utils.c
+++ b/libcache/libcache_utils.c
@@ -23,38 +23,70 @@
 #include <unistd.h>
 
 
-void test_genCharFile(void)
+int test_genCharFile(void)
 {
+	int i, ret = -1;
 	FILE *file;
-	int i;
 	char num;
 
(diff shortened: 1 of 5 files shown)


damianloew (changes requested):
Good to have tests that cover the implementation of a new library :)

Do You also meet this issue? Please resolve it:

Screenshot from 2022-09-05 12-20-51

libcache/test_libcache.c
```diff
@@ -0,0 +1,989 @@
...
+			false, false, false -> false (no short-circuit)
+ */
+
+/* true, true, true -> true (short-circuit) */
+TEST(test_read_write, cache_write_uninitialized_cache)
+{
+	ssize_t written = 0;
+	cachectx_t *cache = NULL;
+	size_t count = 10 * sizeof(char);
+	uint64_t addr = 1143552ULL;
```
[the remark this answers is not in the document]
malwro: This is a random address with no particular meaning. It reoccurs in a few places due to the fact that similar test bodies are partial copies of each other. If necessary, I can change them to avoid confusion in the future.

Please advise.

libcache/test_libcache.c
```diff
@@ -0,0 +1,989 @@
...
+{
+	int ret = -1;
+
+	ret = close(fileRead);
+	TEST_ASSERT_EQUAL_INT(0, ret);
+
+	ret = close(fileWrite);
+	TEST_ASSERT_EQUAL(0, ret);
+
+	ret = remove("/var/write.txt");
```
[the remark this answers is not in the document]
malwro: In the tests, `/var/write.txt` and `/var/read.txt` simulate a source memory device that the `libcache` library can write to and read from by the use of suitable callbacks. Here the write and read operations are 'separated' and they are performed on two different files unlike in real settings, where they happen on the same source memory. This approach has helped me to empirically evaluate the outcome of the operations, especially the write operation.

I think I may need to 'merge' the files to simulate the source memory more accurately. This would also reduce the overhead caused by reoccurring creation of additional files.

libcache/libcache_utils.h
```diff
@@ -0,0 +1,46 @@
+#ifndef _LIBCACHE_UTILS_H
+#define _LIBCACHE_UTILS_H
+
+#include "cache.h"
+
+
+#define LIBCACHE_SRC_MEM_SIZE 1417216ULL
```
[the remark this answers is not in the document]
malwro: The `cache_init` functions takes as an argument the maximum capacity of a cached source memory (`size_t srcMemSize`). In the tests two text files make up for the physical memory, creating an abstraction that has to respect some real conditions anyway. Hence `LIBCACHE_SRC_MEM_SIZE` is only an arbitrary value needed to create the abstraction of the physical memory in the tests.

libcache/test_libcache.c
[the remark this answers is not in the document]
malwro: This above is no longer an issue as I've merged the files into one.

damianloew (changes requested):
I am considering storing memory addresses in hex to make it more readable. Above that there is some problem with executing the test two times in a row, please resolve it.

jsarzy:
Please note that Unity asserts should only be used in the main thread context. It's a macro for longjmp therefore when the test fails some weird things can happen. Threads could communicate by shared state and the main thread eventually can verify it using assertions. (similar to https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/121)

libcache/libcache_utils.c
```diff
@@ -0,0 +1,110 @@
...
+	return -EIO;
+}
+
+void *test_cache_write(void *args)
+{
+	ssize_t written = -1;
+	test_write_args_t *arguments = args;
+
+	written = cache_write(arguments->cache, arguments->addr, arguments->buffer, arguments->count, arguments->policy);
+	TEST_ASSERT_EQUAL_INT(arguments->count, written);
```
jsarzy, line 96: Please note that Unity asserts should only be used in the main thread context. It's a macro for longjmp therefore when the test fails some weird things can happen. Threads could communicate by shared state and the main thread eventually can verify it using assertions. (similar to https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/121)

libcache/test_libcache.c
```diff
@@ -0,0 +1,1148 @@
...
+	uint64_t addr = 1156339ULL, offset = addr & offMask;
+	size_t count = 127 * sizeof(char);
+	size_t remainder = 0, flushed = 0;
+	char *buffer, *actual, *expected;
+
+	buffer = "^#$%^$#%^&$#&^*$(^*&^)_)_(++(_)_(*)(&^%^*%^$#%$@#$@!# @!$#$#%$ $#%##$^$#%^#$$!@!*!!~~~!@#@$$_#@_+$ 4#$%#$%#%#$%^^#$^$#^#$^%@#$$";
+
+	remainder = (count - (LIBCACHE_LINE_SIZE - offset)) % LIBCACHE_LINE_SIZE;
+	flushed = offset + count + (LIBCACHE_LINE_SIZE - remainder);
+	actual = calloc(flushed, sizeof(char));
```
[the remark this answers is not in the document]
malwro: No longer needed, thank you. Changed to `malloc`.

libcache/libcache_utils.h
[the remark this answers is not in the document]
malwro: I halved the value and added a short comment. Thank you.

libcache/libcache_utils.c
malwro: Thank you for pointing that out. This assertion is important from the point of view of the test though, as I need to test the number of bytes written by `cache_write` function.

I've figured out a workaround: I capture the value returned from the thread function by passing a pointer to a buffer to obtain that value. Unfortunately this approach causes an overhead of a few additional `malloc` and `free` calls.

A possible solution without `malloc` would be adding an additional field to `test_write_args_t` to capture the number of written bytes so that I could make an assertion `TEST_ASSERT_EQUAL_INT(args->written, ...) in the main thread context before proceeding.

I think that the first solution is better as it simply makes use of the `**__thread_return` argument of the `int pthread_join(pthread_t __th, void **__thread_return)` function.

libcache/test_libcache.c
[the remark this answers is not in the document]
malwro: I've grouped addresses with the same values and created suitable defines. Thank you.

libcache/test_libcache.c
```diff
@@ -0,0 +1,1148 @@
...
+TEST(test_init, cache_init_sizeModLineSizeNotZero)
+{
+	size_t size = 1754, lineSize = 17; /* size % lineSize = 3, (int)(size / lineSize) % LIBCACHE_NUM_WAYS = 3 */
+	cachectx_t *cache = NULL;
+
+	cache = cache_init(LIBCACHE_SRC_MEM_SIZE, size, lineSize, &ops);
+	TEST_ASSERT_NULL(cache);
+}
+
+TEST(test_init, cache_init_sizeModLineSizeZero)
```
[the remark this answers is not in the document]
malwro: Thank you, the new name reflects the purpose of the test much better.

agkaminski (commented):
General remark regarding libcache_utils.c: return values checks are missing, from what I see in test it's ok, as it seem everything is checked using unity.

libcache/libcache_utils.c
```diff
@@ -0,0 +1,151 @@
...
+{
+	return -EIO;
+}
+
+void *test_cache_write(void *args)
+{
+	ssize_t *ret;
+	test_write_args_t *arguments = args;
+
+	ret = malloc(sizeof(ssize_t));
```
agkaminski, line 136: Alternative solution: put this into test_write_args_t struct.

libcache/libcache_utils.c
```diff
@@ -0,0 +1,151 @@
...
+
+	return (void *)ret;
+}
+
+void *test_cache_read(void *args)
+{
+	ssize_t *ret;
+	test_read_args_t *arguments = args;
+
+	ret = malloc(sizeof(ssize_t));
```
agkaminski, line 147: Alternative solution: put this into test_read_args_t struct.

libcache/test_libcache.c
```diff
@@ -0,0 +1,1349 @@
...
+	int ret = -1;
+	ssize_t readCount;
+	cachectx_t *cache;
+	size_t count = 16 * sizeof(char);
+	char *buffer;
+
+	cache = cache_init(LIBCACHE_SRC_MEM_SIZE, LIBCACHE_LINE_SIZE, LIBCACHE_LINES_CNT, &ops);
+	TEST_ASSERT_NOT_NULL(cache);
+
+	buffer = malloc(count);
```
agkaminski, line 325: No malloc retval check

libcache/test_libcache.c
```diff
@@ -0,0 +1,1349 @@
...
+
+	cache = cache_init(LIBCACHE_SRC_MEM_SIZE, LIBCACHE_LINE_SIZE, LIBCACHE_LINES_CNT, &ops);
+	TEST_ASSERT_NOT_NULL(cache);
+
+	bufferW = "FSDGSGDGSDGDSGDF";
+
+	writeCount = cache_write(cache, LIBCACHE_SRC_MEM_SIZE - 10, bufferW, count, LIBCACHE_WRITE_BACK);
+	TEST_ASSERT_EQUAL_INT(10 * sizeof(char), writeCount);
+
+	bufferR = malloc(count);
```
agkaminski, line 351: No malloc retval check

libcache/test_libcache.c
```diff
@@ -0,0 +1,1349 @@
...
+	test_read_args_t args1, args2;
+	uint64_t addr1 = LIBCACHE_ADDR_OFF_27, addr2 = 0x23dbULL;
+	cachectx_t *cache;
+	size_t count1 = 164 * sizeof(char);
+	size_t count2 = 10 * sizeof(char);
+
+	buffer1 = malloc(count1 * sizeof(char));
+	buffer2 = malloc(count2 * sizeof(char));
+	buffer3 = malloc(count1 * sizeof(char));
+	buffer4 = malloc(count2 * sizeof(char));
```
agkaminski, line 628: No malloc retval check

libcache/test_libcache.c
```diff
@@ -0,0 +1,1349 @@
...
+	cache = cache_init(LIBCACHE_SRC_MEM_SIZE, LIBCACHE_LINE_SIZE, LIBCACHE_LINES_CNT, &ops);
+	TEST_ASSERT_NOT_NULL(cache);
+
+	writeCount = cache_write(cache, begAddr, buffer1, 64, LIBCACHE_WRITE_BACK);
+	TEST_ASSERT_EQUAL_INT(count, writeCount);
+
+	ret = cache_flush(cache, LIBCACHE_SRC_MEM_SIZE - 10, LIBCACHE_SRC_MEM_SIZE + 30);
+	TEST_ASSERT_EQUAL_INT(0, ret);
+
+	buffer2 = malloc(count);
```
agkaminski, line 749: No malloc retval check

libcache/test_libcache.c
```diff
@@ -0,0 +1,1349 @@
...
+	cachectx_t *cache = NULL;
+	uint64_t addr = LIBCACHE_ADDR_DUMMY;
+	char *bufferR;
+	size_t count = 14 * sizeof(char);
+
+	ops.readCb = test_readCbErr;
+	cache = cache_init(LIBCACHE_SRC_MEM_SIZE, LIBCACHE_LINE_SIZE, LIBCACHE_LINES_CNT, &ops);
+	TEST_ASSERT_NOT_NULL(cache);
+
+	bufferR = malloc(count);
```
agkaminski, line 1159: No malloc retval check

agkaminski (approved):
LGTM

PR #135, opened 2022-09-30
strtod test

DawidSzpejna:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work: https://github.com/phoenix-rtos/libphoenix/pull/193


nalajcie:
Did not review, I've just run Your tests on host (with glibc) and one of the tests fails - please verify if it's correct:
image

Also - please change order of params to `TEST_ASSERT_*` macros:
image

So the error messages would make more sense :)

libc/strtod.c
```diff
@@ -0,0 +1,189 @@
...
+TEST(strtod_family, errno_test)
+{
+	errno = 0;
+	const char num1[] =
+		"17976931348623157081452742373170435679807056752584499659891747680315"
+		"72607800285387605895586327668781715404589535143824642343213268894641"
+		"82768467546703537516986049910576551282076245490090389328944075868508"
+		"45513394230458323690322294816580855933212334827479782620414472316873"
+		"8177180919299881250404026184124858368123111";
+	strtod(num1, NULL);
```
agkaminski, line 154: Hmm, shouldn't we also assert that `strtod` retval is `NAN`?

nalajcie:
Please update with upstream now that we have new `strtod` implementation which should pass the tests

libc/strtod.c
DawidSzpejna: It is very reasonable to check retval, i will add it. However, i think retval should be equal to `\0` character.

libc/strtod.c
```diff
@@ -0,0 +1,198 @@
...
+		"27976931348623157081452742373170435679807056752584499659891747680315"
+		"72607800285387605895586327668781715404589535143824642343213268894641"
+		"82768467546703537516986049910576551282076245490090389328944075868508"
+		"45513394230458323690322294816580855933212334827479782620414472316873"
+		"8177180919299881250404026184124858368";
+	TEST_ASSERT_DOUBLE_IS_INF(strtod(num3, &retval));
+	TEST_ASSERT_EQUAL_STRING("", retval);
+}
+
+TEST(strtod_family, errno_test)
```
niewim19, line 154: Why splitting into `errno_test` and `too_long_numbers`? The lateral should also cause errno to be set, we can check it there without separate test group. There is no scenario where conversion fails and errno is not set.

DawidSzpejna: I decided to split into `errno_test` and `too_long_numbers` because `strtod` may fail with `EINVAL` if no conversion could be performed.

niewim19 (commented):
Not all edge cases are covered, but this would be a good task for our test-team @damianloew  

For me it is good to go, please update this branch to current version. I am adding @damianloew as reviewer.

damianloew:
> Not all edge cases are covered, but this would be a good task for our test-team @damianloew
> 
> For me it is good to go, please update this branch to current version. I am adding @damianloew as reviewer.

Yeah, we will upgrade this test to provide a better coverage, but let's merge the current version for now (of course after resolving conflicts).

PR #139, opened 2022-10-06
libc: add new test cases to the getopt test 

damianloew:
## Description

- libc: add new test cases to the getopt

## Motivation and Context

 - JIRA: PD-305

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (host-pc).

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work ((https://github.com/phoenix-rtos/libphoenix/pull/208)).

### DIFF
--- a/libc/unistd_getopt.c
+++ b/libc/unistd_getopt.c
@@ -14,6 +14,7 @@
  */
 
 #include <unistd.h>
+#include <string.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
@@ -39,11 +40,8 @@ struct getopt_t {
 };
 
 
-int testargc = 0;
-struct getopt_t ret;
-
 /* updates `results` structure according to getopt() operations */
-int testmain(int argc, char *argv[], struct getopt_t *results, const char *optstring)
+static int testmain(int argc, char *const argv[], struct getopt_t *results, const char *optstring)
 {
 	int c;
 	int arg;
@@ -85,15 +83,8 @@ TEST_GROUP(unistd_getopt);
 
 TEST_SETUP(unistd_getopt)
 {
-	/* default presets of results structure */
-	ret.aflag = 0;
-	ret.bflag = 0;
-	ret.cvalue = NULL;
-	ret.err = NOERR;
-	ret.nonopts = 0;
-
 	/* reset of getopt() index value optind */
-	optind = 1;
+	optind = 0;
 }
 
 
@@ -104,8 +95,9 @@ TEST_TEAR_DOWN(unistd_getopt)
 TEST(unistd_getopt, getopt_zeroargs)
 {
 	/* mocks of main() arguments */
-	char *testargv[] = { "cmd" };
-	testargc = 1;
+	char *const testargv[] = { "cmd", NULL };
+	const int testargc = sizeof(testargv) / sizeof(testargv[0]) - 1;
+	struct getopt_t ret = { 0 };
 
 	testmain(testargc, testargv, &ret, "abc:");
 
@@ -120,8 +112,9 @@ TEST(unistd_getopt, getopt_zeroargs)
 TEST(unistd_getopt, getopt_normal_flags)
 {
 	/* mocks of main() arguments */
-	char *testargv[] = { "cmd", "-a", "-b" };
-	testargc = 3;
+	char *testargv[] = { "cmd", "-a", "-b", NULL };
+	const int testargc = sizeof(testargv) / sizeof(testargv[0]) - 1;
+	struct getopt_t ret = { 0 };
 
 	testmain(testargc, testargv, &ret, "abc:");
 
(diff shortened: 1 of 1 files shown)


libc/unistd_getopt.c
```diff
@@ -314,6 +314,33 @@ TEST(unistd_getopt, getopt_unknownopt_noarg)
...
+	TEST_ASSERT_NULL(ret.cvalue);
+	TEST_ASSERT_EQUAL_INT(0, ret.nonopts);
+
+	/* Assert that parsing argument work correctly after passing multiple unknown options */
+	testargv[1] = "-ab";
+	testargv[2] = "-c";
+	testargv[3] = "value";
+	testargc = 4;
+
+	testmain(testargc, testargv, &ret, "abc:");
```
nalajcie, line 336: invalid - calling `getopt` again without resetting it, also without resetting `ret` struct.

Also: make `ret` struct static

libc/unistd_getopt.c
```diff
@@ -314,6 +314,33 @@ TEST(unistd_getopt, getopt_unknownopt_noarg)
 }
 
 
+TEST(unistd_getopt, getopt_unknownopt_multiple)
+{
+	/* mocks of main() arguments */
+	char *testargv[] = { "cmd", "-xxx", "", "" };
```
nalajcie, line 320: use two separate `const char *` arrays if needed for test

libc/unistd_getopt.c
```diff
@@ -314,6 +314,33 @@ TEST(unistd_getopt, getopt_unknownopt_noarg)
...
+	testargc = 2;
```
nalajcie, line 321: make it a local variable

nalajcie (changes requested):
For sure something is wrong with the test - it fails also on host (with glibc `getopt` implementation):
image

libc/unistd_getopt.c
damianloew: I would leave it without `const` here, tried to change, but it becomes complicated then (getopt requires char * const argv[])

damianloew:
> For sure something is wrong with the test - it fails also on host (with glibc `getopt` implementation): image

Changes applied and some new cases added.

libc/unistd_getopt.c
```diff
@@ -314,6 +313,101 @@ TEST(unistd_getopt, getopt_unknownopt_noarg)
 }
 
 
+TEST(unistd_getopt, getopt_unknownopt_multiple)
+{
+	/* mocks of main() arguments */
+	char *testargv1[] = { "cmd", "-xxx", "", "" };
```
nalajcie, line 319: no need for empty strings as `testargc = 2`

libc/unistd_getopt.c
```diff
@@ -39,8 +39,7 @@ struct getopt_t {
 };
 
 
-int testargc = 0;
-struct getopt_t ret;
+static struct getopt_t ret;
```
nalajcie, line 42: thanks for adding `static`, IMHO `ret` could be a local variable in tests (small enough to fit comfortably on stack on every platform), initialized to `0`, eg:
`struct getopt_t ret = { 0 };`

`testmain` could also be static, and `argv` `const` below

libc/unistd_getopt.c
```diff
@@ -314,6 +313,101 @@ TEST(unistd_getopt, getopt_unknownopt_noarg)
...
+	TEST_ASSERT_EQUAL_INT(UNKNOWNOPT, ret.err);
+	TEST_ASSERT_EQUAL_INT(0, ret.aflag);
+	TEST_ASSERT_EQUAL_INT(0, ret.bflag);
+	TEST_ASSERT_NULL(ret.cvalue);
+	TEST_ASSERT_EQUAL_INT(0, ret.nonopts);
+
+	/* Assert that parsing argument works correctly after passing multiple unknown options
+	   There was issue related to this case */
+
+	/* default presets of results structure */
```
nalajcie, line 333: either provide function for resetting `struct getopt_t` type or use `memset`.

Please note that this doesn't need to be a single test (if the getopt resetting is not working - some of the tests executed after this one would fail

nalajcie (changes requested):
pointed out some small things which could be improved further. Did not run the test on host.

libc/unistd_getopt.c
damianloew: Yeah, I agree on that, but I would rather leave it to show by case name the reason of fail

libc/unistd_getopt.c
damianloew: > pointed out some small things which could be improved further. Did not run the test on host.

Thanks, applied.

libc/unistd_getopt.c
```diff
@@ -314,6 +318,107 @@ TEST(unistd_getopt, getopt_unknownopt_noarg)
...
+	ret.aflag = 0;
```
nalajcie, line 340: no need to reset fields individually, You're doing `memset()` 2 lines below anyway

libc/unistd_getopt.c
```diff
@@ -314,6 +318,107 @@ TEST(unistd_getopt, getopt_unknownopt_noarg)
...
+	ret.bflag = 0;
+	ret.cvalue = NULL;
+	ret.err = NOERR;
+	ret.nonopts = 0;
+
+	/* reset getopt() index value optind and ret structure */
+	optind = 0;
+	memset(&ret, 0, sizeof(ret));
```
nalajcie, line 348: fix compilation warning

nalajcie (changes requested):
two final things to fix and we can merge this :)

damianloew:
> two final things to fix and we can merge this :)

Thanks, resolved.

nalajcie (approved):

PR #149, opened 2022-11-22
psh: test-touch fixes, enabling psh rootfs-tests on zynq7000-zedboard

damianloew:
## Description

 - psh: touch: move rootfs cases to separate test
 - psh: enable psh rootfs tests on zynq7000 target
 - psh: fix naming in test-touch

## Motivation and Context

 - The case with hardlinks was named improperly (they were called symlinks)
 - The hardlinks testcase was executed on non-rootfs targets
 - Because of using `value` instead of `exclude` rootfs targets were not executed on `zynq7000-zedboard` target

## Types of changes 

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/psh/test-touch-rootfs.py
+++ b/psh/test-touch-rootfs.py
@@ -12,19 +12,19 @@
 # %LICENSE%
 
 import psh.tools.psh as psh
-from psh.tools.common import assert_timestamp_change
+from psh.tools.common import assert_mtime
 
 
 def assert_hardlinks(p):
-    uptimes = {}
+    dates = {}
     # all psh commands are hardlinks
-    psh_cmds = psh.get_commands(p)
-    for psh_cmd in psh_cmds:
-        uptimes[psh_cmd] = psh.uptime(p)
-        msg = f"Prompt hasn't been seen after the hardlink touch: /bin/{psh_cmd}"
-        psh.assert_cmd(p, f'touch /bin/{psh_cmd}', '', msg)
+    for hardlink in psh.get_commands(p):
+        hardlink_path = f'/bin/{hardlink}'
+        dates[hardlink] = psh.date(p)
+        msg = f"Prompt hasn't been seen after the hardlink touch: {hardlink_path}"
+        psh.assert_cmd(p, f'touch {hardlink_path}', msg=msg)
 
-    assert_timestamp_change(p, psh_cmds, uptimes, '/bin')
+    assert_mtime(p, dates, '/bin')
 
 
 @psh.run
--- a/psh/test-touch-rootfs.py
+++ b/psh/test-touch-rootfs.py
@@ -1,0 +1,32 @@
+# Phoenix-RTOS
+#
+# phoenix-rtos-tests
+#
+# psh touch command test for rootfs targets
+#
+# Copyright 2022 Phoenix Systems
+# Author: Damian Loewnau
+#
+# This file is part of Phoenix-RTOS.
+#
+# %LICENSE%
+
+import psh.tools.psh as psh
+from psh.tools.common import assert_timestamp_change
+
+
+def assert_hardlinks(p):
+    uptimes = {}
+    # all psh commands are hardlinks
+    psh_cmds = psh.get_commands(p)
+    for psh_cmd in psh_cmds:
+        uptimes[psh_cmd] = psh.uptime(p)
+        msg = f"Prompt hasn't been seen after the hardlink touch: /bin/{psh_cmd}"
+        psh.assert_cmd(p, f'touch /bin/{psh_cmd}', '', msg)
+
+    assert_timestamp_change(p, psh_cmds, uptimes, '/bin')
+
+
+@psh.run
+def harness(p):
+    assert_hardlinks(p)
--- a/psh/test-touch.py
+++ b/psh/test-touch.py
@@ -2,7 +2,7 @@
 #
 # phoenix-rtos-tests
 #
-# psh touch command text
+# psh touch command test
 #
 # Copyright 2022 Phoenix Systems
 # Author: Damian Loewnau
@@ -15,34 +15,16 @@
 import psh.tools.psh as psh
 import trunner.config as config
 from psh.tools.common import (CHARS, assert_present, assert_file_created, assert_dir_created,
-                              assert_random_files, get_rand_strings, create_testdir)
+                              assert_random_files, get_rand_strings, create_testdir, assert_timestamp_change)
 
 ROOT_TEST_DIR = 'test_touch_dir'
 
 
-def assert_timestamp_change(p, touched_files, uptimes, dir=''):
-    files = psh.ls(p, dir)
-    for file in files:
-        if file.name in touched_files:
-            uptime = uptimes[file.name]
-            # uptime has to be got before touch
-            # to prevent failed assertion when typing uptime in 00:59 and touch in 01:00
-            # 1 min margin is applied
-            next_minute = f'{(int(uptime.minute)+1):02d}'
-            if file.date.time == f'{uptime.hour}:{next_minute}':
-                uptime_str = f'{uptime.hour}:{next_minute}'
-            else:
-                uptime_str = f'{uptime.hour}:{uptime.minute}'
-            # when writing test, system date is always set to Jan 01 00:00
-            # so all touched files should have the following timestamp: Jan 01 00:00 + uptime
-            msg = f"The file's timestamp hasn't been changed after touch: {dir}/{file.name}, timestamp = {file.date}"
-            assert file.date == ('Jan', '1', uptime_str), msg
-
-
-def assert_symlinks(p):
+def assert_hardlinks(p):
     uptimes = {}
-    # all psh commands are symlinks
+    # all psh commands are hardlinks
     psh_cmds = psh.get_commands(p)
+    psh_cmds.remove('psh')
     for psh_cmd in psh_cmds:
         uptimes[psh_cmd] = psh.uptime(p)
         msg = f"Prompt hasn't been seen after the symlink touch: /bin/{psh_cmd}"
@@ -130,6 +112,6 @@ def harness(p):
     assert_created_dir(p)
     assert_existing_dirs(p)
 
-    assert_symlinks(p)
+    assert_hardlinks(p)
     assert_devices(p)
     assert_executable(p)
--- a/psh/test-touch.py
+++ b/psh/test-touch.py
@@ -20,19 +20,6 @@ from psh.tools.common import (CHARS, assert_present, assert_file_created, assert
 ROOT_TEST_DIR = 'test_touch_dir'
 
 
-def assert_hardlinks(p):
-    uptimes = {}
-    # all psh commands are hardlinks
-    psh_cmds = psh.get_commands(p)
-    psh_cmds.remove('psh')
-    for psh_cmd in psh_cmds:
-        uptimes[psh_cmd] = psh.uptime(p)
-        msg = f"Prompt hasn't been seen after the symlink touch: /bin/{psh_cmd}"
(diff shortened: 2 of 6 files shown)


psh/test-touch-rootfs.py
```diff
@@ -0,0 +1,32 @@
...
+import psh.tools.psh as psh
+from psh.tools.common import assert_timestamp_change
+
+
+def assert_hardlinks(p):
+    uptimes = {}
+    # all psh commands are hardlinks
+    psh_cmds = psh.get_commands(p)
+    for psh_cmd in psh_cmds:
+        uptimes[psh_cmd] = psh.uptime(p)
```
nalajcie, line 23: IMHO invalid, the times would be set to system time - not uptime (so if somebody/something would change the system time - eg. `date` test run previously) this test would fail even though the functionality might work correctly.

Please at least name it `dates` not to confuse it with uptime value in the future (applies elsewhere also)

psh/test-touch-rootfs.py
```diff
@@ -0,0 +1,32 @@
...
+
+import psh.tools.psh as psh
+from psh.tools.common import assert_timestamp_change
+
+
+def assert_hardlinks(p):
+    uptimes = {}
+    # all psh commands are hardlinks
+    psh_cmds = psh.get_commands(p)
+    for psh_cmd in psh_cmds:
```
nalajcie, line 22: You're talking about files (hardlinks) here, not about commands - please change the variable names accordingly

psh/tools/common.py
```diff
@@ -38,6 +38,25 @@ def create_testdir(p, dirname):
     psh.assert_cmd(p, f'mkdir {dirname}', '', msg)
 
 
+def assert_timestamp_change(p, touched_files, uptimes, dir=''):
```
nalajcie, line 41: if this is a **common** routine, remove all references to the `touch` test (eg. variable names)

psh/tools/common.py
```diff
@@ -38,6 +38,25 @@ def create_testdir(p, dirname):
...
+    files = psh.ls(p, dir)
+    for file in files:
+        if file.name in touched_files:
+            uptime = uptimes[file.name]
```
nalajcie, line 45: not checking if all files from `touched_files` are present in `ls`. IMHO the loop should be over `touched_files` not the other way around

psh/tools/common.py
```diff
@@ -38,6 +38,25 @@ def create_testdir(p, dirname):
...
+            # uptime has to be got before touch
+            # to prevent failed assertion when typing uptime in 00:59 and touch in 01:00
```
nalajcie, line 47: uh, it would probably better to convert the date to unix timestamp or python datetime object (inside `ls` to be reused between the tests?) and just compare the unix timestamps with some margin (instead of operations on strings).

Please also note that You're only verifying the modification time (access time change is not verified by this test, for that we would need to `stat` the file).

psh/tools/common.py
```diff
@@ -38,6 +38,25 @@ def create_testdir(p, dirname):
...
+            # 1 min margin is applied
+            next_minute = f'{(int(uptime.minute)+1):02d}'
+            if file.date.time == f'{uptime.hour}:{next_minute}':
+                uptime_str = f'{uptime.hour}:{next_minute}'
+            else:
+                uptime_str = f'{uptime.hour}:{uptime.minute}'
+            # when writing test, system date is always set to Jan 01 00:00
```
nalajcie, line 54: not true, I didn't write tests in 1970 :P.

Still, it will not be true if we will run tests without DUT reboot and the `date` or `ntp` test would run beforehand. IMHO it would be better to remove this assumption as we're working on this test now.

psh/test-touch-rootfs.py
```diff
@@ -0,0 +1,32 @@
...
+# %LICENSE%
+
+import psh.tools.psh as psh
+from psh.tools.common import assert_timestamp_change
+
+
+def assert_hardlinks(p):
+    uptimes = {}
+    # all psh commands are hardlinks
+    psh_cmds = psh.get_commands(p)
```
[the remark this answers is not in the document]
damianloew: I need base names below, so there is my proposal:

```python
def assert_hardlinks(p):
    datetimes = {}
    # all psh commands are hardlinks
    psh_hardlinks = psh.get_commands(p)
    for hardlink in psh_hardlinks:
        hardlink_path = f'/bin/{hardlink}'
        datetimes[hardlink] = psh.date(p)
        msg = f"Prompt hasn't been seen after the hardlink touch: {hardlink_path}"
        psh.assert_cmd(p, f'touch {hardlink_path}', '', msg)

    assert_modiftime_change(p, psh_hardlinks, datetimes, '/bin')
```

damianloew (commented):
@nalajcie thanks for the suggestions. I've taken up all of them, please check the latest commits

psh/tools/psh.py
```diff
@@ -182,6 +184,22 @@ def uptime(pexpect_proc):
...
+    pexpect_proc.expect(rf'(?P<year>\d+):(?P<month>\d+):(?P<day>\d+):(?P<hour>\d+):(?P<min>\d+):(?P<sec>\d+){EOL}')
+
+    m = pexpect_proc.match
+    return datetime(
+        int(m.group('year')),
+        int(m.group('month')),
+        int(m.group('day')),
+        int(m.group('hour')),
+        int(m.group('min')),
+        int(m.group('sec')))
```
jsarzy, line 200:
```suggestion
    return datetime(*map(int, m.groups()))
```

psh/test-touch.py
```diff
@@ -15,63 +15,32 @@
...
-    assert_timestamp_change(p, psh_cmds, uptimes, '/bin')
-
-
 def assert_executable(p):
     # psh is the example of executable which exists in file system
     msg = "Prompt hasn't been seen after the executable touch: /usr/bin/hello"
-    uptime = psh.uptime(p)
+    date = psh.date(p)
     if config.CURRENT_TARGET in config.SYSEXEC_TARGETS:
         psh.assert_cmd(p, 'touch /syspage/psh', '', msg)
```
jsarzy, line 28:
```suggestion
        psh.assert_cmd(p, 'touch /syspage/psh', msg=msg)
```
Don't pass default arguments, use keywords.

psh/test-touch-rootfs.py
```diff
@@ -0,0 +1,33 @@
...
+
+import psh.tools.psh as psh
+from psh.tools.common import assert_mtime_change
+
+
+def assert_hardlinks(p):
+    dates = {}
+    # all psh commands are hardlinks
+    psh_hardlinks = psh.get_commands(p)
+    for hardlink in psh_hardlinks:
```
jsarzy, line 22:
```suggestion
    for hardlink in psh.get_commands(p):
```
Maybe like this? We don't use hardlinks else where and it nicely says us that psh command is a hardlink.

psh/test-touch-rootfs.py
```diff
@@ -0,0 +1,33 @@
...
+        hardlink_path = f'/bin/{hardlink}'
+        dates[hardlink] = psh.date(p)
+        msg = f"Prompt hasn't been seen after the hardlink touch: {hardlink_path}"
+        psh.assert_cmd(p, f'touch {hardlink_path}', '', msg)
```
jsarzy, line 26: If we need basename then I would not define new variable and keep just "/bin" prefix, but can be as is :).
Also don't need to pass empty string, just use `msg=msg`

jsarzy (changes requested):

psh/tools/common.py
```diff
@@ -38,6 +41,30 @@ def create_testdir(p, dirname):
...
+
+        # to prevent failed assertion when typing date in 00:59 and touch in 01:00
+        if file_datetime.minute != target_datetime.minute and abs(file_datetime.minute - target_datetime.minute) == 1:
+            file_datetime = file_datetime.replace(minute=target_datetime.minute)
+        # we do not want to compare years and seconds (not printed by ls -l)
+        file_datetime = file_datetime.replace(year=target_datetime.year, second=target_datetime.second)
+
+        assert file_datetime == target_datetime, "".join(
+            f'The modification time for {fname} is not equal to the target one!',
+            f'file datetime: {file_datetime} target datetime: {target_datetime}')
```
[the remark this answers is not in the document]
damianloew: I got such a suggestion one time:
Screenshot from 2022-11-29 10-20-47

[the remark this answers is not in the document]
jsarzy: Okey, so leave as is :+1:

psh/tools/common.py
```diff
@@ -38,6 +41,30 @@ def create_testdir(p, dirname):
...
+    existing_fnames = [file.name for file in existing_files]
+    assert set(files_to_check).issubset(set(existing_fnames)), 'Files to check are not present in the passed directory!'
+
+    for fname in files_to_check:
+        target_datetime = datetimes[fname]
+        existing_file_id = existing_fnames.index(fname)
+        file_datetime = existing_files[existing_file_id].datetime
+
+        # to prevent failed assertion when typing date in 00:59 and touch in 01:00
+        if file_datetime.minute != target_datetime.minute and abs(file_datetime.minute - target_datetime.minute) == 1:
```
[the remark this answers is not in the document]
damianloew: To cover 59:59 and 00:00 case, the `date.minute` is then lower than `target_datetime.minute`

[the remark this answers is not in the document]
jsarzy: Yeap you are right, but the date.minute is int type, so the difference will be 59. You might use `-` operator on datetime type and  then check particular fields.

psh/test-touch-rootfs.py
damianloew: I would rather leave it then.

psh/tools/common.py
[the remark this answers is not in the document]
damianloew: Yeah, I've finally done it this way:

```python
        # to prevent failed assertion when typing date in 00:59 and touch in 01:00
        if date - target_datetime == timedelta(minutes=1):
            date = target_datetime
```

psh/tools/common.py
```diff
@@ -38,6 +41,30 @@ def create_testdir(p, dirname):
...
+
+    dir_files = {file.name: file for file in psh.ls(p, dir)}
+
+    for filename, target_datetime in datetimes.items():
+        assert filename in dir_files, f'File {filename} is not present in the passed directory!'
+        date = dir_files[filename].datetime
+
+        # we do not want to compare years and seconds (not printed by ls -l)
+        date = date.replace(year=target_datetime.year)
+        target_datetime = target_datetime.replace(second=0)
```
jsarzy, line 57: Maybe we just can set `date.second` to `target_datetime.second` like this:
```suggestion
        date = date.replace(year=target_datetime.year, second=target_datetime.second)
```
The ls function sets seconds to zero anyways so we don't need the second line

psh/tools/common.py
```diff
@@ -38,6 +41,30 @@ def create_testdir(p, dirname):
...
 
+def assert_mtime(p, datetimes: Dict[str, datetime], dir=''):
+    ''' Asserts that files (keys in datetimes dictionary) have modification time
+    equal to corresponding datetime values with 1 min margin.
+    Year and seconds are not checked. All files shall be present in the `dir` directory. '''
+
+    dir_files = {file.name: file for file in psh.ls(p, dir)}
+
+    for filename, target_datetime in datetimes.items():
+        assert filename in dir_files, f'File {filename} is not present in the passed directory!'
```
jsarzy, line 52: Maybe instead of "passed" just use dir variable

jsarzy (commented):
Minor improvements and will be ready to merge

damianloew:
> Minor improvements and will be ready to merge

Thanks! Two suggested changes applied and rebased in the last fp.

jsarzy (approved):

PR #153, opened 2022-11-30
libc: make getpw.c host compatible

damianloew:
## Description

 - libc: make getpw.c host compatible
 - libc: make unistd_uids.c host/POSIX compatible 

## Motivation and Context

 - JIRA: PD-321
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic-qemu phoenix, host).

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/353).

### DIFF
--- a/libc/getpw.c
+++ b/libc/getpw.c
@@ -22,6 +22,15 @@
 #include <unity_fixture.h>
 
 
+/* "/" is a root dir and the supported shell is sh on Phoenix-RTOS */
+#ifdef __phoenix__
+#define ROOT_WORKDIR "/"
+#define ROOT_SHELL   "/bin/sh"
+#else
+#define ROOT_WORKDIR "/root"
+#define ROOT_SHELL   "/bin/bash"
+#endif
+
 static unsigned int ispasswdfile;
 
 
@@ -52,8 +61,8 @@ TEST(getpwd, getpwnam_getroot)
 		TEST_ASSERT_EQUAL_INT(0, pw->pw_uid);
 		TEST_ASSERT_EQUAL_INT(0, pw->pw_gid);
 		TEST_ASSERT_EQUAL_STRING("root", pw->pw_gecos);
-		TEST_ASSERT_EQUAL_STRING("/", pw->pw_dir);
-		TEST_ASSERT_EQUAL_STRING("/bin/sh", pw->pw_shell);
+		TEST_ASSERT_EQUAL_STRING(ROOT_WORKDIR, pw->pw_dir);
+		TEST_ASSERT_EQUAL_STRING(ROOT_SHELL, pw->pw_shell);
 	}
 	else {
 		TEST_ASSERT_NULL(pw);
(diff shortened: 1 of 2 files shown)


libc/getpw.c
```diff
@@ -52,8 +52,14 @@ TEST(getpwd, getpwnam_getroot)
...
 		TEST_ASSERT_EQUAL_INT(0, pw->pw_gid);
 		TEST_ASSERT_EQUAL_STRING("root", pw->pw_gecos);
+		/* "/" is a root dir and the supported shell is sh on Phoenix-RTOS */
+#ifdef __phoenix__
 		TEST_ASSERT_EQUAL_STRING("/", pw->pw_dir);
 		TEST_ASSERT_EQUAL_STRING("/bin/sh", pw->pw_shell);
+#else
+		TEST_ASSERT_EQUAL_STRING("/root", pw->pw_dir);
+		TEST_ASSERT_EQUAL_STRING("/bin/bash", pw->pw_shell);
+#endif
```
niewim19, line 62: How about using two macros for these  strings which are selected based on proposed `ifdef`? Less `ifdef`s in the test body

libc/getpw.c
```diff
@@ -147,5 +159,9 @@ TEST_GROUP_RUNNER(getpwd)
 	RUN_TEST_CASE(getpwd, getpwuid_getroot);
 	RUN_TEST_CASE(getpwd, getpwnam_getnull);
 	RUN_TEST_CASE(getpwd, getpwuid_getnull);
+	/* we can't rename/delete the /etc/passwd file on host */
+#ifndef __linux__
 	RUN_TEST_CASE(getpwd, getpwnam_nopasswdfile);
+	RUN_TEST_CASE(getpwd, getpwuid_nopasswdfile);
+#endif
```
niewim19, line 166: imho if we ammend test only for our platform lets do it only for it, so instead og `#ifndef __linux__` lets use `#ifdef __phoenix__`

libc/unistd_uids.c
```diff
@@ -110,7 +112,7 @@ TEST(unistd_uids, setpuids_setsid)
 	TEST_ASSERT_GREATER_THAN_INT(0, parent[1]);  /* nonzero pgid */
 	TEST_ASSERT_GREATER_THAN_INT(0, parent[2]);  /* nonzero sid */
 	TEST_ASSERT_EQUAL_INT(parent[0], parent[1]); /* pid == pgid */
-	TEST_ASSERT_EQUAL_INT(parent[1], parent[2]); /* pgid == sid */
+	/* we don't know the initial session leader, so we don't assert the exact value */
```
niewim19, line 115: This comment is on place regarding this commit, but is out of context in this file. Better use something like:
```suggestion
	/* we don't check parent pgid against sid as we don't know who was parents session leader */
```

niewim19 (changes requested):

libc/unistd_uids.c
```diff
@@ -90,13 +90,15 @@ TEST(unistd_uids, setpuids_setsid)
...
+		sidret = setsid();
 
 		child[3] = getpid();
 		child[4] = getpgrp();
 		child[5] = getsid(getpid());
+		/* sid return value == sid */
+		TEST_ASSERT_EQUAL_INT(sidret, child[5]);
 
-		_exit(siderr != 0);
+		_exit(0);
```
niewim19, line 101: Throwing unity errors not from main thread causes strange behaviors as the test may continue from the main thread.

damianloew: Yeah, in case of fail it would be handled incorrectly. Switched to volatile local variables.

niewim19 (approved):

PR #154, opened 2022-12-06
make pthread tests host/POSIX compatible and add test-libc to host tests

damianloew:
## Description

 - libc: fix getline_allocated test case
 - add test-libc to host tests
 - libc: make pthread tests host/POSIX compatible

## Motivation and Context

JIRA: PD-326
JIRA: PD-322

POSIX compliant code

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-build/pull/136, https://github.com/phoenix-rtos/libphoenix/pull/225).

### DIFF
--- a/libc/string.c
+++ b/libc/string.c
@@ -27,6 +27,8 @@
 #define STR_PLACEHOLDER "klmnopqrstu"
 
 
+/* NOTE: strlcpy not available on linux, enable tests only on phoenix */
+#ifdef __phoenix__
 TEST_GROUP(string_strlcpy);
 
 
@@ -40,8 +42,6 @@ TEST_TEAR_DOWN(string_strlcpy)
 }
 
 
-/* NOTE: strlcpy not available on linux, enable tests only on phoenix */
-#ifdef __phoenix__
 TEST(string_strlcpy, strlcpy_fullcopy)
 {
 	const char source[] = STR_SRC;
@@ -118,6 +118,7 @@ TEST_GROUP_RUNNER(string_strlcpy)
 }
 
 
+#ifdef __phoenix__
 TEST_GROUP(string_strlcat);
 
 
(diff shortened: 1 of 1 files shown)


libc/pthread_cond_test_functions.c
```diff
@@ -70,6 +72,9 @@ void *decrement_count_timed_wait_fail_too_short_timeout(void *args)
 	thread_err_t *thread_err = (thread_err_t *)args;
 	struct timespec time;
 	clock_gettime(CLOCK_REALTIME, &time);
+	/* to set past time and avoid overflow */
+	time.tv_sec = 0;
+	time.tv_nsec = 0;
```
lukileczo, line 77: I think now both functions `decrement_count_timed_wait_fail_incorrect_timeout` and `decrement_count_timed_wait_fail_too_short_timeout` do the same thing, maybe remove one test? Also I think there's no point in calling `clock_gettime` if the next line sets the time to 0.

lukileczo (approved):
LGTM

libc/Makefile
```diff
@@ -3,4 +3,10 @@ NAME := test-libc
 SRCS := $(wildcard $(LOCAL_DIR)*.c)
 DEP_LIBS := unity
 
+# TODO: remove after updating docker build image to Ubuntu 22.04
```
nalajcie, line 6: IMHO `-lpthread` should work everywhere for portability reasons - so I would enable it on all targets.

As for rationale - in POSIX spec (https://pubs.opengroup.org/onlinepubs/9699919799/utilities/c99.html) it is written that `c99` compiler should support this param (not necessarily adding new library to linking process):
image

damianloew: There is the following error then for `ia32-generic-qemu`

Screenshot from 2022-12-07 18-33-25

nalajcie: so we need to fix it in our buildsystem

agkaminski (dismissed):

libc/Makefile
damianloew: Thanks, it's now fixed.

agkaminski (approved):

PR #158, opened 2022-12-22
psh: add test-exit

damianloew:
JIRA: PD-77

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


psh/test-exit.py
```diff
@@ -0,0 +1,58 @@
...
+                   result='success',
+                   msg="Prompt hasn't been seen after first exit (two psh processes should be run)")
+
+    assert _count_sleep_pshs(p) == already_spawned, "".join(
+        "The count of sleepy psh processes haven't been decreased after calling exit!")
+
+
+def assert_multi_procs(p, count, already_spawned):
+
+    create_psh_processes(p, count, already_spawned)
```
mateusz-bloch, line 38: Reminder: In the newer version of the `test-kill`, the `create_psh_processes` has been modified and no longer has a third argument. However, this does not affect the test's functionality.
```suggestion
    create_psh_processes(p, count)
```

psh/test-exit.py
```diff
@@ -0,0 +1,58 @@
...
+def assert_one_proc(p, already_spawned):
+    psh.assert_exec(p, 'psh', msg="Prompt hasn't been seen when creating new psh process using runfile")
+    assert _count_sleep_pshs(p) == already_spawned + 1, "New psh process not seen after launching psh by runfile!"
+    psh.assert_cmd(
+                   p,
+                   'exit',
+                   result='success',
+                   msg="Prompt hasn't been seen after first exit (two psh processes should be run)")
+
+    assert _count_sleep_pshs(p) == already_spawned, "".join(
```
mateusz-bloch, line 32: ` "".join` not needed only one message is used

psh/test-exit.py
```diff
@@ -0,0 +1,58 @@
...
+        "The count of sleepy psh processes haven't been decreased after calling exit!")
+
+
+def assert_multi_procs(p, count, already_spawned):
+
+    create_psh_processes(p, count, already_spawned)
+    assert _count_sleep_pshs(p) == already_spawned + count, "".join(
```
mateusz-bloch, line 39: ` "".join` not needed only one message is used

psh/test-exit.py
```diff
@@ -0,0 +1,58 @@
+# Phoenix-RTOS
+#
+# phoenix-rtos-tests
+#
+# simple psh exit command test
```
mateusz-bloch, line 5:
```suggestion
# psh exit command test
```

mateusz-bloch (changes requested):
It looks good, but I would consider adding a few more test cases. For example: 
- Checking if 'exit' with an argument executes correctly
- Maybe something along the lines of redirection, like whether the command `exit > file.txt `creates a file.

psh/test-exit.py
damianloew: It's done due to decreasing the length of line.

damianloew:
> It looks good, but I would consider adding a few more test cases. For example:
> 
> * Checking if 'exit' with an argument executes correctly
> * Maybe something along the lines of redirection, like whether the command `exit > file.txt `creates a file.

Thanks for the review, I think that it's not a place to test pipes, I'd focus on checking the proper exit execution there.

psh/test-exit.py
damianloew: You are right, thanks. It's corrected now.

mateusz-bloch (approved):
LGTM

PR #162, opened 2023-02-06
Fix meterfs tests on imx6ull

badochov:
JIRA: RTOS-346

## Description
Add path to meterfs prefix to functions.
Add validation to argv[0].
Fix wrong paths.

## Motivation and Context
Tests were assuming that the mount path of meterfs was `/`.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imx6ull.

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/phoenix-rtos-filesystems/pull/90/files

### DIFF
--- a/meterfs/file_phx.c
+++ b/meterfs/file_phx.c
@@ -233,6 +233,27 @@ int file_eraseAll(void)
 }
 
 
+int file_devInfo(file_fsInfo_t *fsInfo)
+{
+	msg_t msg;
+	meterfs_i_devctl_t *iptr = (meterfs_i_devctl_t *)msg.i.raw;
+	meterfs_o_devctl_t *optr = (meterfs_o_devctl_t *)msg.o.raw;
+
+	file_prepareDevCtl(&msg);
+
+	iptr->type = meterfs_fsInfo;
+
+	TEST_ASSERT_EQUAL(0, msgSend(meterfs.port, &msg));
+
+	fsInfo->filecnt = optr->fsInfo.filecnt;
+	fsInfo->fileLimit = optr->fsInfo.fileLimit;
+	fsInfo->sz = optr->fsInfo.sz;
+	fsInfo->sectorsz = optr->fsInfo.sectorsz;
+
+	return optr->err;
+}
+
+
 int file_init(const char *path)
 {
 	pathPrefix = path;
--- a/meterfs/file_phx.c
+++ b/meterfs/file_phx.c
@@ -3,8 +3,8 @@
  *
  * Meterfs test file abstraction
  *
- * Copyright 2021 Phoenix Systems
- * Author: Aleksander Kaminski, Andrzej Glowinski, Tomasz Korniluk
+ * Copyright 2021, 2023 Phoenix Systems
+ * Author: Aleksander Kaminski, Andrzej Glowinski, Tomasz Korniluk, Hubert Badocha
  *
  * %LICENSE%
  */
@@ -19,8 +19,9 @@
 #include "unity_fixture.h"
 
 static oid_t meterfs;
+static const char *pathPrefix;
 
-static inline void file_prepareDevCtl(msg_t* msg)
+static inline void file_prepareDevCtl(msg_t *msg)
 {
 	msg->type = mtDevCtl;
 	msg->i.data = NULL;
@@ -30,11 +31,19 @@ static inline void file_prepareDevCtl(msg_t* msg)
 }
 
 
+static int lookup_rel(const char *name, oid_t *file, oid_t *dev)
+{
+	char buffer[64];
+	(void)snprintf(buffer, sizeof(buffer), "%s/%s", pathPrefix, name);
+	return lookup(buffer, file, dev);
+}
+
+
 int file_lookup(const char *name)
 {
 	oid_t oid;
 
-	return lookup(name, &oid, NULL);
+	return lookup_rel(name, &oid, NULL);
 }
 
 
@@ -44,7 +53,7 @@ int file_open(const char *name)
 	int err;
 	id_t id;
 
-	if ((err = lookup(name, &msg.i.openclose.oid, NULL)) < 0)
+	if ((err = lookup_rel(name, &msg.i.openclose.oid, NULL)) < 0)
 		return err;
 
 	id = msg.i.openclose.oid.id;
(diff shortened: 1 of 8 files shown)


meterfs/file_phx.c
```diff
@@ -30,11 +31,19 @@ static inline void file_prepareDevCtl(msg_t* msg)
 }
 
 
+static int lookup_rel(const char *name, oid_t *file, oid_t *dev)
+{
+	char buffer[64];
+	(void)snprintf(buffer, sizeof(buffer), "%s%s", pathPrefix, name);
```
agkaminski, line 37:
```suggestion
	if (snprintf(buffer, sizeof(buffer), "%s%s", pathPrefix, name) >= sizeof(buffer)) {
		return -ENAMETOOLONG;
```

meterfs/file_phx.c
```diff
@@ -44,7 +53,7 @@ int file_open(const char *name)
 	int err;
 	id_t id;
 
-	if ((err = lookup(name, &msg.i.openclose.oid, NULL)) < 0)
+	if ((err = lookup_rel(name, &msg.i.openclose.oid, NULL)) < 0)
```
agkaminski, line 56:
```suggestion
	err = lookup_rel(name, &msg.i.openclose.oid, NULL)
	if (err < 0) {
```

meterfs/file_phx.c
agkaminski: Also, do `"%s/%s"` just to be sure (path might be specified without trailing `/`)

meterfs/test_meterfs_miscellaneous.c
```diff
@@ -100,15 +100,16 @@ TEST(meterfs_miscellaneous, resize_bigger)
...
 	file_info_t info = { 2, 2000, 20, 0 };
-	char name[] = "file0";
+	const char *name = "file0";
 	int i;
 
 	TEST_ASSERT_EQUAL(0, file_allocate(name, info.sectors, info.filesz, info.recordsz));
 
-	for (i = 0; i < 5; ++i)
-		TEST_ASSERT_GREATER_OR_EQUAL(0, file_lookup(name));
+	(void)snprintf(common.buffMsg, sizeof(common.buffMsg), "/%s", name);
```
agkaminski, line 108: If you fix `lookup_rel` leading `/` will not be needed

meterfs/test_meterfs_writeread.c
```diff
@@ -72,24 +72,24 @@ TEST_TEAR_DOWN(meterfs_writeread)
...
 	for (i = 0; i < 255; ++i) {
 		snprintf(common.buffMsg, sizeof(common.buffMsg), "iter=%d", i);
-		wroteLen = i % (info.recordsz + 1);
+		writeLen = i % (info.recordsz + 1);
 		i % 2 ? snprintf(common.pattern, sizeof(common.pattern), "aaaaa") : snprintf(common.pattern, sizeof(common.pattern), "zzzzz");
 
-		if (wroteLen) {
-			TEST_ASSERT_EQUAL_MESSAGE(info.recordsz, file_write(common.fd, common.pattern, wroteLen), common.buffMsg);
-			common_readContent(common.fd, info.recordcnt * info.recordsz, common.buffRec, info.recordsz, common.pattern, wroteLen, common.buffMsg);
+		if (writeLen) {
```
agkaminski, line 85: rule 14.4
```suggestion
		if (writeLen != 0) {
```

meterfs/test_meterfs_writeread.c
```diff
@@ -261,10 +261,11 @@ void runner(void)
 
 int main(int argc, char *argv[])
 {
-	file_init(argv[1]);
+	TEST_ASSERT_NOT_NULL(argv[1]);
+	TEST_ASSERT_EQUAL(0, file_init(argv[1]));
 	TEST_ASSERT_EQUAL(0, file_eraseAll());
```
jsarzy, line 266: I'm not doing full review, just a small note. These asserts under the hood use setjmp/longjmp to jump out of failed tests, so they should only be used in testing and setup/teardown functions.

jsarzy (changes requested):

meterfs/test_meterfs_writeread.c
badochov: what should be used in this scenario then? classic assert? If followed by return 1?

jsarzy: You can print message what kind of error occurred and then return with 1. The runner will fail because there will be not output from unity framework, but it will parse the error message and print it to the user.

meterfs/test_meterfs_miscellaneous.c
jsarzy: Do we need these (void) casting?

meterfs/test_meterfs_allocate.c
```diff
@@ -95,10 +98,10 @@ TEST(meterfs_allocate, var_init_args)
 TEST_GROUP_RUNNER(meterfs_allocate)
 {
 	RUN_TEST_CASE(meterfs_allocate, big_file);
-	RUN_TEST_CASE(meterfs_allocate, many_files);
+	/* RUN_TEST_CASE(meterfs_allocate, many_files);
```
agkaminski, line 101: Well, disabling tests is not exactly fixing them ;)

meterfs/test_meterfs_miscellaneous.c
```diff
@@ -15,21 +15,21 @@
...
 
 static void writeReadCheck(file_info_t *info)
 {
-	int i;
+	size_t i;
 
-	for (i = 0; i < info->filesz / info->recordsz; ++i) {
-		snprintf(common.buffMsg, sizeof(common.buffMsg), "a%04d", i);
+	for (i = 0; i < (info->filesz / info->recordsz); ++i) {
+		(void)snprintf(common.buffMsg, sizeof(common.buffMsg), "a%04zu", i);
```
agkaminski, line 27: See warning

meterfs/test_meterfs_miscellaneous.c
agkaminski: @jsarzy 
MISRA Rule 17.7 The value returned by a function having non-void return type shall be used

This rule is a required one

P.S. I guess this `/` is not needed now.

badochov: it fails on PC without it

meterfs/test_meterfs_writeread.c
badochov: Ok I'll add message printing

meterfs/test_meterfs_allocate.c
badochov: Forgot to reenable it, my bad

agkaminski (approved):

PR #163, opened 2023-02-09
libc: enable test on dummyfs targets

damianloew:
## Description

 - libc: enable on dummyfs targets
 - libc: temporary disable truncate/ftruncate up cases
 - libc: adjust setpuids_setsid case in unistd_uids.c to dummyfs targets
 - libc: create tmp and etc if didn't exist
 - libc: increase stack size

## Motivation and Context

 - JIRA: PD-329

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/main.c
+++ b/libc/main.c
@@ -14,13 +14,13 @@
  */
 
 #include <stdlib.h>
+#include <string.h>
 #include "unity_fixture.h"
+#include "common.h"
 
 /* no need for forward declarations, RUN_TEST_GROUP does it by itself */
 void runner(void)
 {
-	setenv("POSIXLY_CORRECT", "y", 1);
-
 	RUN_TEST_GROUP(stdio_fopenfclose);
 	RUN_TEST_GROUP(stdio_line);
 	RUN_TEST_GROUP(stdio_getput);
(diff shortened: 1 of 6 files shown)


libc/main.c
```diff
@@ -15,12 +15,19 @@
...
 /* no need for forward declarations, RUN_TEST_GROUP does it by itself */
 void runner(void)
 {
 	setenv("POSIXLY_CORRECT", "y", 1);
 
+	/* the following files may not be present on dummyfd targets,
+	create them to make libc tests common */
+	create_dir_if_missing("/tmp");
+	create_dir_if_missing("/etc");
+	create_file_if_missing("/etc/passwd", "root:0B1ANiYi45IhxkfmUW155/GBd4IRE=:0:0:root:/:/bin/sh");
```
jsarzy, line 29: Did you make sure that we can use `TEST_ASEERT*` here? It is allowed to call it in runner function or maybe only in `RUN_GROUP_TEST*`?

damianloew: Yeah, it wasn't allowed. Thanks for this suggestion and please check the new approach

libc/main.c
```diff
@@ -39,13 +39,68 @@ void runner(void)
...
+{
+	struct stat buffer;
 
-	unsetenv("POSIXLY_CORRECT");
+	if ((stat(path, &buffer) != 0) && (errno == ENOENT)) {
+		if (mkdir(path, S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH) != 0) {
+			fprintf(stderr, "Creating %s directory by mkdir failed: %s\n", path, strerror(errno));
+			return -1;
+		}
+	}
```
agkaminski, line 55: What if `errno != ENOENT`?

libc/main.c
```diff
@@ -39,13 +39,68 @@ void runner(void)
...
+static int libc_createFileIfMissing(const char *path, const char *charData)
+{
+	struct stat buffer;
+
+	if ((stat(path, &buffer) != 0) && (errno == ENOENT)) {
+		if (_create_file(path, charData) != 0) {
+			fprintf(stderr, "Creating %s file failed: %s\n", path, strerror(errno));
+			return -1;
+		}
+	}
```
agkaminski, line 71: What if `errno != ENOENT`?

agkaminski, line 71:
```suggestion
	if (stat(path, &buffer) != 0) {
		if (errno == ENOENT) {
			if (_create_file(path, charData) != 0) {
				fprintf(stderr, "Creating %s file failed: %s\n", path, strerror(errno));
				return -1;
			}
			else {
				return -1;
			}
		}
	}
```

libc/main.c
```diff
@@ -39,13 +39,80 @@ void runner(void)
...
+
+	if (setenv(var, "y", 1) != 0) {
+		fprintf(stderr, "Setting %s environment variable failed: %s\n", var, strerror(errno));
+		return 1;
+	}
+
+	/* the following files may not be present on dummyfd targets,
+	create them to make libc tests common */
+	if (libc_createDirIfMissing("/tmp") < 0) {
+		return 1;
```
agkaminski, line 101: Sorry, missed that on the previous review. If it is important to `unsetenv` before exit, we should do this on these error exit paths too

damianloew: Yeah, but I think not so important to check its return code every time, so I would prefer sth like this:
```C
{
	const char *var = "POSIXLY_CORRECT";

	if (setenv(var, "y", 1) != 0) {
		fprintf(stderr, "Setting %s environment variable failed: %s\n", var, strerror(errno));
		return 1;
	}

	/* the following files may not be present on dummyfd targets,
	create them to make libc tests common */
	if (libc_createDirIfMissing("/tmp") < 0) {
		unsetenv(var);
		return 1;
	}
	if (libc_createDirIfMissing("/etc") < 0) {
		unsetenv(var);
		return 1;
	}
	if (libc_createFileIfMissing("/etc/passwd", "root:0B1ANiYi45IhxkfmUW155/GBd4IRE=:0:0:root:/:/bin/sh")) {
		unsetenv(var);
		return 1;
	}

	UnityMain(argc, (const char **)argv, runner);

	if (unsetenv(var) != 0) {
		fprintf(stderr, "Unsetting %s environment variable failed: %s\n", var, strerror(errno));
		return 1;
	}

	return 0;
}
```

agkaminski (approved):

libc/main.c
```diff
@@ -39,13 +39,80 @@ void runner(void)
...
+			return -1;
+		}
+	}
+
+	return 0;
+}
+
+
+/* crete file with optional data (may be NULL) unless it exists */
+static int libc_createFileIfMissing(const char *path, const char *charData)
```
nalajcie, line 68: IMHO naming functions which are not part of a libc - `libc_something` is really misleading

PR #164, opened 2023-02-15
Maska989/test scanf

maska989:
## Description

Added fscanf/sscanf/vfscanf and ssfcanf test cases
## Motivation and Context

[CI-193](https://jira.phoenix-rtos.com/browse/CI-193)

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


damianloew (changes requested):
Just a quick review. Above that, please add more tests with edge values as we were talking personally, so please consider the following values for each specifier, by long int example: You want to scan using %ld:

 - long_min
 - long_min/2 (or alternatively some other positive value)
 - 0
 - long_max/2 (or alternatively some other negative value)
 - long_max

nalajcie:
Your code has too many styling errors, please fix them locally and setup your environment correctly (eg. pre-commit clang-format check - see https://confluence.phoenix-rtos.com/x/1QkW for details)

libc/stdio_scanf.c
```diff
@@ -0,0 +1,1590 @@
...
+		TEST_IGNORE();
+	#endif	
+	fscantest_num("%ti", td, 0);
+
+}
+
+
+TEST(stdio_fscanf, format_unsigned_int)
+{
+	int i;
```
nalajcie, line 264: these values should be unsigned

libc/stdio_scanf.c
```diff
@@ -0,0 +1,1590 @@
...
+	long int ld;
+	long long int lld;
+	short int hi;
+	intmax_t jd;
+	size_t zd;
+	ptrdiff_t td;
+	char hh;
+
+	/*Unsigned decimal integer*/
+	fscantest_num("%u", i, 0);
```
nalajcie, line 274: the macros for testing NEED to be for unsigned types

nalajcie (dismissed):

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3506 @@
...
+
+	TEST_ASSERT_EQUAL_INT64(LLONG_MIN, int_min);
+	TEST_ASSERT_EQUAL_INT64(LLONG_MAX, int_max);
+	TEST_ASSERT_EQUAL_INT64(0, int_zero);
+	TEST_ASSERT_EQUAL_INT64(LLONG_MIN / 2, int_half_min);
+	TEST_ASSERT_EQUAL_INT64(LLONG_MAX / 2, int_half_max);
+}
+
+
+TEST(SCANF_EXT_D, JD)
```
damianloew, line 311: Please also check `uintmax_t`

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3506 @@
...
+	TEST_ASSERT_EQUAL_INT64(LONG_MAX / 2, int_half_max);
+}
+
+
+TEST(SCANF_EXT_D, ZD)
+{
+	char FreeSpace[400];
+	size_t int_max, int_zero, int_half_max;
+
+	sprintf(FreeSpace, "%zd %zd %d", SSIZE_MAX, SSIZE_MAX / 2, 0);
```
damianloew, line 361: Please resolve the following warnings: I think the best option there is to cast. As it's described in PR's checklist we don't want to increase the number of warnings. Now you can check all warnings related to your code in PR->Files changed 

Screenshot from 2023-03-03 13-04-33

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3506 @@
...
+	TEST_ASSERT_EQUAL_INT64(0, int_zero);
+	TEST_ASSERT_EQUAL_INT64(SSIZE_MAX / 2, int_half_max);
+}
+
+
+TEST(SCANF_EXT_D, TD)
+{
+	char FreeSpace[400];
+	ptrdiff_t int_max, int_min, int_zero, int_half_min, int_half_max;
+	sprintf(FreeSpace, "%ld %ld %d %ld %ld", PTRDIFF_MAX, PTRDIFF_MAX / 2, 0, PTRDIFF_MIN / 2, PTRDIFF_MIN);
```
damianloew, line 398: There are also some warnings.

damianloew (changes requested):
Just reviewing to `lld` test case for now, please apply those suggestions to the rest of the code approprietelly.

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3701 @@
...
+
+	TEST_ASSERT_EQUAL_INT(1, fscanf(rd, "%c", &chr));
+	TEST_ASSERT_EQUAL_CHAR(T, chr);
+
+	TEST_ASSERT_EQUAL_INT(1, vsscanf_wrapper(tested_str, "%c", &chr));
+	TEST_ASSERT_EQUAL_CHAR(T, chr);
+
+	TEST_ASSERT_EQUAL_INT(1, sscanf(tested_str, "%c", &chr));
+	TEST_ASSERT_EQUAL_CHAR(T, chr);
+
```
niewim19, line 3301: This part is redundant to the next. Both checks if the read is correct if there is expected character. I think that you can make it shorter (3 characters seems of), write it to a string/file `str`, and then perform a three character read and check each character match. This way with just one string and one write you check if:
 - first read is ok
 - consecutive reads are ok
 - read before EOF are ok

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3701 @@
...
+	TEST_ASSERT_EQUAL_INT(1, sscanf(tested_str, "%c", &chr));
+	TEST_ASSERT_EQUAL_CHAR(tstr[0], chr);
+}
+
+TEST(stdio_scanf_str_chr_ptr, str_path)
+{
+
+	char tested_str[50];
+	char chr[50];
```
niewim19, line 3325:
```suggestion
	char tested_str[50];
	char chr[50];
```

 - Please follow the variable naming convention
 - these names are a little misleading - both are strings, yet one is called chr (this applies in the next tests also)
 - magic sizes of buffers. what about a macro?

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3701 @@
...
+	TEST_ASSERT_EQUAL_INT(1, sscanf(tested_str, "%s", chr));
+	TEST_ASSERT_EQUAL_STRING(std_path, chr);
+}
+
+TEST(stdio_scanf_str_chr_ptr, str_special)
+{
+
+	char tested_str[50];
+	char chr[50];
+	char spc[] = "\4399\0ns";
```
niewim19, line 3351: I doubt if `#` is a special character. Using `\43` is just different way of passing a character value (without an actual character. 

When it comes to the `\0` character it just obscures the next part of the string but we really don`t check it. We don`t know it the `ns` was read or not because by checking it with `TEST_ASSERT_EQUAL_STRING` we only assess the part before `\0` character. We don't check what happened after that, so `ns` is just not used. The idea of something after `\0` is good. Maybe assert equality of arrays, not strings?

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3701 @@
...
+	TEST_ASSERT_EQUAL_INT(1, sscanf(tested_str, "%s", chr));
+	TEST_ASSERT_EQUAL_STRING("#99", chr);
+}
+
+TEST(stdio_scanf_str_chr_ptr, str_percent)
+{
+
+	char tested_str[50];
+	char chr[50];
+	char spc[] = "%tak";
```
niewim19, line 3377: Random string :) Maybe use a macro for a placeholder strings?

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3701 @@
...
+	TEST_ASSERT_EQUAL_INT(123, val);
+}
+
+TEST(stdio_scanf_str_chr_ptr, str_no_spaces)
+{
+
+	char tested_str[50];
+	char chr[50];
+	int val_hex;
+	char spc[] = "somethingishappeneing123rightnow/0x0005";
```
niewim19, line 3442: Please use more generic string ;)

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3701 @@
...
+	TEST_ASSERT_EQUAL_INT(1, sscanf(tested_str, "%[^\n]%*c", chr));
+	TEST_ASSERT_EQUAL_STRING(spc, chr);
+}
+
+TEST(stdio_scanf_str_chr_ptr, pointers)
+{
+
+	char tested_str[50];
+	void **chr;
+	char spc[] = "0x12ab34cd";
```
niewim19, line 3495: Please use one of _hexspeak_ values to identify that this is particularly not a valid address.

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3701 @@
...
+	TEST_ASSERT_EQUAL_PTR(std_path, chr);
+}
+
+TEST(stdio_scanf_str_chr_ptr, str_chain)
+{
+
+	char tested_str[1500];
+	char chr[50];
+	char spc[1500];
+	char ret[1500];
```
niewim19, line 3543: Does it have to be so big?

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3701 @@
...
+	TEST_ASSERT_EQUAL_INT(1, sscanf(tested_str, "%s", ret));
+	TEST_ASSERT_EQUAL_STRING(spc, ret);
+}
+
+TEST(stdio_scanf_str_chr_ptr, str_overflow)
+{
+
+	char tested_str[1500];
+	char chr[10];
+	char buff[] = "Something is writen here with more chars then chr";
```
niewim19, line 3576: Few remarks about this test as a whole:
 - this test tests undefined behavior as defined in posix:
   > If this object does not have an appropriate type, or if the result of the conversion cannot be represented in the space provided, the behavior is undefined.
 - the `tested_str` is way too big to hold a one sentence 
 - I think this test is not memory safe as it may attempt to write outside of buffer

niewim19 (changes requested):
Hey - I have gone through this PR and made general remarks about coding style. As this test is very long I have made remarks mainly in the first occurrences of some mistakes so please be when correcting them to check if they occur in the rest of the file. General remarks can be summarized to:
 - please do not resolve conversations if they are not corrected or explicitly told not to be implemented. Now there is need to go through most of the previous remarks and check if they are corrected because in many cases they are not.
 - please foolow out coding convention when it comes to variable names. You use snake_case when we use camelCase for variables. More often in this code there is no even a need for camelCase as many variables can be cut short to one word or `<letter>word` construction
 -  there are many cases when you check for specific type specifier but construct the test data without the use of it.

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3399 @@
...
+	sprintf(buf, "%s", spc);
+	create_file(stdPath, buf);
+	rd = fopen(stdPath, "r");
+
+	TEST_ASSERT_EQUAL_INT(1, vfscanfWrapper(rd, "%[^\n]%*c", chr));
+	TEST_ASSERT_EQUAL_STRING(spc, chr);
+
+	rewind(rd);
+
+	TEST_ASSERT_EQUAL_INT(1, fscanf(rd, "%[^\n]%*c", chr));
```
niewim19, line 3241: Please describe why such format is used.

libc/stdio_scanf.c
```diff
@@ -0,0 +1,1270 @@
...
+TEST_GROUP_RUNNER(stdio_scanf_specials)
+{
+	RUN_TEST_CASE(stdio_scanf_specials, special_percent);
+	RUN_TEST_CASE(stdio_scanf_specials, secial_space_before_hex);
+	RUN_TEST_CASE(stdio_scanf_specials, secial_without_space_before_hex);
+	RUN_TEST_CASE(stdio_scanf_specials, secial_empty_file);
+	RUN_TEST_CASE(stdio_scanf_specials, secial_strings);
+	RUN_TEST_CASE(stdio_scanf_specials, secial_numbers);
+	/*RUN_TEST_CASE(stdio_scanf_specials, secial_string_char);*/
+}
```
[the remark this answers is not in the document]
niewim19: Please do not resolve remarks that are not corrected.

niewim19 (changes requested):
Thanks for implementing previous remarks! Review of such a big file (3,5k lines) may take some iterations before the merge so don`t be bothered by the amount of remarks :)

General remarks:
 - remove each-test-file-creation feature and use unity fixtures `TEST_SETUP` and `TEST_TEAR_DOWN`. This will save a lot of repetitive code.
 - Each `j` length modifier test, regardless to the type specifier (signed/unsigned) tests for both `uintmax_t` and `intmax_t` which is somewhat wrong as we print value using wrong specifier and read it using wrong specifier which, in the end, gives us falsely positive result
 - in tests of uncommon type specifiers (like `[`) please describe what the test is actually testing as it is sometimes quite hard to deduce just from looking at the format specifier.

niewim19 (changes requested):
Thank you for implementing suggested changes. This PR has undergone a lot of changes since its creation and i think that we are almost done with it :+1:. It looks much more clear and structured now.

Although I have added about 30 remarks most of them are pure editorial and you will sweep through them quickly (some are repeated multiple times as these tests are quite repetitive).

Summing up:
 - Take care of file handling: 
   - creating files with `create_file` in `TEST_SETUP` is unnecessary. `fopen` + return value check will do. 
   - don`t reopen files which are already opened as you have multiple memory leaks in tests because of this
   - **close files**: you actually never close these files! Each file pointer opened via `fopen()` must be closed using `fclose()`. Also don't use `unlink()` it is not designed for that.
 - in integer tests, in cases with `uintmax_t`/`intmax_t` you are double checking these length specifiers - regardless of the type specifier signage you almost always check both signed and unsigned values. These tests pass but only by accident.
 - I think that you should rethink case-by-case the testgroup `stdio_scanf_squareBraces` with @damianloew  as format strings used in many are not exactly doing what they are described to do
 - please follow the test ignore pattern mentioned in the review.

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3509 @@
...
+
+		TEST_ASSERT_EQUAL_INT(1, vfscanfWrapper(rd, "%c", &c));
+		TEST_ASSERT_EQUAL_CHAR(i, c);
+
+		fseek(rd, i - 1, SEEK_SET);
+
+		TEST_ASSERT_EQUAL_INT(1, fscanf(rd, "%c", &c));
+		TEST_ASSERT_EQUAL_CHAR(i, c);
+
+		/* This fseek is used because of issue #639 */
```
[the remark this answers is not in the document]
damianloew: Yeah, we can run it only on phoenix, but I see that fix is coming, so I hope we will remove this line.

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3510 @@
...
+	rewind(rd);
+
+	TEST_ASSERT_EQUAL_INT(1, fscanf(rd, "%% %s", chr));
+	TEST_ASSERT_EQUAL_STRING("yes", chr);
+
+	TEST_ASSERT_EQUAL_INT(1, vsscanfWrapper(buff, "%% %s", chr));
+	TEST_ASSERT_EQUAL_STRING("yes", chr);
+
+	TEST_ASSERT_EQUAL_INT(1, sscanf(buff, "%% %s", chr));
+	TEST_ASSERT_EQUAL_STRING("yes", chr);
```
damianloew, line 3036: Let's check, whether % does not parse not existent percents:

```suggestion
	char buff[BUFF_LEN_STR] = "%yes% --- yes";
	char correct[BUFF_LEN_STR] = { 0 };
	char wrong[BUFF_LEN_STR] = { 0 };

	fprintf(rd, "%s", buff);
	rewind(rd);

	TEST_ASSERT_EQUAL_INT(1, vfscanfWrapper(rd, "%% %s%% --- %% %s", correct, wrong));
	TEST_ASSERT_EQUAL_STRING("yes", correct);
	TEST_ASSERT_EQUAL_STRING("", wrong);
	
	// and please correct the next lines respectively

	rewind(rd);

	TEST_ASSERT_EQUAL_INT(1, fscanf(rd, "%% %s", chr));
	TEST_ASSERT_EQUAL_STRING("yes", chr);

	TEST_ASSERT_EQUAL_INT(1, vsscanfWrapper(buff, "%% %s", chr));
	TEST_ASSERT_EQUAL_STRING("yes", chr);

	TEST_ASSERT_EQUAL_INT(1, sscanf(buff, "%% %s", chr));
	TEST_ASSERT_EQUAL_STRING("yes", chr);
```

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3510 @@
...
+	TEST_ASSERT_EQUAL_STRING("emipsum", chr);
+	TEST_ASSERT_EQUAL_STRING("dolorsit", chr2);
+}
+
+
+TEST(stdio_scanf_squareBrackets, squareBrackets_spaces)
+{
+	int valHex;
+	char buff[BUFF_LEN_STR] = "LoremIpsumDolorSitAmet,Vestibulum123AnteIpsumPrimis0x0005";
+	char chr[BUFF_LEN_STR] = { 0 };
```
damianloew, line 3262:
```suggestion
	char res[BUFF_LEN_STR] = { 0 };
```

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3510 @@
...
+	TEST_ASSERT_EQUAL_STRING("emipsum", chr);
+	TEST_ASSERT_EQUAL_STRING("dolorsit", chr2);
+
+	TEST_ASSERT_EQUAL_INT(2, sscanf(buff, "Lor%[^do]%s", chr, chr2));
+	TEST_ASSERT_EQUAL_STRING("emipsum", chr);
+	TEST_ASSERT_EQUAL_STRING("dolorsit", chr2);
+}
+
+
+TEST(stdio_scanf_squareBrackets, squareBrackets_spaces)
```
damianloew, line 3258: Why spaces? I don't see any spaces there. To be honest I don't see anything new comparing to the previous test case.

damianloew (changes requested):
I will consider how to how to tidy the rest of squareBrackets cases and let you know in next suggestions. Please apply the current ones.

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3510 @@
...
+TEST(stdio_scanf_aefg, f)
+{
+	/* <posix incmpliance> Scanf doesn't support longer floating-point numbers
+
+   Problems occur when scanf needs to read big numbers after coma.
+   example: 0.234234345345 scanf will read
+   example 2.212121e-100 scanf will doesn't read because this is long float after coma
+
+   Issue link: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/652
+   */
```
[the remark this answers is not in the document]
niewim19: Ok, that`s on me - this style of comments regarding posix incompatibility was firstly used by me for example [here](https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/44b967805c2fdba1971623ee82cf778c3bc9460e/libc/unistd_file.c#L314-L320). We should look for `<posix incompliance>` marker and correct these comments in a spare time.

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3701 @@
...
+		&int_min);
+	rewind(rd);
+	fscanf(rd, ext, &int_max, &int_half_max, &int_zero, &int_half_min, &int_min);
+	vsscanf_wrapper(tested_str, ext, &int_max, &int_half_max, &int_zero,
+		&int_half_min, &int_min);
+	sscanf(tested_str, ext, &int_max, &int_half_max, &int_zero, &int_half_min,
+		&int_min);
+
+#pragma GCC diagnostic pop
+}
```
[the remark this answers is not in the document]
damianloew: It was fine to call it without checking result as it's undefined behavior. The fact is
```C
	create_file(std_path, tested_str);
	rd = fopen(std_path, "r");
```
is called twice, and those 2 lines should be removed. Above that such case can be misleading and a comment should be left there. Sth like /* the result is not checked intentionally, as POSIX indicates undefined behavior. We only assure that test won't crash */

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3510 @@
...
+	TEST_ASSERT_EQUAL_STRING("Loremi", chr);
+	TEST_ASSERT_EQUAL_STRING("psumdolorsit", chr2);
+
+	rewind(rd);
+	memset(chr, 0, sizeof(chr));
+	memset(chr2, 0, sizeof(chr2));
+
+	/* 
+	 * Scanf will try to find elements after "Lor" and save them to "chr" until first
+	 * occurrence of "do" everything after that with content of [] will be saved to "chr2"
```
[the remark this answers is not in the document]
maska989: "with content of []" is needed to explain the read elements and read cursor placed in the sentence before []

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3769 @@
...
+
+	TEST_ASSERT_EQUAL_INT(1, test_vsscanfWrapper(buff, "%[^\n]%*c", res));
+	TEST_ASSERT_EQUAL_STRING(buff, res);
+
+	TEST_ASSERT_EQUAL_INT(1, sscanf(buff, "%[^\n]%*c", res));
+	TEST_ASSERT_EQUAL_STRING(buff, res);
+}
+
+
+TEST(stdio_scanf_squareBrackets, s_mix)
```
damianloew, line 3615: To remove as `stdio_scanf_rest` with `mix` test case will be created.

```suggestion
```

damianloew (changes requested):

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3769 @@
...
+	RUN_TEST_CASE(stdio_scanf_aefg, G);
+	RUN_TEST_CASE(stdio_scanf_aefg, inf_nan_f);
+	RUN_TEST_CASE(stdio_scanf_aefg, inf_nan_a);
+	RUN_TEST_CASE(stdio_scanf_aefg, inf_nan_e);
+	RUN_TEST_CASE(stdio_scanf_aefg, inf_nan_g);
+	remove(TESTFILE_PATH);
+}
+
+
+TEST_GROUP_RUNNER(stdio_scanf_cspn)
```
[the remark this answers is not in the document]
maska989: This is a great idea, but unity is unable to hold the '%' sign as the name of TEST_GROUP and TEST_CASE

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3510 @@
...
+	char buff[BUFF_LEN_STR] = "Loremipsumdolorsit";
+	char chr[BUFF_LEN_STR] = { 0 };
+	char chr2[BUFF_LEN_STR] = { 0 };
+
+	fprintf(rd, "%s", buff);
+	rewind(rd);
+
+	/* 
+	 * Scanf will read the string and match word "psu" as a point of stop and save this is "chr"
+	 * After that it will set the file position indicator before "psu"
```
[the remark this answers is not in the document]
damianloew: Right, resolved.

libc/stdio_scanf.c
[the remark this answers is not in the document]
damianloew: Then please change `occurrence of "do" everything after that with content of [] will be saved to "chr2"` to `occurrence of "do" everything after that including this "do" will be saved to "chr2"`

libc/stdio_scanf.c
```diff
@@ -0,0 +1,3769 @@
...
+
+
+TEST_GROUP_RUNNER(stdio_scanf_cspn)
+{
+	RUN_TEST_CASE(stdio_scanf_cspn, c);
+	RUN_TEST_CASE(stdio_scanf_cspn, c_ascii);
+	RUN_TEST_CASE(stdio_scanf_cspn, s_path);
+	RUN_TEST_CASE(stdio_scanf_cspn, s_torn);
+	RUN_TEST_CASE(stdio_scanf_cspn, s_pick);
+	RUN_TEST_CASE(stdio_scanf_cspn, percent);
```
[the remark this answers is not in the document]
maska989: This is a great idea, but unity is unable to hold the '%' sign as the name of TEST_GROUP and TEST_CASE

libc/stdio_scanf.c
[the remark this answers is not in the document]
nalajcie: it's not unity, just a C standard - no special chars can be present in function/variable names

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4210 @@
...
+TEST(stdio_scanf_rest, m_s)
+{
+/* Disabled because of issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/667 */
+#ifdef __phoenix__
+	TEST_IGNORE();
+#endif
+
+	const char *lorem = "LoremIpsumDolorSitAmet,Vestibulum";
+	char *res;
+	char chrArray[BUFF_LEN];
```
damianloew, line 3861:
```suggestion
	char *res, chrArray[BUFF_LEN];
```

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4210 @@
...
+	int i;
+
+	for (i = 0; i < BUFF_LEN - 1; i++) {
+		chrArray[i] = 'a';
+		if (i == BUFF_LEN - 5) {
+			chrArray[i] = 'b';
```
damianloew, line 3867: Why do we want to have this 'b' inside the tested string? In my opinion not needed.

```suggestion
		if (i == BUFF_LEN - 5) {
			chrArray[i] = 'b';
```

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4210 @@
...
+	TEST_ASSERT_EQUAL_STRING("123", res3);
+	free(res1);
+	free(res2);
+	free(res3);
+}
+
+
+/* Tests with %mc are disabled because of problems with host-generic-pc on git-hub ci */
+IGNORE_TEST(stdio_scanf_rest, m_c)
+{
```
damianloew, line 4015:
```suggestion
/* Address sanitizer used on Ubuntu 22.04 fails in such case, because of the following issue:
 * https://github.com/llvm/llvm-project/issues/61768
 */
#ifdef __linux__
__attribute__((no_sanitize_address))
#endif
TEST(stdio_scanf_rest, m_c)
{
/* Disabled because of issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/667 */
#ifdef __phoenix__
	TEST_IGNORE();
#endif
```

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4210 @@
...
+	RUN_TEST_CASE(stdio_scanf_squareBrackets, s_mix);
+	remove(TESTFILE_PATH);
+}
+
+
+TEST_GROUP_RUNNER(stdio_scanf_rest)
+{
+	RUN_TEST_CASE(stdio_scanf_rest, modifiers_mix);
+	// RUN_TEST_CASE(stdio_scanf_rest, m_s);
+	// RUN_TEST_CASE(stdio_scanf_rest, m_brackets);
```
damianloew, line 4206: Please uncomment those cases.

damianloew (changes requested):

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4291 @@
...
+
+
+TEST(stdio_scanf_cspn, ptr)
+{
+	char buff[BUFF_LEN_STR];
+	void *const expPtr = (void *)0xDEADBEEF;
+	void *ptrVal;
+	char *const expPtrZero = (void *)0x00000000;
+	void *const expPtrMax = (void *)INTPTR_MAX;
+	void *const expPtrMin = (void *)INTPTR_MIN;
```
damianloew, line 3292:
```suggestion
	char buff[BUFF_LEN_STR];
	void *const expPtr = (void *)0xDEADBEEF;
	char *const expPtrZero = (void *)0x00000000;
	void *const expPtrMax = (void *)INTPTR_MAX;
	void *const expPtrMin = (void *)INTPTR_MIN;
	void *ptrVal;
```

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4291 @@
...
+			buff[i - 1] = i;
+		}
+	}
+
+	fprintf(filep, "%s", buff);
+	rewind(filep);
+
+	strcpy(format, "%[");
+	strcat(format, buff);
+	strcat(format, "]");
```
damianloew, line 3787: Maybe:
```suggestion
        sprintf(format, "%%[%s]", buff);
```

damianloew (changes requested):
I hope it will be the final review. Some general remark: you don't have to change it in this file, but please remember about it:

```C
	char buff[BUFF_LEN] = { 0 };
	char max, min, zero, hmin, hmax;
```

we can simplify it to:

```C
	char max, min, zero, hmin, hmax, buff[BUFF_LEN] = { 0 };
```

Above that, there is a lot of code, so we can split it to a few commits:

 - ibc: scanf tests: add test groups for d, i, o, u, x specifiers
 - libc: scanf tests: add test groups for a, e, f, g specifiers
 - libc: scanf tests: add test groups for c, s, p, n specifiers
 - libc: scanf tests: add test groups for *, m and mixed specifiers

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4291 @@
...
+	RUN_TEST_CASE(stdio_scanf_squareBrackets, simple);
+	RUN_TEST_CASE(stdio_scanf_squareBrackets, crircumflex);
+	RUN_TEST_CASE(stdio_scanf_squareBrackets, pos);
+	RUN_TEST_CASE(stdio_scanf_squareBrackets, white_spaces);
+	RUN_TEST_CASE(stdio_scanf_squareBrackets, ascii);
+	remove(TESTFILE_PATH);
+}
+
+
+TEST_GROUP_RUNNER(stdio_scanf_rest)
```
damianloew, line 4283: Please also add a test case that checks different `optional non-zero decimal integers` before a few specifiers, please focus on edge cases. For example getting "%10s", when a string contains 11 elements, various cases with integers, floating point numbers

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4300 @@
...
+
+	char buff[BUFF_LEN], valStr[BUFF_LEN_STR], str[] = "LoreIpsumDolorSitAmet";
+	int intMax = 2147483647, intMin = -2147483647, valIntMin, valIntMax;
+	float fltMax = 3.40282347e+7F, fltMin = 3.40282347e-4F, valFltMin, valFltMax;
+
+	sprintf(buff, "%s %d %d %f %f", str, intMax, intMin, fltMax, fltMin);
+	fprintf(filep, "%s", buff);
+	rewind(filep);
+
+	TEST_ASSERT_EQUAL_INT(5, fscanf(filep, "%4s %*s %5d %*d %5d %*d %5f %*f %f", valStr, &valIntMax, &valIntMin, &valFltMax, &valFltMin));
```
damianloew, line 4116: "%4s %*s %5d %*d %5d %*d %5f %*f %f" repeats a few times, please use  `const char *format` as in the most of cases above. Look through the rest of code and please replace the similar occurrences.

damianloew (changes requested):
Please create an issue in phoenix-rtos-tests about missing errno coverage.

niewim19 (dismissed):

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4371 @@
...
+	char thirdWord[MAX_TESTSTR_WORDLEN];
+	char fourthWord[MAX_TESTSTR_WORDLEN];
+	char decimalWord[MAX_TESTSTR_WORDLEN];
+	char hexWord[MAX_TESTSTR_WORDLEN];
+	const char *format = "%s %s %s %s amet,Vestibulum ante ipsum primis in faucibus orci luctus %s et ultrices posuere cubilia "
+						 "curae %s";
+	fprintf(filep, "%s", buff);
+	rewind(filep);
+
+	/* clang-format off */
```
nalajcie, line 3136: why `clang-format off`?

if needed, please at least fix style by hand to be as compliant as possible

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4371 @@
...
+{
+/* Disabled because of issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/667 */
+#ifdef __phoenix__
+	TEST_IGNORE();
+#endif
+
+/* Address sanitizer used on Ubuntu 22.04 fails in such case, because of the following issue:
+ * https://github.com/llvm/llvm-project/issues/61768
+ */
+#ifndef __phoenix__
```
nalajcie, line 4107: IMHO would be better to disable this test only if AddressSanitizer is enabled.
Something like: 
```c
#if __has_feature(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
```

(taken from https://github.com/llvm/llvm-project/blob/1a2eaebc09c6a200f93b8beb37130c8b8aab3934/compiler-rt/include/sanitizer/asan_interface.h#L51)

libc/stdio_scanf.c
```diff
@@ -0,0 +1,4371 @@
...
+	 * (after finding a white sign in this case)
+	 */
+	TEST_ASSERT_EQUAL_INT(3, fscanf(filep, format, res1, res2, res3));
+	TEST_ASSERT_EQUAL_STRING("loremIPSUM", res1);
+	TEST_ASSERT_EQUAL_STRING("IPSUM", res2);
+	TEST_ASSERT_EQUAL_STRING("dolor", res3);
+
+	rewind(filep);
+
+	TEST_ASSERT_EQUAL_INT(3, test_vfscanfWrapper(filep, format, res1, res2, res3));
```
nalajcie, line 3780: for test to be more thorough you should reset the value of all output params before each call - right now you're assuming that if *scanf returns N it means N output params were set. If there would be an error of not setting one of them - the test would still pass because the variable holds the (valid) result) from the previous call.

This is cumbersome so I'm not enforcing the change right now - just be sure to know (and maybe write in comment) the current limitation.

Applies everywhere where *scanf function is called.

nalajcie (commented):
changing review to `comment`, skimmed through code and pointed out some oddities :).

No throughout code review was done by me as others did it already :medal_sports: .

libc/stdio_scanf.c
maska989: A bit of work but added. Thanks 4 suggestion.

libc/stdio_scanf.c
maska989: There are some problems with __has_feature but is easy to go over it using only ifdef no_SANITIZE_ADDRESS and checking if his value is 1, which means is active

libc/stdio_scanf.c
maska989: Bit of leftovers after the cleanup situation with formats (Used because of too long format specifiers).

damianloew (dismissed):
Thanks for the last changes, I think we can merge this main part of scanf tests. Missing cases that were problematic/not possible to cover now are mentioned in https://github.com/phoenix-rtos/phoenix-rtos-tests/issues

libc/stdio_scanf.c
nalajcie: isn't `defined(__SANITIZE_ADDRESS__)` enough? That's how the ASAN internally checks if it's enabled.

if not, please make it a single line
`#if defined(__SANITIZE_ADDRESS__) && (__SANITIZE_ADDRESS__ == 1)`

maska989: You were right just `#if defined(__SANITIZE_ADDRESS__)` was enough.

damianloew (approved):

nalajcie:
typo in the last commit message :P

image

damianloew (approved):

PR #166, opened 2023-02-24
psh: fix links case in runfile applet test

damianloew:
## Description

psh: improve touch-rootfs test
psh: fix links case in runfile applet test
## Motivation and Context

JIRA: CI-217
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/psh/test-touch-rootfs.py
+++ b/psh/test-touch-rootfs.py
@@ -15,18 +15,18 @@ import psh.tools.psh as psh
 from psh.tools.common import assert_mtime
 
 
-def assert_hardlinks(p):
+def assert_bin(p):
+    ''' Asserts that all files in /bin are able to be touched'''
     dates = {}
-    # all psh commands are hardlinks
-    for hardlink in psh.get_commands(p):
-        hardlink_path = f'/bin/{hardlink}'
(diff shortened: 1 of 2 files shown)


psh/test-touch-rootfs.py
```diff
@@ -15,18 +15,18 @@
 from psh.tools.common import assert_mtime
 
 
-def assert_hardlinks(p):
+def assert_bin(p):
+    ''' Asserts that all executables/hardlinks available in bin are able to touch'''
```
nalajcie, line 19:
```suggestion
    ''' Asserts that all files in /bin are able to be touched'''
```
There might be also symlinks in /bin (I don't know if tourching symlink should change it's mtime (or the mtime of the target) so this test might still fail in the future.

Edit, it will probably fail with symlinks, there is a special option for POSIX touch not to follow symlinks:
image


IMHO we can leave as is for now and worry in the future.

nalajcie (commented):
small nitpicks, otherwise LGTM

damianloew:
> small nitpicks, otherwise LGTM

Thanks, done.

nalajcie (approved):

PR #169, opened 2023-02-28
Maska989/testStrchr

maska989:
## Description

Added Strchr and Strrchr test cases
## Motivation and Context

JIRA: CI-211

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/string_lenchr.c
+++ b/libc/string_lenchr.c
@@ -3,10 +3,13 @@
  *
  * POSIX.1-2017 standard library functions tests
  * HEADER:
- *    - string.h
+ * 	- string.h
  * TESTED:
- *    - strlen()
- *    - strnlen()
+ * 	- strlen()
+ * 	- strnlen()
+ * 	- strchr()
+ * 	- strrchr()
+ * 	- memchr()
  *
  * Copyright 2023 Phoenix Systems
  * Author: Damian Modzelewski
@@ -29,6 +32,7 @@
 
 
 TEST_GROUP(string_len);
+TEST_GROUP(string_chr);
 
 
 TEST_SETUP(string_len)
@@ -43,33 +47,33 @@ TEST_TEAR_DOWN(string_len)
 
 TEST(string_len, ascii)
 {
-	const char empty[] = "";
-	const char pangram[] = "The quick brown fox jumps over the lazy dog";
-	const char torn[] = "foo\0bar";
-	const char doubleNul[] = "\0\0abc";
-	const char specials[] = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~";
-	const char whites[] = " \v\t\r\n";
+	const char empty[] = "",
+			   pangram[] = "The quick brown fox jumps over the lazy dog",
+			   torn[] = "foo\0bar",
+			   doubleNul[] = "\0\0abc",
+			   specials[] = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~",
+			   whites[] = " \v\t\r\n";
 	char asciiSet[128] = { 0 };
 	int sz, i;
 
 	TEST_ASSERT_EQUAL_INT(0, strlen(""));
 
-	/* pangram with a whole alphabet set */
+	/* Pangram with a whole alphabet set */
 	sz = sizeof(pangram) - 1;
 	TEST_ASSERT_EQUAL_INT(sz, strlen(pangram));
 	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(pangram, sz - 1));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(pangram, sz));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(pangram, sz + 1));
 
-	/* text with null character in the middle*/
-	sz = (sizeof(torn) / 2) - 1;
+	/* Text with null character in the middle */
+	sz = sizeof(torn) / 2 - 1;
 	TEST_ASSERT_EQUAL_INT(sz, strlen(torn));
 	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(torn, 3 - 1));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(torn, 3));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(torn, 3 + 1));
 
+	/* End of string */
 	sz = 0;
-	/* end of string */
 	TEST_ASSERT_EQUAL_INT(sz, strlen(doubleNul));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(doubleNul, 0));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(doubleNul, 1));
@@ -77,21 +81,21 @@ TEST(string_len, ascii)
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(empty, 0));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(empty, 1));
 
-	/* special characters */
+	/* Special characters */
 	sz = sizeof(specials) - 1;
 	TEST_ASSERT_EQUAL_INT(sz, strlen(specials));
 	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(specials, sz - 1));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(specials, sz));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(specials, sz + 1));
 
-	/* white spaces */
+	/* White spaces */
 	sz = sizeof(whites) - 1;
 	TEST_ASSERT_EQUAL_INT(sz, strlen(whites));
 	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(whites, sz - 1));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(whites, sz));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(whites, sz + 1));
 
-	/*Checking ascii charset*/
+	/* Checking ascii charset */
 	for (i = 1; i < sizeof(asciiSet); i++) {
 		asciiSet[i - 1] = i;
 	}
@@ -117,7 +121,7 @@ TEST(string_len, not_ascii)
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(notAsciiString, sz));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(notAsciiString, sz + 1));
 
-	/* Checking  out of ASCII bytes */
+	/* Checking  out of ascii bytes */
 	for (i = 128; i <= 255; i++) {
 		notAsciiSet[i - 128] = i;
 	}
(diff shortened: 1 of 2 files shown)


libc/string_chr.c
```diff
@@ -0,0 +1,253 @@
...
+
+
+TEST_TEAR_DOWN(string_strchr)
+{
+}
+
+
+TEST(string_strchr, strchr_standard_use)
+{
+	char str[PATH_MAX] = "a";
```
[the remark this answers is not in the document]
maska989:
``` c
for (int i = 0; i < PATH_MAX - 5; i++) {

		if (i == PATH_MAX - 10) {
			strcat(str, "b");
		}
		else
			strcat(str, "a");
	}
```
That's why it is not const.

libc/string_chr.c
```diff
@@ -0,0 +1,253 @@
...
+	TEST_ASSERT_EQUAL_STRING(NULL, strchr(LOREM_IPSUM, 'x'));
+
+	for (int i = 0; i < PATH_MAX - 5; i++) {
+
+		if (i == PATH_MAX - 10) {
+			strcat(str, "b");
+		}
+		else
+			strcat(str, "a");
+	}
```
niewim19, line 60:
```suggestion
	memset(str, 'a', PATH_MAX - 5);
	str[PATH_MAX - 10] = 'b';
```

Much shorter. Though it needs `str` declaration as follows:
```c
char str[PATH_MAX] = { 0 };
```

Also:
 - `-5` and `-10` seems rather arbitrary. 
 - this test may be split into two: one with _loremipsum_ usage and other with `PATH_MAX` case. But that`s just a suggestion.
 - it is unclear why we use one type of string for forward search (`strchr`) but very different string for reverse search (strrchr).

libc/string_chr.c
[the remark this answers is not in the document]
damianloew: Yeah, that's right, but there were other cases.

libc/string_chr.c
niewim19: Or if you`re very eager on not using `memset` you can just:
```suggestion
	for (i = 0; i < PATH_MAX - 5; i++) {
		str[i] = 'a';
	}
	str[PATH_MAX - 10] = 'b';
```

libc/string_lenchr.c
```diff
@@ -150,9 +169,259 @@ TEST(string_len, big)
...
+		 *xwhites = " \v\t\r\nx";
+
+	int sz, i;
+	char str[INT8_MAX] = { 0 };
+	const char *format = "\399\45\43\0";
+
+	for (i = 1; i < sizeof(str); i++) {
+		str[i - 1] = i;
+	}
+	str[i - 1] = '\0';
```
damianloew, line 264: Not needed as str is initialized to zeros.

libc/string_lenchr.c
```diff
@@ -150,9 +169,259 @@ TEST(string_len, big)
...
+	TEST_ASSERT_EQUAL_STRING("x", strrchr(xwhites, 'x'));
+	TEST_ASSERT_EQUAL_STRING("x", memchr(xwhites, 'x', sz));
+
+	TEST_ASSERT_EQUAL_STRING(whites, strchr(whites, ' '));
+	TEST_ASSERT_EQUAL_STRING(whites, strrchr(whites, ' '));
+	TEST_ASSERT_EQUAL_STRING(whites, memchr(whites, ' ', strlen(whites)));
+
+	TEST_ASSERT_EQUAL_STRING(NULL, strchr(whites, 'x'));
+	TEST_ASSERT_EQUAL_STRING(NULL, strrchr(whites, 'x'));
+	TEST_ASSERT_EQUAL_STRING(NULL, memchr(whites, 'x', strlen(whites)));
```
damianloew, line 322: We decided to split test cases to smaller ones in `chr` tests, so please create a separate test case `whitespacees` and place the mentioned code there. Above that I know that I suggested you the solution with `xwhites`, but now I see that checking \0 as the last character would be sufficient, so please take the `" \v\t\r\n"` string and est chr functions for space, termination char, some char in the middle, for example \t and x which is not present in the mentioned string.

```suggestion
```

libc/string_lenchr.c
```diff
@@ -150,9 +169,259 @@ TEST(string_len, big)
...
+	for (i = 128; i <= 255; i++) {
+		notAsciiSet[i - 128] = i;
+	}
+	notAsciiSet[128] = 0;
+
+	/* Testing capability of functions to hold and read not ASCII set */
+	sz = sizeof(notAsciiSet) - 1;
+	TEST_ASSERT_EQUAL_STRING(notAsciiSet, strchr(notAsciiSet, notAsciiSet[0]));
+	TEST_ASSERT_EQUAL_STRING(notAsciiSet, strrchr(notAsciiSet, notAsciiSet[0]));
+	TEST_ASSERT_EQUAL_STRING(notAsciiSet, memchr(notAsciiSet, notAsciiSet[0], sz));
```
damianloew, line 348: We can also check notAsciiSet[128] and notAsciiSet[64].

libc/string_lenchr.c
```diff
@@ -150,9 +169,259 @@ TEST(string_len, big)
...
+
+
+TEST(string_chr, empty)
+{
+	char *empty = "";
+	int sz;
+
+	/* Checking if we can get an empty string */
+	sz = 1;
+	TEST_ASSERT_EQUAL_STRING("", strchr(empty, '\0'));
```
damianloew, line 359: In my opinion it's not needed to declare empty variable, we can just pass "" everywhere.

```suggestion
	TEST_ASSERT_EQUAL_STRING("", strchr("", '\0'));
```

libc/string_lenchr.c
```diff
@@ -150,9 +169,259 @@ TEST(string_len, big)
...
+	TEST_ASSERT_EQUAL_STRING(NULL, strrchr(empty, '\v'));
+	TEST_ASSERT_EQUAL_STRING(NULL, memchr(empty, '\v', sz));
+}
+
+
+TEST(string_chr, memchr_size)
+{
+	char lorem[] = "Lorem";
+
+	/* Testing capability of size in memchr with different sizes of reads */
```
damianloew, line 394:
```suggestion
	/* Testing capability of setting size in memchr */
```

libc/string_lenchr.c
```diff
@@ -150,9 +169,259 @@ TEST(string_len, big)
...
+	TEST_ASSERT_EQUAL_STRING(exp, strrchr(strWhites, 'I'));
+	TEST_ASSERT_EQUAL_STRING(exp, memchr(strWhites, 'I', strlen(strWhites)));
+
+	TEST_ASSERT_EQUAL_STRING(exp, strchr(strMod, 'I'));
+	TEST_ASSERT_EQUAL_STRING(exp, strrchr(strMod, 'I'));
+	TEST_ASSERT_EQUAL_STRING(exp, memchr(strMod, 'I', strlen(strMod)));
+}
+
+
+TEST(string_chr, ascii_cases)
```
damianloew, line 250:
```suggestion
TEST(string_chr, ascii)
```

libc/string_lenchr.c
```diff
@@ -150,9 +169,259 @@ TEST(string_len, big)
...
+{
+	char str[PATH_MAX] = { 0 };
+
+	/* Long string case */
+	memset(str, 'a', PATH_MAX - 1);
+	str[PATH_MAX - 5] = 'b';
+
+	TEST_ASSERT_EQUAL_STRING("baaa", strchr(str, 'b'));
+	TEST_ASSERT_EQUAL_STRING("baaa", strrchr(str, 'b'));
+	TEST_ASSERT_EQUAL_STRING("baaa", memchr(str, 'b', sizeof(str)));
```
damianloew, line 224: You can add there checks similar to the ones from the previous test cases, so above that you can add checking:

- the termination character
- the first character
- the character, which is not present in the string

damianloew (changes requested):
A few changes required to simplify this tests and provide the better coverage.

libc/string_lenchr.c
```diff
@@ -150,9 +154,291 @@ TEST(string_len, big)
...
+	TEST_ASSERT_EQUAL_STRING(NULL, strrchr(str, 'x'));
+	TEST_ASSERT_EQUAL_STRING(NULL, memchr(str, 'x', sizeof(str)));
+}
+
+
+TEST(string_chr, special)
+{
+	const char *exp = "Ipsum",
+			   *strWhites = "Lorem \n \t \e \r \b \v \f \\ Ipsum",
+			   *strMod = "Lorem %d/%f/%x/%i/%c/%s Ipsum";
```
damianloew, line 224: (strWhites) You have now the separate test case for white spaces, so it's not needed there. Above that special characters do not include white spaces, so it's not the place for it, we want to test only special characters there.

(strMod) Modifiers are nothing more than a combination of a % (one of the special characters, which are tested in this test case) and a letter (tested in the `basic` test case). That's why there is no need to test `strMod`. Modifiers are specific for functions like `scanf` or `printf` and then the behavior for different modifiers should be tested.

```suggestion
```

libc/string_lenchr.c
```diff
@@ -150,9 +154,291 @@ TEST(string_len, big)
...
+
+	TEST_ASSERT_EQUAL_STRING(NULL, strchr(torn, 'b'));
+	TEST_ASSERT_EQUAL_STRING(NULL, strrchr(torn, 'b'));
+	TEST_ASSERT_EQUAL_STRING(NULL, memchr(torn, 'b', sz));
+}
+
+
+TEST(string_chr, whitespaces)
+{
+	const char whites[] = " \v\t\r\n";
```
damianloew, line 345: What about `\f` and other white spaces?

libc/string_lenchr.c
```diff
@@ -150,9 +154,291 @@ TEST(string_len, big)
...
+	TEST_ASSERT_EQUAL_STRING(NULL, strrchr("", '\t'));
+	TEST_ASSERT_EQUAL_STRING(NULL, memchr("", '\t', sz));
+
+	TEST_ASSERT_EQUAL_STRING(NULL, strchr("", '\r'));
+	TEST_ASSERT_EQUAL_STRING(NULL, strrchr("", '\r'));
+	TEST_ASSERT_EQUAL_STRING(NULL, memchr("", '\r', sz));
+
+	TEST_ASSERT_EQUAL_STRING(NULL, strchr("", '\v'));
+	TEST_ASSERT_EQUAL_STRING(NULL, strrchr("", '\v'));
+	TEST_ASSERT_EQUAL_STRING(NULL, memchr("", '\v', sz));
```
damianloew, line 401: Not necessarily needed, we have the edge cases like first character (and last in the same time - \0) and the character which does not exist in the string (space)

```suggestion
```

libc/string_lenchr.c
```diff
@@ -150,9 +154,291 @@ TEST(string_len, big)
...
+	/* Checking if we can get an empty string */
+	sz = 1;
+	TEST_ASSERT_EQUAL_STRING("", strchr("", '\0'));
+	TEST_ASSERT_EQUAL_STRING("", strrchr("", '\0'));
+	TEST_ASSERT_EQUAL_STRING("", memchr("", '\0', sz));
+
+	TEST_ASSERT_EQUAL_STRING("", strchr("", 0));
+	TEST_ASSERT_EQUAL_STRING("", strrchr("", 0));
+	TEST_ASSERT_EQUAL_STRING("", memchr("", 0, sz));
+
```
damianloew, line 381: \0 and 0 are the same character

```suggestion
	TEST_ASSERT_EQUAL_STRING("", strchr("", '\0'));
	TEST_ASSERT_EQUAL_STRING("", strrchr("", '\0'));
	TEST_ASSERT_EQUAL_STRING("", memchr("", '\0', sz));
```

libc/string_lenchr.c
```diff
@@ -150,9 +154,291 @@ TEST(string_len, big)
...
+TEST_GROUP_RUNNER(string_chr)
+{
+	RUN_TEST_CASE(string_chr, basic);
+	RUN_TEST_CASE(string_chr, big);
+	RUN_TEST_CASE(string_chr, special);
+	RUN_TEST_CASE(string_chr, ascii);
+	RUN_TEST_CASE(string_chr, not_ascii);
+	RUN_TEST_CASE(string_chr, torn);
+	RUN_TEST_CASE(string_chr, whitespaces);
```
damianloew, line 441: Please move the whitespaces case between `special` and `ascii`. Now we have tested whitespaces in the `ascii` case, so it's not so valuable. We want to test some basic stuff, special characters and whites in the separate cases before the whole ascii set to eventually catch smaller issues (for example if handling white spaces didn't work we want to see error of `whitespaces` test case, not `ascii`, which would also fail)

damianloew (changes requested):
Thanks for changes, as you can see it looks much cleaner now :) Just a few suggestions left

damianloew (commented):
Seems to be fine :+1: Please squash the changes to one commit accroding to our convention, so it would be `libc: string.h: add chr functions tests` (and of course the appropriate JIRA tag below)

damianloew (approved):

PR #174, opened 2023-03-13
Maska989/test cpy

maska989:
## Description
Added 
 *  memcpy,
 *  memccpy,
 *  strncpy,
 *  stpncpy,
 *  strcpy,
 *  stpcpy
 
function tests and merged file with tests of strlcpy.

## Motivation and Context
JIRA: CI-213

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


libc/string_cpy.c
```diff
@@ -0,0 +1,752 @@
...
+#endif
+}
+
+
+/* 
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
+ */
+
+
+TEST_SETUP(string_strncpy)
```
damianloew, line 260: Please apply the similar changes and split tests to more test groups in `strncpy` and `spncpy`. Please remember about the functionality you want to test in the written test group/test case. For example test group `string_strncpy, copy_paste` implies that copy paste utility will be tested, but in fact it's too general, we want to test for example for those functions above some basic stuff I mentioned above we would like to test:

Based on this requirement:
`(bytes that follow a NUL character are not copied)` - POSIX strncpy documentation

 - capability of copying the NUL terminating character: test case could be named `nul_term_copy`
 - not copying the character after NUL terminatin character: `nul_term_stop`

Based on this piece of doc:
`If the array pointed to by s2 is a string that is shorter than n bytes, NUL characters shall be appended to the copy in the array pointed to by s1, until n bytes in all are written`
 
we could test:

 - capability of appending NUL characters: test case named `append_nul`

and of course based on this line:
`The stpncpy() and strncpy() functions shall copy not more than n bytes ... from the array pointed to by s2 to the array pointed to by s1`

we could check whether:

- not more characters are copied, a test case called `no_more`
- different packages of data (various sizes, values) can be copied, so we could create cases like `basic` , `big`, `every_byte`, `one_byte`, `zero` and so on

It's only example, you don't have to follow all of those names and ideas, but think about it, what utility you want to test in this specific test case. Of course we sometimes do a few things in one test case, but then a name shall indicate it.

damianloew (changes requested):
In general those tests are not bad, a lot of utilities are tested, but it's a little bit mixed. You can look through suggestions from memmove tests PR, apply those mentioned below and then please request for a review once again :)

libc/string_cpy.c
```diff
@@ -0,0 +1,1269 @@
...
+
+static char long_str[PATH_MAX];
+
+
+static int big_string(void)
+{
+	for (int i = 1; i < PATH_MAX; i++) {
+		long_str[i - 1] = i % MAX_ASCII_SET;
+	}
+	return 1;
```
damianloew, line 62: It's a common practice to return 0 as success
```suggestion
	return 0;
```

libc/string_cpy.c
```diff
@@ -0,0 +1,1269 @@
...
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(strText, input, sizeof(strText));
+}
+
+
+TEST(string_memcpy, big)
+{
+	char buff[1024] = { 0 };
+
+	/* Checking capability of handling big blocks of data */
+	TEST_ASSERT_EQUAL_INT(1, big_string());
```
damianloew, line 221: You could call it once for each test group, for example in memccpy tests it's called a few times.

libc/string_cpy.c
```diff
@@ -0,0 +1,1269 @@
...
+
+	char buff[BIG_NUMB] = { 0 };
+
+	/* Checking capability of handling big blocks of data */
+	TEST_ASSERT_EQUAL_INT(1, big_string());
+	memccpy(buff, long_str, '\xFF', BIG_NUMB);
+	TEST_ASSERT_EQUAL_STRING(buff, long_str);
+	TEST_ASSERT_EQUAL_CHAR(buff[0], '\1');
+	TEST_ASSERT_EQUAL_CHAR(buff[MAX_ASCII_SET / 4], 'A');
+	TEST_ASSERT_EQUAL_CHAR(buff[MAX_ASCII_SET / 2], '\x81');
```
damianloew, line 493: Why are there those checks if it was assertet that buff equals long_str?
```suggestion
```

libc/string_cpy.c
```diff
@@ -0,0 +1,1269 @@
...
+
+	/* Offset big chunk of data */
+	TEST_ASSERT_NOT_NULL(memccpy(&buff[128], long_str, '\xFF', BIG_NUMB - MAX_ASCII_SET));
```
damianloew, line 496: Why `BIG_NUMB - MAX_ASCII_SET`, so 768?

libc/string_cpy.c
```diff
@@ -0,0 +1,1269 @@
...
+	TEST_ASSERT_EQUAL_CHAR(buff[0], '\1');
```
damianloew, line 497: You could simply check the whole content of buff using 3x `TEST_EQUAL_CHAR_ARRAY` - first 128 chars {1..127}, next 128 chars {1..127} and the rest {0, 0, ....}

libc/string_cpy.c
```diff
@@ -0,0 +1,1269 @@
...
+	TEST_ASSERT_EQUAL_PTR(NULL, memccpy(strDest, TEST_STR3, 'x', sizeof(TEST_STR3)));
+	TEST_ASSERT_EQUAL_STRING(TEST_STR3, strDest);
+	TEST_ASSERT_NOT_EQUAL_CHAR(strDest[0], TEST_STR1[0]);
+	TEST_ASSERT_NOT_EQUAL_INT(0, strcmp(strDest, TEST_STR1));
+
+	/* Checking ability to copy one char */
+	TEST_ASSERT_EQUAL_PTR(NULL, memccpy(&chrDest, &chr, 'x', sizeof(chr)));
+	TEST_ASSERT_NOT_EMPTY(&chrDest);
+	TEST_ASSERT_EQUAL_CHAR(chr, chrDest);
+
```
damianloew, line 289: You could also create a simple check, which will copy less bytes than available in the buffer.

libc/string_cpy.c
```diff
@@ -0,0 +1,1269 @@
...
+	char buff[MAX_ASCII_SET] = { 0 },
+		 ascii[MAX_ASCII_SET] = { 0 };
+	int i;
+
+	for (i = 1; i < MAX_ASCII_SET; i++) {
+		ascii[i - 1] = i;
+	}
+
+	/* This case test not only ascii set copied but different sizes of copy */
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(ascii, strncpy(buff, ascii, i), i);
```
damianloew, line 574: You should also check whether buff equals ascii (return value could not be buff)

libc/string_cpy.c
```diff
@@ -0,0 +1,1269 @@
...
+
+	/* Checking if strncpy correctly read the string and if it is sensitivity on Null terminating zero */
+	TEST_ASSERT_EQUAL_STRING(testStr, strncpy(buff, testStr, sizeof(buff)));
+	TEST_ASSERT_NOT_EQUAL_CHAR(testStr[sizeof(testStr) / 2], buff[sizeof(testStr) / 2]);
+
+	memset(buff, 0, sizeof(buff));
+	memset(testStr, 0, sizeof(testStr));
+
+	/* Checking if we can copy place where NULL element is present */
+	TEST_ASSERT_EQUAL_PTR(&buff[0], strncpy(buff, "", sizeof(buff)));
```
damianloew, line 591:
```suggestion
	TEST_ASSERT_EQUAL_PTR(buff, strncpy(buff, "", sizeof(buff)));
```

libc/string_cpy.c
```diff
@@ -0,0 +1,1269 @@
...
+}
+
+
+TEST(string_strncpy, basic)
+{
+	int num = 12345678;
+	char payload[MAX_STR_LEN] = { 0 },
+		 buff[MAX_STR_LEN] = { 0 };
+
+	TEST_ASSERT_EQUAL_STRING(TEST_STR1, strncpy(buff, TEST_STR1, sizeof(buff)));
```
damianloew, line 530: I think it would be better to set the the number of bytes intended to copy as `n`, so in this case `sizeof(TEST_STR1) - 1` by default and create a separate case, where we will check the ability to copy only bytes before a nul teminatin character.

damianloew (changes requested):
It looks much better now, I would like to change a few things more before change. Not seen at `stpcpy`, `strlcpy`, `stpncpy` yet, but please apply the mentioned suggestions in all similar fragments of code in this file.

damianloew (changes requested):
It looks much better now, I would like to change a few things more before change. Not seen at `stpcpy`, `strlcpy`, `stpncpy` yet, but please apply the mentioned suggestions in all similar fragments of code in this file.

libc/string_cpy.c
maska989: After applying those changes and running the system on host-generic-pc this warning occurred:
Screenshot from 2023-05-12 11-58-14

libc/string_cpy.c
```diff
@@ -0,0 +1,1382 @@
...
+	 * Memccpy will return a byte after found endpoint
+	 * in case of find a smaller dataset it will always point to on next byte in the array
+	 * In this case we use long_str and try to find all chars between [1 and 255]
+	 * and check if the pointer that memccpy return is the same as the point where value should be
+	 */
+	for (i = 0; i < MAX_ASCII_SET; i++) {
+		TEST_ASSERT_EQUAL_PTR(&bigStrDest[i + 1], memccpy(bigStrDest, long_str, long_str[i], MAX_ASCII_SET));
+	}
+
+	/* Search range out of bounds */
```
damianloew, line 314: I think it's a check for an another test case, we could also check all the possible values, which are not present in the string - the test case could be named `stop_char_not_found`,  Above that you could test various `n` values - so you can extend the `zero` case - rename it to `various_sizes` and add the following checks:

 - `n` that equals 0
 - `n` that is lower than stop character position,
 - `n` that is greater than stop character position
 - `n` that is greater than stop character position and has the maximum possible value

libc/string_cpy.c
```diff
@@ -0,0 +1,1382 @@
...
+	TEST_ASSERT_NULL(memccpy(bigStrDest, long_str, '\xFF', 254));
+	TEST_ASSERT_EQUAL_CHAR(bigStrDest[253], '\xFE');
+
+#endif
+}
+
+
+TEST(string_memccpy, stop_cpy_point_int)
```
damianloew, line 322: Please apply the suggestions from the previous test case: `stop_cpy_point_char`

libc/string_cpy.c
```diff
@@ -0,0 +1,1382 @@
...
+
+TEST(string_memccpy, one_byte)
+{
+#ifdef __phoenix__
+	TEST_IGNORE();
+#endif
+#ifndef __phoenix__
+
+	char testArray[MAX_ASCII_SET] = { 0 },
+		 input[MAX_ASCII_SET] = { 1 };
```
damianloew, line 434: It may be misleading as you had an occasion to experience, this way it is clear that the whole buffer is set to zeros:

```suggestion
		 input[MAX_ASCII_SET] = { 0 };
```

libc/string_cpy.c
```diff
@@ -0,0 +1,1382 @@
...
+	TEST_ASSERT_EQUAL_PTR(testArray, memcpy(testArray, input, MAX_STR_LEN));
+	TEST_ASSERT_NOT_EQUAL_INT(0, strcmp(testArray, TEST_STR1));
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(input, testArray, MAX_STR_LEN);
+	TEST_ASSERT_EQUAL_INT(sizeof(input), sizeof(testArray));
+
+#endif
+}
+
+
+TEST(string_memccpy, zero_size)
```
damianloew, line 472: Please see the suggestion regarding this case above.

libc/string_cpy.c
```diff
@@ -0,0 +1,1382 @@
...
+
+	char buff[MAX_ASCII_SET] = { 0 },
+		 ascii[MAX_ASCII_SET] = { 0 };
+	int i;
+
+	for (i = 1; i < MAX_ASCII_SET; i++) {
+		ascii[i - 1] = i;
+
+		/* This case test not only ascii set copied but different sizes of copy */
+		TEST_ASSERT_EQUAL_CHAR_ARRAY(ascii, strncpy(buff, ascii, i), i);
```
damianloew, line 588: You don't check the content of `buff` each time, but maybe the check after for loop will be sufficient - we make sure that the `buff` is also a returned value, which is asserted in every iteration.

libc/string_cpy.c
```diff
@@ -0,0 +1,1382 @@
...
+	}
+
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(buff, ascii, sizeof(buff));
+
+#pragma GCC diagnostic pop
+}
+
+
+TEST(string_strncpy, append_null)
```
damianloew, line 597: It does not check the ability of appending NUL characters, but the ability of finding it. Please create a separate test case for testing append utility.

libc/string_cpy.c
```diff
@@ -0,0 +1,1382 @@
...
+{
+	char buff[MAX_ASCII_SET] = { 0 },
+		 ascii[MAX_ASCII_SET] = { 0 };
+	int i;
+
+	/* Copy one by one from created ascii table */
+	for (i = 1; i < MAX_ASCII_SET; i++) {
+		ascii[i - 1] = i;
+
+		TEST_ASSERT_EQUAL_CHAR_ARRAY(buff, strncpy(buff, &ascii[i], 1), sizeof(buff));
```
damianloew, line 647: We don't check the content of buff there.

libc/string_cpy.c
```diff
@@ -0,0 +1,1382 @@
...
+	TEST_ASSERT_EQUAL_STRING(bigBuff, strncpy(bigBuff, long_str, BIG_NUMB));
+	TEST_ASSERT_EQUAL_STRING(long_str, bigBuff);
+
+	/* Checking if a string doesn't contain any of the data after the size of input */
+	TEST_ASSERT_EQUAL_STRING(bigBuff, strncpy(bigBuff, payload, BIG_NUMB));
+	TEST_ASSERT_EQUAL_STRING(payload, bigBuff);
+	TEST_ASSERT_EQUAL_STRING(&bigBuff[BIG_NUMB / 2], "");
+	TEST_ASSERT_EQUAL_STRING(&bigBuff[BIG_NUMB / 3], "");
+	TEST_ASSERT_EQUAL_STRING(&bigBuff[BIG_NUMB / 4], "");
+	TEST_ASSERT_EQUAL_STRING(&bigBuff[BIG_NUMB / 5], "");
```
damianloew, line 674: That's checking the ability of appending NUL characters, but I would rather check all of the appended characters (and do it in the separate test case as mentioned somewhere above)

damianloew (changes requested):

libc/string_cpy.c
```diff
@@ -0,0 +1,1451 @@
...
+	TEST_ASSERT_EQUAL_STRING(bigBuff, strncpy(bigBuff, payload, BIG_NUMB));
+	TEST_ASSERT_EQUAL_STRING(payload, bigBuff);
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(bigBuff, payload, sizeof(bigBuff));
+
+	/* Input bigger than target */
+	memset(payload, 0, sizeof(payload));
+	memset(bigBuff, 'a', sizeof(bigBuff));
+
+	TEST_ASSERT_EQUAL_PTR(payload, strncpy(payload, bigBuff, sizeof(payload)));
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(payload, &bigBuff[sizeof(bigBuff) - sizeof(payload)], sizeof(payload));
```
damianloew, line 696: In the first check in this test case `TEST_ASSERT_EQUAL_STRING(bigBuff, strncpy(bigBuff, longStr, BIG_NUMB));` input is also bigger than target, so I am not sure whether the lines before are necessary.

libc/string_cpy.c
```diff
@@ -0,0 +1,1451 @@
...
+/* defines for string_strlcat */
+#define STR_SRC1        "abc"
+#define STR_SRC2        "defgh"
+#define STR_PLACEHOLDER "klmnopqrstu"
+
+
+static char longStr[BIG_NUMB];
+
+
+static int big_string(void)
```
damianloew, line 53: There is a common convention to indicate what the function is doing in a name.

```suggestion
static int create_big_string(void)
```

damianloew, line 53: Above that you can change the return type to void as you always return `0` and don't assert this value.

libc/string_cpy.c
```diff
@@ -0,0 +1,1451 @@
...
+		}
+		else {
+			input[i] = longStr[i];
+		}
+	}
+
+	/* stpncpy will search for end of array by looking for nul term */
+	input[i] = '\0';
+
+	TEST_ASSERT_EQUAL_PTR(buff, strncpy(buff, input, sizeof(buff)));
```
damianloew, line 721: You don't need buff there, you can use the second half of longStr:

```suggestion
	TEST_ASSERT_EQUAL_PTR(buff, strncpy(buff, &longStr[BIG_NUMB / 2], BIG_NUMB / 2));
```

libc/string_cpy.c
```diff
@@ -0,0 +1,1451 @@
...
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(buff, input, sizeof(input) - 1);
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(&buff[sizeof(input)], zeroHolder, sizeof(input));
```
damianloew, line 723: You don't need so big buffer with zeros, you can assert each element separately in for loop.

libc/string_cpy.c
```diff
@@ -0,0 +1,1451 @@
...
+	TEST_ASSERT_EQUAL_PTR(payload, strncpy(payload, bigBuff, sizeof(payload)));
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(payload, &bigBuff[sizeof(bigBuff) - sizeof(payload)], sizeof(payload));
+
+#pragma GCC diagnostic pop
+}
+
+
+TEST(string_strncpy, append_null)
+{
+	char input[BIG_NUMB / 2] = { 0 }, zeroHolder[BIG_NUMB / 2] = { 0 }, buff[BIG_NUMB] = { 0 };
```
damianloew, line 704: If in `buff` are already zeros, you don't know exactly whether the nul characters were appended or not.

libc/string_cpy.c
```diff
@@ -0,0 +1,1451 @@
...
+{
+}
+
+
+TEST_TEAR_DOWN(string_stpncpy)
+{
+}
+
+
+TEST(string_stpncpy, basic)
```
damianloew, line 742: Not looked at those tests in detail yet, please apply all the suggestions also in those tests and I will verify that during the next review.

damianloew (changes requested):

libc/string_cpy.c
maska989: `input` and `sizeof(buff)` are used because input holds only chars that are not null terms and buff size is 2 times bigger than the input size

libc/string_cpy.c
```diff
@@ -746,19 +783,25 @@ TEST(string_stpncpy, basic)
...
+	char payload[MAX_STR_LEN] = { '1', '2', '3', '4', '5', '6', '7', '8' },
 		 buff[MAX_STR_LEN] = { 0 };
+	int i;
 
 	TEST_ASSERT_EQUAL_PTR(&buff[strlen(buff)], stpncpy(buff, TEST_STR1, sizeof(buff)));
 	TEST_ASSERT_EQUAL_STRING(TEST_STR1, buff);
+
 	/* Checking if we doesn't overwrite elements after end of input*/
-	TEST_ASSERT_EQUAL_CHAR_ARRAY(&buff[strlen(TEST_STR1)], "\0\0\0\0\0\0", sizeof(buff) - sizeof(TEST_STR1));
+	for (i = strlen(buff); i < sizeof(buff) - strlen(buff); i++) {
```
damianloew, line 794: `i < sizeof(buff) - strlen(buff)`?

libc/string_cpy.c
```diff
@@ -901,18 +943,43 @@ TEST(string_stpncpy, one_byte)
 }
 
 
+TEST(string_stpncpy, various_sizes)
```
damianloew, line 946: Is it possible to pass a size bigger than the source, or some other edge values there? If yes please add it

libc/string_cpy.c
```diff
@@ -0,0 +1,1527 @@
...
+	/* stpncpy will search for end of array by looking for nul term */
+	input[i] = '\0';
+
+	/* To check append ability we need to setup values in buff to other than 0 */
+	memset(buff, 1, sizeof(buff));
+
+	TEST_ASSERT_EQUAL_PTR(&buff[strlen(buff)], stpncpy(buff, input, sizeof(buff)));
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(buff, input, sizeof(input) - 1);
+
+	for (i = BIG_NUMB / 2 - 1; i < BIG_NUMB / 2; i++) {
```
damianloew, line 1034: `i < BIG_NUMB / 2`?

libc/string_cpy.c
```diff
@@ -0,0 +1,1527 @@
...
+
+TEST_TEAR_DOWN(string_stpncpy)
+{
+}
+
+
+TEST(string_stpncpy, basic)
+{
+#ifdef __phoenix__
+	TEST_IGNORE();
```
damianloew, line 782: Why are those tests ignored? Missing comment.

damianloew (changes requested):

libc/string_cpy.c
```diff
@@ -0,0 +1,1527 @@
...
+#endif
+#ifndef __phoenix__
+
+	char bigStrDest[BIG_NUMB] = { 0 };
+	/*
+	 * Memccpy accepts int as an element of stop and converts it to uint8
+	 * We use -1 and +1 to MAX_ASCII_SET to not go over with values and stick to 255 and -255
+	 */
+	int bigIntDest[] = { 1, -MAX_ASCII_SET + 1, MAX_ASCII_SET - 1, INT_MAX / 5, INT_MAX / 3, INT_MAX, INT_MIN / 5, INT_MIN / 3, INT_MIN + 1 };
```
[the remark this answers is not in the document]
maska989: Our variable longStr itself has null terms. Every time the number from `bigIntDest` converted to `uint8` equals `0` we will always acquire a spot where the first null term is present, in our case it will be a place defined by `MAX_ASCII_SET`. Created case in case of this situation has a place.

A solution in case of this situation takes place will be created and the values of `bigIntDest` could be chosen freely.

libc/string_cpy.c
```diff
@@ -0,0 +1,1482 @@
...
+	/* Pragma truncation disable for the length of string tests */
+
+	/* Ability to copy long string */
+	TEST_ASSERT_EQUAL_PTR(bigBuff, strncpy(bigBuff, longStr, sizeof(bigBuff)));
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(bigBuff, longStr, sizeof(bigBuff));
+
+	free((void *)longStr);
+
+#pragma GCC diagnostic pop
+}
```
damianloew, line 694: We can avoid disabling this warning by passing a length of string (not counting NUL term) instead of its size as we have the following note in doc: `bytes that follow a NUL character are not copied`

damianloew (changes requested):
I hope that thanks to those changes we can avoid most of the pragmas (I know that you already removed many of them, so thanks for that). Above those suggestions please split commit `libc: string.h: move tests from string.c and add new cpy test cases` to `libc: move tests from string.c to string_cpy.c` and `libc: string: add new cpy tests` (and list what functions are tested below).

damianloew (changes requested):
Only one thing more

damianloew (approved):
Thanks for all the changes. I would like to merge those tests, as there are a lot of functional tests with various operations on strings. I think probably it will be better to test `memcpy()` function as it was done in recently merged memmove/memset tests (taking care of modulo 8), but for let's leave this version (of course for the rest of functions tested there, excluding memccpy this approach is IMHO better)

PR #177, opened 2023-03-16
Mateusz bloch/test malloc

mateusz-bloch:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


libc/stdlib_alloc.c
```diff
@@ -0,0 +1,277 @@
...
+	TEST_ASSERT_NOT_NULL(ptr);
+
+	int *ptr1 = realloc(ptr, 0);
+	TEST_ASSERT_NULL(ptr1);
+}
+
+TEST(stdlib_malloc, realloc_resize)
+{
+	int NUMBERS[BLOCK_SIZE];
+	int *ptr = (int *)calloc(BLOCK_SIZE, sizeof(int));
```
jsarzy, line 181: Add assert if ptr is not NULL and if content of ptr is zeroed

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,277 @@
...
+	int *ptr1 = (int *)realloc(ptr, sizeof(int));
+	TEST_ASSERT_NOT_NULL(ptr1);
+	TEST_ASSERT_EQUAL_INT(1, ptr1[0]);
+
+	free(ptr1);
+}
+
+TEST(stdlib_malloc, realloc_resize_large)
+{
+	int *ptr = (int *)calloc(BLOCK_SIZE, 2 * sizeof(int));
```
jsarzy, line 209: Add check for NULL and zeros. This function do not test calloc but it doesn't hurt us to do additional checks.

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,277 @@
...
+TEST(stdlib_malloc, realloc_multiple)
+{
+	int *ptr = calloc(1, BLOCK_SIZE);
+	TEST_ASSERT_NOT_NULL(ptr);
+
+	int *ptr1 = realloc(ptr, BLOCK_SIZE);
+	TEST_ASSERT_NOT_NULL(ptr1);
+	ptr = NULL;
+	TEST_ASSERT_NULL(ptr);
+	TEST_ASSERT_TRUE(ptr != ptr1);
```
jsarzy, line 226: There is no need for these checks. We made them already - `ptr1` is not NULL and we set `ptr` to NULL.
Applies also below

jsarzy (changes requested):
Nice to have these tests, I pointed few things that I noticed.

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,277 @@
...
+TEST_TEAR_DOWN(stdlib_malloc)
+{
+}
+
+TEST(stdlib_malloc, malloc_basic)
+{
+	void *ptr;
+	ptr = malloc(BLOCK_SIZE);
+	TEST_ASSERT_NOT_NULL(ptr);
+	TEST_ASSERT_TRUE(ptr != 0);
```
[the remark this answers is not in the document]
nalajcie: Yeah, this is exactly the same - p is of pointer type so there is implicit conversion to `(char*) 0` which is the `NULL` definition by C99 standard (see https://stackoverflow.com/a/9894047 and https://c-faq.com/null/ for more details)

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,318 @@
...
+	TEST_ASSERT_EACH_EQUAL_HEX8(0xb4, ptr, BLOCK_SIZE * LARGE_SIZE);
+	free(ptr);
+}
+
+TEST(stdlib_malloc, malloc_multiple)
+{
+	void *ptr0;
+	ptr0 = malloc(sizeof(int));
+	TEST_ASSERT_NOT_NULL(ptr0);
+	TEST_ASSERT_TRUE((uintptr_t)ptr0 % sizeof(int) == 0);
```
jsarzy, line 88: I suggest using `TEST_ASSERT_EQ` (or EQUAL if EQ doesn't exist)
```suggestion
	TEST_ASSERT_EQ((uintptr_t)ptr0 % sizeof(int),  0);
```

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,318 @@
...
+	TEST_ASSERT_EQUAL_INT(1, ptr1[0]);
+	TEST_ASSERT_TRUE((uintptr_t)ptr1 % sizeof(int) == 0);
+
+	free(ptr1);
+}
+
+TEST(stdlib_malloc, realloc_resize_large)
+{
+	int NUMBERS[BLOCK_SIZE];
+	int *ptr = (int *)calloc(BLOCK_SIZE, 2 * sizeof(int));
```
jsarzy, line 240: Be consistent with casting - in one part of test cases you are casting the pointer returned by malloc/calloc and in other part not. I would omit casting as is not needed.

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,318 @@
...
+	TEST_ASSERT_NOT_NULL(ptr1);
+	TEST_ASSERT_EQUAL_INT_ARRAY(NUMBERS, ptr1, BLOCK_SIZE);
+	TEST_ASSERT_TRUE((uintptr_t)ptr1 % sizeof(int) == 0);
+
+	free(ptr1);
+}
+
+TEST(stdlib_malloc, realloc_resize_small)
+{
+	int *ptr = (int *)calloc(1, 2 * sizeof(int));
```
jsarzy, line 224: I think this test should be made also with malloc function. We treat malloc/calloc as a black box and we do not do any assumptions how they work. Maybe realloc called after malloc doesn't work as good as combined with calloc ;).

So what I suggest: do these tests also with malloc and rename them to `realloc_[malloc/calloc]_resize_*`

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,318 @@
...
+	TEST_ASSERT_NULL(ptr0);
+#else
+	TEST_ASSERT_NOT_NULL(ptr0);
+#endif
+
+	free(ptr0);
+	free(ptr1);
+}
+
+TEST(stdlib_malloc, calloc_one)
```
jsarzy, line 164: We should also have a test that tries to alloc too much memory, e.g. `calloc(1, SIZE_MAX)`, `calloc(SIZE_MAX, SIZE_MAX)`, `calloc(SIZE_MAX, 1)` and check if errno is set to `ENOMEM`. It also applies to malloc and realloc.

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,318 @@
...
+TEST(stdlib_malloc, calloc_zero)
+{
+	int *ptr0 = calloc(0, sizeof(int));
+#ifdef phoenix
+	TEST_ASSERT_NULL(ptr0);
+#else
+	TEST_ASSERT_NOT_NULL(ptr0);
+#endif
+
+	int *ptr1 = calloc(BLOCK_SIZE, 0);
```
jsarzy, line 153: One case is missing `calloc(0, 0)`

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,318 @@
...
+
+TEST(stdlib_malloc, calloc_large)
+{
+	int *ptr = calloc(BLOCK_SIZE * LARGE_SIZE, sizeof(int));
+	TEST_ASSERT_NOT_NULL(ptr);
+	TEST_ASSERT_EACH_EQUAL_INT(0, ptr, BLOCK_SIZE * LARGE_SIZE);
+	free(ptr);
+}
+
+TEST(stdlib_malloc, realloc_null)
```
jsarzy, line 181: Add test case with `realloc(NULL, 0)`

libc/stdlib_alloc.c
jsarzy, line 240:
```suggestion
	int *ptr = (int *)calloc(2 * BLOCK_SIZE, sizeof(int));
```

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,318 @@
...
+
+	free(ptr1);
+}
+
+TEST(stdlib_malloc, realloc_resize_large)
+{
+	int NUMBERS[BLOCK_SIZE];
+	int *ptr = (int *)calloc(BLOCK_SIZE, 2 * sizeof(int));
+	TEST_ASSERT_NOT_NULL(ptr);
+	TEST_ASSERT_EACH_EQUAL_INT(0, ptr, BLOCK_SIZE);
```
jsarzy, line 242:
```suggestion
	TEST_ASSERT_EACH_EQUAL_INT(0, ptr, 2 * BLOCK_SIZE);
```

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,318 @@
...
+	for (int i = 0; i < BLOCK_SIZE; i++) {
+		ptr[i] = i;
+		NUMBERS[i] = i;
+	}
+	TEST_ASSERT_EQUAL_INT_ARRAY(NUMBERS, ptr, BLOCK_SIZE);
+
+	int *ptr1 = (int *)realloc(ptr, LARGE_SIZE * sizeof(int));
```
jsarzy, line 249:
```suggestion
	int *ptr1 = (int *)realloc(ptr, BLOCK_SIZE * sizeof(int));
```
Or maybe you mean `BLOCK_SIZE * LARGE_SIZE * sizeof(int)`?

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,318 @@
...
+
+	int *ptr1 = (int *)realloc(ptr, sizeof(int));
+	TEST_ASSERT_NOT_NULL(ptr1);
+	TEST_ASSERT_EQUAL_INT(1, ptr1[0]);
+	TEST_ASSERT_TRUE((uintptr_t)ptr1 % sizeof(int) == 0);
+
+	free(ptr1);
+}
+
+TEST(stdlib_malloc, realloc_resize_large)
```
jsarzy, line 237: This test do the same thing as `realloc_resize_small` but uses larger block sizes. Although, both of them shrinks allocated memory. I would like to see a test that is similar to this but memory is increased -> e.g. malloc(BLOCK_SIZE), realloc(ptr, BLOCK_SIZE*2).

To sum up:
I suggest having two test cases: `realloc_resize_smaller` and `realloc_resize_larger`.
In the first one we are resizing block from larger to smaller (e.g two cases: `2*sizeof(int)` -> `sizeof(int)` and `2*(big chunk of memory)` -> `(big chunk of memory)`.
The second one is similar but from smaller to larger.

jsarzy (changes requested):
Thanks for the changes, I pointed few missing test cases.

I think we should also have a test that iterates over sizes and tries to alloc them (something similar to snippet code)

```
for (size_t s = 1; s <= 4096; s++) {
    malloc(s); // for both malloc and calloc
    free(s);
}

File: libc/stdlib_alloc.c [hunk 14]
```diff
@@ -0,0 +1,561 @@
...
+
+	free(ptr1);
+}
+
+
+TEST(stdlib_malloc, realloc_calloc_resize_smaller)
+{
+	int *ptr = calloc(1, 2 * sizeof(int));
+	TEST_ASSERT_NOT_NULL(ptr);
+	TEST_ASSERT_EQUAL_INT(0, ptr[0]);
```
jsarzy, line 340: Check both of ptr[0] and ptr[1] so I suggest using EACH_EQUAL and maybe change to `calloc(2, sizeof(int))` to emphasize that this is a memory block of two ints.

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+	ptr[0] = 1;
+
+	int *ptr1 = realloc(ptr, sizeof(int));
+	TEST_ASSERT_NOT_NULL(ptr1);
+	TEST_ASSERT_EQUAL_INT(1, ptr1[0]);
+	TEST_ASSERT_EQUAL_INT(0, (uintptr_t)ptr1 % sizeof(int));
+
+	int *ptr2 = calloc(1, 2 * (BLOCK_SIZE * LARGE_SIZE));
+	TEST_ASSERT_NOT_NULL(ptr2);
+	TEST_ASSERT_EQUAL_INT(0, ptr2[0]);
```
jsarzy, line 350: ditto, you can use char* to avoid allocating too much memory

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+	ptr2[0] = 1;
```
jsarzy, line 351: Set the whole prefix of size `BLOCK_SIZE * LARGE_SIZE`, not the single value.

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+
+	int *ptr3 = realloc(ptr2, BLOCK_SIZE * LARGE_SIZE);
+	TEST_ASSERT_NOT_NULL(ptr3);
+	TEST_ASSERT_EQUAL_INT(1, ptr3[0]);
```
jsarzy, line 355: ditto

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+	TEST_ASSERT_NOT_NULL(ptr);
+	TEST_ASSERT_EQUAL_INT(0, *ptr);
+	ptr[0] = 1;
+
+	int *ptr1 = realloc(ptr, 2 * sizeof(int));
+	TEST_ASSERT_NOT_NULL(ptr1);
+	TEST_ASSERT_EQUAL_INT(0, (uintptr_t)ptr1 % sizeof(int));
+	TEST_ASSERT_EQUAL_INT(1, ptr1[0]);
+
+	int *ptr2 = calloc(BLOCK_SIZE / 8, BLOCK_SIZE * LARGE_SIZE);
```
jsarzy, line 375:
```suggestion
	int *ptr2 = calloc(BLOCK_SIZE * LARGE_SIZE, sizeof(int));
```

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+	TEST_ASSERT_NOT_NULL(ptr2);
+	TEST_ASSERT_EACH_EQUAL_INT(0, ptr2, BLOCK_SIZE);
```
jsarzy, line 377:
```suggestion
	TEST_ASSERT_EACH_EQUAL_INT(0, ptr2, BLOCK_SIZE * LARGE_SIZE);
```

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+	for (int i = 0; i < BLOCK_SIZE; i++) {
+		ptr2[i] = i;
+		NUMBERS[i] = i;
+	}
```
jsarzy, line 381: Set the whole prefix, I mean `BLOCK_SIZE * LARGE_SIZE`. You don't need to use additional `NUMBERS` array to make this check, just afterwards check if each  value is equal to 0, 1, 2, ...

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+	TEST_ASSERT_EQUAL_INT_ARRAY(NUMBERS, ptr2, BLOCK_SIZE);
+
+	int *ptr3 = realloc(ptr2, 2 * (BLOCK_SIZE * LARGE_SIZE));
```
jsarzy, line 384:
```suggestion
	int *ptr3 = realloc(ptr2, 2 * sizeof(int) * (BLOCK_SIZE * LARGE_SIZE));
```
If this is too much memory for some devices use `unsigned char*` to allocate ptr2

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+	TEST_ASSERT_NOT_NULL(ptr3);
+	TEST_ASSERT_EQUAL_INT_ARRAY(NUMBERS, ptr3, BLOCK_SIZE);
```
jsarzy, line 386:
```suggestion
	for (int i = 0; i < BLOCK_SIZE * LARGE_SIZE; i++) {
	    TEST_ASSERT_EQ(ptr[i], i)
	}
```
If you want to use `unsigned char*` type make sure to take modulo: `TEST_ASSERT_EQ(ptr[i], i % sizeof(unsigned char))`

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+	free(ptr1);
+	free(ptr3);
+}
+
+
+TEST(stdlib_malloc, realloc_multiple)
+{
+	int *ptr = calloc(BLOCK_SIZE, BLOCK_SIZE);
+	TEST_ASSERT_NOT_NULL(ptr);
+	TEST_ASSERT_EACH_EQUAL_INT(0, ptr, BLOCK_SIZE);
```
jsarzy, line 465: You only have checked `sizeof(int) * BLOCK_SIZE` bytes, we allocated `BLOCK_SIZE * BLOCK_SIZE` bytes. Fix the sizes or use char \*/unsigned char\*

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,561 @@
...
+	errno = 0;
+	void *ptr = malloc(BLOCK_SIZE);
+	TEST_ASSERT_NOT_NULL(ptr);
+	TEST_ASSERT_NULL(realloc(ptr, SIZE_MAX));
+	TEST_ASSERT_EQUAL_INT(ENOMEM, errno);
+
+	errno = 0;
+	int *ptr1 = calloc(1, BLOCK_SIZE);
+	TEST_ASSERT_NOT_NULL(ptr);
+	TEST_ASSERT_EQUAL_INT(0, *ptr1);
```
jsarzy, line 516: ditto

jsarzy (changes requested):
Thanks for the changes, I've found a few more errors. Let's fix them and I think we will be ready to merge :). Remember to squash the commits before the final merge

libc/stdlib_alloc.c
```diff
@@ -0,0 +1,572 @@
...
+	TEST_ASSERT_NOT_NULL(ptr3);
+	TEST_ASSERT_EACH_EQUAL_HEX8(0x80, ptr3, LARGE_SIZE);
+
+	free(ptr1);
+	free(ptr3);
+}
+
+
+TEST(stdlib_malloc, realloc_multiple)
+{  // to zrobić
```
jsarzy, line 473: remove comment

PR #178, opened 2023-03-20
add memset and memmove tests

adamdebek:
JIRA: CI-230

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


damianloew (changes requested):
I would also add the test cases named `large` (moving some large buffer) and `various_values` (moving a buffer that contains all possible bytes) - I don't have an idea for a better name for now.

libc/string_mem.c
```diff
@@ -0,0 +1,237 @@
...
+
+#include <string.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <ulimit.h>
+#include <stdint.h>
+#include <unity_fixture.h>
+
+#define LEN         (sizeof(int64_t) * 2)
```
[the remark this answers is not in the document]
adamdebek: I found this value optimal.

libc/string_mem.c
```diff
@@ -0,0 +1,373 @@
...
+}
+
+
+TEST_TEAR_DOWN(string_memmove_big)
+{
+	free(mem_big);
+}
+
+
+TEST(string_memmove_big, no_overlap_big1)
```
damianloew, line 259: You don't have to suffix test case names by `big` , as the test group is called this way. Please correct those cases like above. 

```suggestion
TEST(string_memmove_big, no_overlap_far)
```

damianloew (changes requested):
I would also add cases for moving very small buffer, for example 1 byte. It would be good to think about making some parts common, maybe creating some common macro, just not to copy 3 times.

libc/string_mem.c
```diff
@@ -0,0 +1,373 @@
...
+	TEST_ASSERT_EQUAL_PTR(ptr, ret);
+	TEST_ASSERT_EQUAL_MEMORY(TEST_STR, ptr, TEST_STR_SIZE);
+	TEST_ASSERT_EQUAL_MEMORY(TEST_STR, ((int8_t *)mem) + OFFSET, TEST_STR_SIZE);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, ((int8_t *)mem), OFFSET);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, ((int8_t *)mem) + OFFSET + TEST_STR_SIZE, move_offset - (OFFSET + TEST_STR_SIZE));
+	TEST_ASSERT_EACH_EQUAL_INT8(0, ((int8_t *)mem) + move_offset + TEST_STR_SIZE, MEM_SIZE - (move_offset + TEST_STR_SIZE));
+}
+
+
+TEST(string_memmove, no_overlap3)
```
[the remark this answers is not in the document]
adamdebek: I would rather call it no_overlap_adjacent

libc/string_mem.c
```diff
@@ -0,0 +1,404 @@
...
+/* next values exceed 1 byte, (int8_t)385 = -127, (int8_t)511 = -1 and so on */
+
+
+typedef struct over_write {
+	int32_t val1;
+	int32_t val2;
+} over_write_t;
+
+
+TEST_GROUP(string_memset);
```
damianloew, line 72: The new approach. Thanks to it we can see all test groups provided in this file.

```suggestion
TEST_GROUP(string_memset);
TEST_GROUP(string_memmove);
TEST_GROUP(string_memmove_big);
```

damianloew (changes requested):

libc/string_mem.c
```diff
@@ -0,0 +1,404 @@
...
+	TEST_ASSERT_EACH_EQUAL_INT8(0, ((int8_t *)mem), OFFSET);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, ((int8_t *)mem) + OFFSET + (TEST_STR_SIZE * 2), MEM_SIZE - (OFFSET + (TEST_STR_SIZE * 2)));
+}
+
+
+TEST(string_memmove, whole_overlap)
+{
+	int8_t *ptr = (int8_t *)mem + OFFSET;
+	/* Move at the same address as block is */
+	ret = memmove(ptr, TEST_STR, TEST_STR_SIZE);
```
nalajcie, line 198: :exclamation:  these above tests do not test overlap at all. 

You need to test overlapping of the **memory** not the **values**

`<ptr, ptr + 20>` memory and `<TEST_STR, TEST_STR + 20>` memory never overlaps.

What You need to test is for example `memmove(ptr, ptr + 10, 20)` and verify if the output memory is not garbled (in `memcpy` it might as the order of copying is not defined - that's the main difference

libc/string_mem.c
```diff
@@ -0,0 +1,404 @@
...
+	TEST_ASSERT_EQUAL_MEMORY(TEST_STR, ptr, TEST_STR_SIZE);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, ((int8_t *)mem), OFFSET);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, ((int8_t *)mem) + OFFSET + TEST_STR_SIZE, MEM_SIZE - (OFFSET + TEST_STR_SIZE));
+}
+
+
+TEST(string_memmove, move_byte)
+{
+	int8_t data = 0xFF;
+	int32_t buffer = 0x0000FF00;
```
nalajcie, line 247: invalid, assumes the machine is Little Endian

libc/string_mem.c
```diff
@@ -0,0 +1,404 @@
...
+	TEST_ASSERT_EACH_EQUAL_INT8(0, ((int8_t *)mem_big), OFFSET_BIG);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, ((int8_t *)mem_big) + OFFSET_BIG + (BIG_SIZE * 2), MEM_SIZE_BIG - (OFFSET_BIG + (BIG_SIZE * 2)));
+}
+
+
+TEST(string_memmove_big, whole_overlap)
+{
+	int8_t *ptr = (int8_t *)mem_big + OFFSET_BIG;
+	/* Move at the same address as block is */
+	ret = memmove(ptr, big_buf, BIG_SIZE);
```
nalajcie, line 329: the same as above, no overlap at all

nalajcie (changes requested):
Added some generic coding hints for increased readability and decreased memory usage. 
Unfortunately all memmove overlapping tests need to be reworked ;).

libc/string_mem.c
damianloew: Thanks for catching it, we will be more careful in future tests.

libc/string_mem.c
```diff
@@ -0,0 +1,396 @@
...
+	TEST_ASSERT_EQUAL_MEMORY(big_buf, mem_big + OFFSET_BIG, BIG_SIZE);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, mem_big + OFFSET_BIG + BIG_SIZE, (MEM_SIZE_BIG - BIG_SIZE) - (OFFSET_BIG + BIG_SIZE));
+	TEST_ASSERT_EACH_EQUAL_INT8(0, mem_big, OFFSET_BIG);
+}
+
+
+TEST(string_memmove_big, no_overlap_between)
+{
+	int move_offset = 2824;
+	uint8_t *ptr = mem_big + 2824;
```
damianloew:
```suggestion
	uint8_t *ptr = mem_big + move_offset;
```

damianloew (approved):
LGTM

libc/string_mem.c
```diff
@@ -0,0 +1,396 @@
...
+	TEST_ASSERT_EACH_EQUAL_INT8(0, mem + OFFSET + TEST_STR_SIZE, move_offset - (OFFSET + TEST_STR_SIZE));
+	TEST_ASSERT_EACH_EQUAL_INT8(0, mem + move_offset + TEST_STR_SIZE, MEM_SIZE - (move_offset + TEST_STR_SIZE));
+}
+
+
+TEST(string_memmove, no_overlap_adjacent)
+{
+	uint8_t *ptr = mem + OFFSET + TEST_STR_SIZE;
+	/* Move so that memory blocks are next to each other*/
+	ret = memmove(ptr, mem + OFFSET, TEST_STR_SIZE);
```
nalajcie, line 164: the readability could be improved by using the same variable as src and dst params (clear relation between src and dst without even reading the comment), eg:

```c
uint8_t *src = mem + OFFSET;
ret = memmove(src + TEST_STR_SIZE, src, TEST_STR_SIZE);
```
please use the same approach in other tests

libc/string_mem.c
```diff
@@ -0,0 +1,396 @@
...
+	TEST_ASSERT_EQUAL_PTR(ptr, ret);
+	TEST_ASSERT_EQUAL_MEMORY(TEST_STR, ptr, TEST_STR_SIZE);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, mem, OFFSET);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, mem + OFFSET + TEST_STR_SIZE, MEM_SIZE - (OFFSET + TEST_STR_SIZE));
+}
+
+
+TEST(string_memmove, move_byte)
+{
+#if __BYTE_ORDER == __LITTLE_ENDIAN
```
nalajcie, line 225: nah, no need for a test like that, If You reaaaly want to test if one-byte memmove works that way, use `uint8_t mem[4]`

SIde note: use EQUAL_UINT32 for unsigned type

libc/string_mem.c
```diff
@@ -0,0 +1,396 @@
...
+}
+
+
+TEST_SETUP(string_memmove)
+{
+	mem = (uint8_t *)calloc(MEM_SIZE, 1);
+	TEST_ASSERT_NOT_NULL(mem);
+	ret = memcpy(mem + OFFSET, TEST_STR, TEST_STR_SIZE);
+	TEST_ASSERT_EQUAL_PTR(mem + OFFSET, ret);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, mem, OFFSET);
```
nalajcie, line 121: use UINT8 everywhere (these are unsigned bytes)

libc/string_mem.c
```diff
@@ -0,0 +1,396 @@
...
+	TEST_ASSERT_EQUAL_MEMORY(big_buf, ptr, BIG_SIZE);
+	TEST_ASSERT_EQUAL_MEMORY(big_buf, mem_big + OFFSET_BIG, BIG_SIZE);
+	TEST_ASSERT_EACH_EQUAL_INT8(0, mem_big + OFFSET_BIG + BIG_SIZE, (MEM_SIZE_BIG - BIG_SIZE) - (OFFSET_BIG + BIG_SIZE));
+	TEST_ASSERT_EACH_EQUAL_INT8(0, mem_big, OFFSET_BIG);
+}
+
+
+TEST(string_memmove_big, no_overlap_between)
+{
+	int move_offset = 2824;
```
nalajcie, line 291: `const`

I don't see the relation to OFFSET_BIG and BIG_SIZE, maybe write the value using defined variables?

libc/string_mem.c
```diff
@@ -0,0 +1,396 @@
...
+	TEST_ASSERT_EQUAL_PTR(&val, ret);
+	TEST_ASSERT_EQUAL_INT8(5, val);
+}
+
+
+TEST(string_memset, big)
+{
+	ret = memset(big_buf, 10, BIG_SIZE);
+	TEST_ASSERT_EQUAL_PTR(big_buf, ret);
+	TEST_ASSERT_EQUAL_MEMORY(expect_mem_big, big_buf, BIG_SIZE);
```
nalajcie, line 91: use `EACH_EQUAL_UINT8` and remove `expect_mem_big`

**Important:**
the `memset` and `memcpy` tests should test lengths and starting pointers not dividable by 2, 4, 8  (so with `len % 8` and `ptr % 8` with all combinations) - common optimization is to use full words for memset / memcpy if possible

libc/string_mem.c
```diff
@@ -0,0 +1,396 @@
...
+#include <string.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <ulimit.h>
+#include <stdint.h>
+#include <unity_fixture.h>
+
+#define BUF_LEN         16
+#define VALUES_NUM      sizeof(values) / sizeof(values[0])
```
nalajcie, line 28: move to just after `values` tab declaration

libc/string_mem.c
```diff
@@ -0,0 +1,396 @@
...
+
+static uint8_t buf[BUF_LEN];
+static uint8_t buf_byte[256];
+static uint8_t buf_byte_exp[256];
+static uint8_t *ret;
+static uint8_t *mem;
+static uint8_t *mem_big;
+static uint8_t big_buf[1024];
+static uint8_t expect_mem_big[BIG_SIZE];
+static int values[] = { 0, 1, 16, 50, 127, 128, 200, 255,
```
nalajcie, line 47: const

nalajcie (changes requested):
Thanks for the fixes, I'm pointing some additional tests to be done and possibilities to improve the readability even further ;)

libc/string_mem.c
adamdebek: I added dst pointer also so it is even more readable in my opinion.
```C
uint8_t *src = mem + OFFSET;
	uint8_t *dst = src + TEST_STR_SIZE;
	/* Move so that memory blocks are next to each other*/
	ret = memmove(dst, src, TEST_STR_SIZE);
```

nalajcie: less readable, I need to analyze the values of `src` and `dst` to know the relation between them.

My version clearly showcases the relative offset between `src` and `dst` pointers  (and the len) by looking at a single line.

libc/string_mem.c
adamdebek: This is arbitrary value so that the block is not adjacent to TEST_STR and end of buffer.
As below:

|...............|TEST_STR|.........................|MOVED_STR|..................|
0--------1500----------2524---------------2824------------3848---------4000

libc/string_mem.c
adamdebek: Do you mean to add test case with odd ptr and len?

libc/string_mem.c
nalajcie: I know I can analyze it, the comment was regarding the code readability and possible issues when somebody would change the define values in the future
```c
const uint8_t *src = mem_big + OFFSET_BIG;
const uint32_t gapSize = 100;

/* ensure we have enough place in the buffer */
if (src + BIG_SIZE + gapSize + BIG_SIZE > mem_big + MEM_SIZE_BIG) {
  FAIL("Invalid test setup");
}
memcpy(src + BIG_SIZE + gapSize, src, BIG_SIZE);  // <--- clearly readable that I will have 2 copies of data separated by gapSize - without extra comments / ascii drawings (but i would appreciate them nevertheless ;))
/* ... */

File: libc/string_mem.c [hunk 16]
```diff
@@ -0,0 +1,391 @@
...
+	768, 257, 272, 306, 383, 384, 456, 511 };
+
+#define VALUES_NUM sizeof(values) / sizeof(values[0])
+
+
+static void setup_mem_big(int offset)
+{
+	memset(mem_big, 0, MEM_SIZE_BIG);
+	uint8_t *dst = mem_big + OFFSET_BIG;
+	TEST_ASSERT_NOT_NULL(mem_big);
```
nalajcie, line 55: too late, will sigsegv in line 53 if mem_big == NULL

libc/string_mem.c
```diff
@@ -0,0 +1,391 @@
...
+
+
+TEST(string_memset, big)
+{
+	int i, j, n = BIG_SIZE - 2 * 8;
+	for (j = 0; j < 8; j++) {
+		for (i = 0; i < 8; i++) {
+			ret = memset(big_buf + i, 10, n + j);
+			TEST_ASSERT_EQUAL_PTR(big_buf + i, ret);
+			TEST_ASSERT_EACH_EQUAL_UINT8(10, big_buf + i, n + j);
```
nalajcie, line 105: not resetting data between the tests, not checking if more than requested data was overwritten

libc/string_mem.c
```diff
@@ -0,0 +1,391 @@
...
+TEST_TEAR_DOWN(string_memset)
+{
+}
+
+
+TEST(string_memset, basic)
+{
+	int i;
+	for (i = 0; i < VALUES_NUM; i++) {
+		ret = memset(buf, values[i], BUF_LEN);
```
nalajcie, line 82: ret should be a local variable, not a global one. Applies everywhere else also

libc/string_mem.c
```diff
@@ -0,0 +1,391 @@
...
+		TEST_ASSERT_EQUAL_PTR(buf, ret);
+		TEST_ASSERT_EACH_EQUAL_INT8((uint8_t)values[i], buf, BUF_LEN);
```
nalajcie, line 84: EQUAL_UINT8

libc/string_mem.c
```diff
@@ -0,0 +1,391 @@
...
+
+TEST(string_memmove_big, no_overlap)
+{
+	uint8_t *const src = mem_big + OFFSET_BIG;
+	uint8_t *dst = mem_big + OFFSET_BIG + BIG_SIZE + 8;
+	/* Move data so that it don't overlap */
+	int i, j, z, n = BIG_SIZE - 8;
+
+	/* Test all combinations with remainders of division by word for len, dst and src */
+	for (z = 0; z < 8; z++) {
```
nalajcie, line 271: maybe use meaningful names like `srcOff`, `dstOff`, also introduce local variable inside the loops `int moveSz = n + z` to make the code more readable.

Right now it's hard to verify if the math is correct at a glance.

Also: when doing nested loops usually the outer loop is `i` and the inner loops get `j`, `k`, ... indices - please follow the common schema to avoid potential readability problems.

libc/string_mem.c
```diff
@@ -0,0 +1,391 @@
...
+
+TEST_SETUP(string_memmove)
+{
+	mem = (uint8_t *)calloc(MEM_SIZE, 1);
+	uint8_t *dst = mem + OFFSET;
+	TEST_ASSERT_NOT_NULL(mem);
+	ret = memcpy(dst, TEST_STR, TEST_STR_SIZE);
+	TEST_ASSERT_EQUAL_PTR(dst, ret);
+	TEST_ASSERT_EACH_EQUAL_UINT8(0, mem, OFFSET);
+	TEST_ASSERT_EACH_EQUAL_UINT8(0, dst + TEST_STR_SIZE, MEM_SIZE - (OFFSET + TEST_STR_SIZE));
```
nalajcie, line 139: If you're keen on using pointers instead of direct addition - I propose to go all the way

```suggestion
	mem = calloc(MEM_SIZE, 1);
	uint8_t *const  mem_end = mem + MEM_SIZE;
	TEST_ASSERT_NOT_NULL(mem);

	uint8_t *const dst = mem + OFFSET;
	uint8_t *const dst_end = dst + TEST_STR_SIZE;

	uint8_t* ret = memcpy(dst, TEST_STR, dst_end - dst);
	TEST_ASSERT_EQUAL_PTR(dst, ret);
	TEST_ASSERT_EACH_EQUAL_UINT8(0, mem, dst - mem);
	TEST_ASSERT_EACH_EQUAL_UINT8(0, dst_end, mem_end - dst_end);
```

The same can be applied in other places.

nalajcie (commented):
Thanks for the fixes.

pointed out some other issues which might be improved.

damianloew (changes requested):

nalajcie (approved):
not reading carefully but I see at a glance that the tests look much better than in the first version. Approving so my previous review won't hinder the merge process

damianloew (approved):

PR #180, opened 2023-03-23
Mateusz bloch/test strerror strsignal

mateusz-bloch:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/main.c
+++ b/libc/main.c
@@ -62,6 +62,8 @@ void runner(void)
 	RUN_TEST_GROUP(string_strcpy_stpcpy);
 	RUN_TEST_GROUP(string_strlcpy);
 	RUN_TEST_GROUP(string_strlcat);
+	RUN_TEST_GROUP(string_errsign);
+	RUN_TEST_GROUP(signal_psignal);
 }
 
 
--- a/libc/signal.c
+++ b/libc/signal.c
@@ -1,0 +1,205 @@
+/*
+ * Phoenix-RTOS
+ *
+ * POSIX.1-2017 standard library functions tests
+ * HEADER:
+ *    - signal.h
+ *
+ * TESTED:
+ *    - psignal()
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Mateusz Bloch
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <signal.h>
+#include <limits.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <unity_fixture.h>
+
+/* Typical error message does not exceed ~60 characters, that's why we expect a maximum value a little bit bigger */
+#define MAX_LEN_STRING 100
+#define STDERR_PATH    "stderr.log"
+#define SIMPLE_MSG     "Simple Message"
+
+static FILE *err_file;
+static int stderr_fd;
+
+extern const int signal_codes[];
+
+extern const unsigned int signal_codes_len;
+
+TEST_GROUP(signal_psignal);
+
+TEST_SETUP(signal_psignal)
+{
+	/* stderr duplication in pre-test state */
+	err_file = fopen(STDERR_PATH, "a+");
+	stderr_fd = dup(STDERR_FILENO);
+}
+
+
+TEST_TEAR_DOWN(signal_psignal)
+{
+	fflush(err_file);
+	/* redirect back stderr */
+	dup2(stderr_fd, STDERR_FILENO);
+	close(stderr_fd);
+	fclose(err_file);
+	remove(STDERR_PATH);
+}
+
+
+TEST(signal_psignal, basic)
+{
+/* Disabled because of #695 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/695 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#695 issue");
+#else
+	/* Since these function shall not return a value, we check only errno */
+	char buff[MAX_LEN_STRING];
+
+	for (int i = 0; i < signal_codes_len; i++) {
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
+		/* we redirect stderr to another file to read it */
+		int fd = open(STDERR_PATH, O_WRONLY, 0666);
+		dup2(fd, STDERR_FILENO);
+
+		errno = 0;
+		psignal(signal_codes[i], "Lorem Ipsum");
+		TEST_ASSERT_EQUAL_INT(0, errno);
+
+		fflush(stderr);
+
+		TEST_ASSERT_NOT_NULL(fgets(buff, MAX_LEN_STRING, err_file));
+		TEST_ASSERT_NOT_NULL(strstr(buff, "Lorem Ipsum"));
+		fflush(err_file);
+		close(fd);
+	}
+#endif
+}
+
+
+TEST(signal_psignal, ascii_string)
+{
+/* Disabled because of #695 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/695 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#695 issue");
+#else
+	char buff[MAX_LEN_STRING];
+
+	char ascii_str[MAX_LEN_STRING] = { 0 };
+
+	/*
+	 * In ASCII table from number 33 starts printable characters
+	 * in this case we want to dodge terminating ASCII signs
+	 */
+	const char first_char = '!';
+	for (char i = first_char; i < 127; i++) {
+		ascii_str[i - first_char] = i;
+	}
+
+	for (int i = 0; i < signal_codes_len; i++) {
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
+		int fd = open(STDERR_PATH, O_WRONLY, 0666);
+		dup2(fd, STDERR_FILENO);
+
+		errno = 0;
+		psignal(signal_codes[i], ascii_str);
+		TEST_ASSERT_EQUAL_INT(0, errno);
+
+		fflush(stderr);
+
+		TEST_ASSERT_NOT_NULL(fgets(buff, MAX_LEN_STRING, err_file));
+		TEST_ASSERT_NOT_NULL(strstr(buff, ascii_str));
+		fflush(err_file);
+		close(fd);
+	}
+#endif
+}
+
+
+TEST(signal_psignal, psignal_strsignal)
+{
+/* Disabled because of #695 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/695 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#695 issue");
+#else
+	char buff[MAX_LEN_STRING];
+
+	for (int i = 0; i < signal_codes_len; i++) {
+		char *signal_output = strsignal(signal_codes[i]);
+
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
+		int fd = open(STDERR_PATH, O_WRONLY, 0666);
+		dup2(fd, STDERR_FILENO);
+
+		errno = 0;
+		psignal(signal_codes[i], SIMPLE_MSG);
+		TEST_ASSERT_EQUAL_INT(0, errno);
+
+		fflush(stderr);
+
+		TEST_ASSERT_NOT_NULL(fgets(buff, MAX_LEN_STRING, err_file));
+		TEST_ASSERT_NOT_NULL(strstr(buff, SIMPLE_MSG));
+		TEST_ASSERT_NOT_NULL(strstr(buff, signal_output));
+		fflush(err_file);
+		close(fd);
+	}
+#endif
+}
+
+
+TEST(signal_psignal, psignal_strsignal_null)
+{
+/* Disabled because of #695 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/695 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#695 issue");
+#else
+	char buff[MAX_LEN_STRING];
+
+	for (int i = 0; i < signal_codes_len; i++) {
+		char *signal_output = strsignal(signal_codes[i]);
+
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
+		int fd = open(STDERR_PATH, O_WRONLY, 0666);
+		dup2(fd, STDERR_FILENO);
+
+		errno = 0;
+		psignal(signal_codes[i], NULL);
+		TEST_ASSERT_EQUAL_INT(0, errno);
+
+		fflush(stderr);
+
+		TEST_ASSERT_NOT_NULL(fgets(buff, MAX_LEN_STRING, err_file));
+		TEST_ASSERT_NOT_NULL(strstr(buff, signal_output));
+		fflush(err_file);
+		close(fd);
+	}
+#endif
+}
+
+
+TEST_GROUP_RUNNER(signal_psignal)
+{
+	RUN_TEST_CASE(signal_psignal, basic);
+	RUN_TEST_CASE(signal_psignal, ascii_string);
+	RUN_TEST_CASE(signal_psignal, psignal_strsignal);
+	RUN_TEST_CASE(signal_psignal, psignal_strsignal_null);
+}
--- a/libc/string_errsig.c
+++ b/libc/string_errsig.c
@@ -1,0 +1,197 @@
+/*
+ * Phoenix-RTOS
+ *
+ * POSIX.1-2017 standard library functions tests
+ * HEADER:
+ *    - string.h
+ *
+ * TESTED:
+ *    - strerror()
+ *    - strerror_r()
+ *    - strsignal()
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Mateusz Bloch
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <signal.h>
+#include <limits.h>
+#include <unistd.h>
+#include <unity_fixture.h>
+
+/* Typical error message does not exceed ~60 characters, that's why we expect a maximum value a little bit bigger */
+#define MAX_LEN_STRING 100
+
+/* ECANCELED, EDQUOT, EIDRM, EMULTIHOP, ENOLINK, ENOSR, ENOSTR, EOWNERDEAD, ESTALE missing in Phoenix-RTOS
+ *  #689 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/689
+ */
+static const int error_codes[] = { E2BIG, EACCES, EADDRINUSE, EADDRNOTAVAIL, EAFNOSUPPORT, EAGAIN, EALREADY, EBADF,
+	EBADMSG, EBUSY, ECHILD, ECONNABORTED, ECONNREFUSED, ECONNRESET, EDEADLK, EDESTADDRREQ, EDOM, EEXIST, EFAULT,
+	EFBIG, EHOSTUNREACH, EILSEQ, EINPROGRESS, EINTR, EINVAL, EIO, EISCONN, EISDIR, ELOOP, EMFILE, EMLINK, EMSGSIZE,
+	ENAMETOOLONG, ENETDOWN, ENETRESET, ENETUNREACH, ENFILE, ENOBUFS, ENODATA, ENODEV, ENOENT, ENOEXEC, ENOLCK, ENOMEM,
+	ENOMSG, ENOPROTOOPT, ENOSPC, ENOSYS, ENOTCONN, ENOTDIR, ENOTEMPTY, ENOTRECOVERABLE, ENOTSOCK, ENOTSUP, ENOTTY,
+	ENXIO, EOPNOTSUPP, EOVERFLOW, EPERM, EPIPE, EPROTO, EPROTONOSUPPORT, EPROTOTYPE, ERANGE, EROFS, ESPIPE, ESRCH,
+	ETIME, ETIMEDOUT, ETXTBSY, EWOULDBLOCK, EXDEV, ENOTBLK };
+
+const int signal_codes[] = { SIGABRT, SIGALRM, SIGBUS, SIGCHLD, SIGCONT, SIGFPE, SIGHUP, SIGILL, SIGINT,
+	SIGKILL, SIGPIPE, SIGQUIT, SIGSEGV, SIGSTOP, SIGTERM, SIGTSTP, SIGTTIN, SIGTTOU, SIGUSR1, SIGUSR2, SIGPROF,
+	SIGSYS, SIGTRAP, SIGURG, SIGVTALRM, SIGXCPU, SIGXFSZ };
+
+static const unsigned int error_codes_len = sizeof(error_codes) / sizeof(error_codes[0]);
+
+const unsigned int signal_codes_len = sizeof(signal_codes) / sizeof(signal_codes[0]);
+
+TEST_GROUP(string_errsign);
+
+
+TEST_SETUP(string_errsign)
+{
+}
+
+
+TEST_TEAR_DOWN(string_errsign)
+{
+}
+
+
+TEST(string_errsign, strerror_basic)
+{
+	for (int i = 0; i < error_codes_len; i++) {
+		errno = 0;
+		TEST_ASSERT_NOT_NULL(strerror(error_codes[i]));
+
+		if (i != 0) {
+			TEST_ASSERT_NOT_EQUAL_INT(0, strcmp(strerror(error_codes[i - 1]), strerror(error_codes[i])));
+		}
+
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+}
+
+
+TEST(string_errsign, strerror_zero)
+{
+	errno = 0;
+	TEST_ASSERT_NOT_NULL(strerror(0));
+	TEST_ASSERT_EQUAL_INT(0, errno);
+}
+
+
+TEST(string_errsign, strerror_r_basic)
+{
+/* Disabled because of #694 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/694 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#694 issue");
+#else
+	char prev_buf[MAX_LEN_STRING] = { 0 };
+	char buf[MAX_LEN_STRING] = { 0 };
+	for (int i = 0; i < error_codes_len; i++) {
+
+		errno = 0;
+		TEST_ASSERT_EQUAL_INT(0, strerror_r(error_codes[i], buf, sizeof(buf)));
+		TEST_ASSERT_NOT_EQUAL_INT(0, strcmp(buf, prev_buf));
+		TEST_ASSERT_EQUAL_INT(0, errno);
+
+		TEST_ASSERT_NOT_NULL(strcpy(prev_buf, buf));
+	}
+#endif
+}
+
+
+TEST(string_errsign, strerror_r_zero)
+{
+/* Disabled because of #694 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/694 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#694 issue");
+#else
+	char buf[MAX_LEN_STRING];
+	errno = 0;
+	TEST_ASSERT_EQUAL_INT(0, strerror_r(0, buf, sizeof(buf)));
+	TEST_ASSERT_EQUAL_INT(0, errno);
+#endif
+}
+
+
+TEST(string_errsign, strerror_r_invaild)
+{
+/* Disabled because of #694 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/694 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#694 issue");
+#else
+	const int values[] = { INT_MIN, INT_MIN / 2, INT_MIN / 4, -1024, -256, 256, 1024, INT_MAX / 4, INT_MAX / 2, INT_MAX };
+	for (int i = 0; i < (sizeof(values) / sizeof(values[0])); i++) {
+		char buf[MAX_LEN_STRING] = { 0 };
+		errno = 0;
+		TEST_ASSERT_EQUAL_INT(EINVAL, strerror_r(values[i], buf, sizeof(buf)));
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+#endif
+}
+
+
+TEST(string_errsign, strerror_r_overflow)
+{
+/* Disabled because of #694 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/694 */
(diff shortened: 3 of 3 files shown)


libc/string_errsig.c
```diff
@@ -0,0 +1,251 @@
...
+{
+/* sigrtmin not yet implemented in Phoenix-RTOS */
+#ifdef __phoenix__
+	TEST_IGNORE();
+#else
+	TEST_ASSERT_EQUAL_STRING("Real-time signal 0", strsignal(SIGRTMIN));
+	TEST_ASSERT_EQUAL_STRING("Real-time signal 1", strsignal(SIGRTMIN + 1));
+	TEST_ASSERT_EQUAL_STRING("Real-time signal 16", strsignal(SIGRTMIN + 16));
+#endif
+}
```
damianloew, line 233: As it's described in doc:
"It shall use the same set of messages as the `psignal()` function"

I know that it wasn't planned, but we can also add `psignal()` tests there and cover this sentence.

libc/string_errsig.c
```diff
@@ -0,0 +1,251 @@
...
+		errno = 0;
+		TEST_ASSERT_TRUE(strerror_r(values[i], buf, sizeof(buf)) == 0 || EINVAL);
+		TEST_ASSERT_TRUE(strlen(buf) <= sizeof(buf));
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+#endif
+}
+
+
+TEST(string_errsing, strerror_r_not_enough_buf)
```
damianloew, line 144:
```suggestion
TEST(string_errsing, strerror_r_overflow)
```

libc/string_errsig.c
```diff
@@ -0,0 +1,251 @@
...
+		errno = 0;
+		TEST_ASSERT_NOT_NULL(strerror(values[i]));
+		TEST_ASSERT_LESS_OR_EQUAL_INT(MAX_LEN_STRING, strlen(strerror(values[i])));
+		TEST_ASSERT_TRUE(errno == 0 || EINVAL);
+	}
+#endif
+}
+
+
+TEST(string_errsing, strerror_r_basic_error_codes)
```
damianloew, line 87:
```suggestion
TEST(string_errsing, strerror_r_basic)
```

libc/string_errsig.c
```diff
@@ -0,0 +1,251 @@
...
+{
+}
+
+
+TEST_TEAR_DOWN(string_errsing)
+{
+}
+
+
+TEST(string_errsing, strerror_basic_error_codes)
```
damianloew, line 47:
```suggestion
TEST(string_errsing, strerror_basic)
```

damianloew (changes requested):

libc/string_errsig.c
```diff
@@ -0,0 +1,251 @@
...
+ */
+
+#include <string.h>
+#include <errno.h>
+#include <signal.h>
+#include <limits.h>
+#include <unity_fixture.h>
+
+
+#define ERROR_CODES    EPERM, ENOENT, ESRCH, EINTR, EIO, ENXIO, E2BIG, ENOEXEC, EBADF, ECHILD, EAGAIN, ENOMEM, EACCES, EFAULT, ENOTBLK, EBUSY, EEXIST, EXDEV, ENODEV, ENOTDIR, EISDIR, EINVAL, ENFILE, EMFILE, ENOTTY, ETXTBSY, EFBIG, ENOSPC, ESPIPE, EROFS, EMLINK, EPIPE, EDOM
```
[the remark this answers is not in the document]
nalajcie: define is not needed, the tab with errnos is const and shared between the tests - just declare and initialize it here.

Also: we need to remember extending this test when adding new errno (ooor we can try to dynamically generate errnos at a compile time, similarity to `errno` tool from `moreutlis` - IMHO not needed right now):
https://github.com/Harvie/moreutils/blob/master/Makefile#L32

nalajcie:
please remember to use imperative mood in commit messages (**add** instead of **added**, etc.)

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+
+
+TEST_TEAR_DOWN(string_errsingn)
+{
+}
+
+
+TEST(string_errsingn, strerror_basic)
+{
+	for (int i = 0; i < (sizeof error_codes / sizeof *error_codes); i++) {
```
nalajcie, line 55: please use parentheses with `sizeof` keyword (`sizeof(error_codes) / sizeof(error_codes[0])` is a commonly used mnemonic for arrays).

IMHO You should add `static const unsigned int error_codes_len = ...` to the global scope - instead of writing it multiple times.

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+/* strerror_r() not yet implemented in Phoenix-RTOS */
+#ifdef __phoenix__
+	TEST_IGNORE();
+#else
+	for (int i = 0; i < (sizeof error_codes / sizeof *error_codes); i++) {
+		char prev_buf[MAX_LEN_STRING] = { 0 };
+		char buf[MAX_LEN_STRING] = { 0 };
+
+		errno = 0;
+		TEST_ASSERT_EQUAL_INT(0, strerror_r(error_codes[i - 1], prev_buf, sizeof(buf)));
```
nalajcie, line 98: buffer oveflow? (accessing `error_codes[-1]`)

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+		TEST_ASSERT_EQUAL_INT(0, strerror_r(error_codes[i], buf, sizeof(buf)));
+		TEST_ASSERT_LESS_OR_EQUAL_INT(MAX_LEN_STRING, strlen(prev_buf));
```
nalajcie, line 100: always true (output would always be truncated) or can't check (no truncation would result in buffer overflow)

You might want to pass `MAX_LEN_STRING / 2` as `buflen` to partially mitigate buffer overflow or intentionally pass too-small `buflen` to verify if truncation happens.

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+		TEST_ASSERT_LESS_OR_EQUAL_INT(MAX_LEN_STRING, strlen(buf));
+		TEST_ASSERT_TRUE(buf != prev_buf);
```
nalajcie, line 102: always true, You're checking pointers for match, not the buffer contents pointers are pointing to

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+
+TEST(string_errsingn, strerror_r_zero)
+{
+/* strerror_r() not yet implemented in Phoenix-RTOS */
+#ifdef __phoenix__
+	TEST_IGNORE();
+#else
+	char buf[MAX_LEN_STRING];
+	errno = 0;
+	TEST_ASSERT_EQUAL_INT(0, strerror_r(0, buf, sizeof(buf)));
```
nalajcie, line 117: invalid, see above

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+}
+
+
+TEST(string_errsingn, strerror_r_invaild)
+{
+/* strerror_r() not yet implemented in Phoenix-RTOS */
+#ifdef __phoenix__
+	TEST_IGNORE();
+#else
+	int values[] = { INT_MIN, INT_MIN / 2, -1024, 1024, INT_MAX / 2, INT_MAX };
```
nalajcie, line 130: const

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+	TEST_ASSERT_EQUAL_STRING("Information request", strsignal(SIGINFO));
+	TEST_ASSERT_EQUAL_STRING("EMT trap", strsignal(SIGEMT));
+#endif
+
+/* Disabled because of #NUMBER issue */
+#ifndef __phoenix__
+	TEST_ASSERT_EQUAL_STRING("I/O possible", strsignal(SIGPOLL));
+#endif
+
+	TEST_ASSERT_EQUAL_STRING("Hangup", strsignal(SIGHUP));
```
nalajcie, line 171: nah, don't test the strings - they are locale-depenant and are not portable!

eg. with `LANG=pl_PL.UTF-8` this returns `Rozłączenie`

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+	TEST_ASSERT_EQUAL_STRING("Profiling timer expired", strsignal(SIGPROF));
+	TEST_ASSERT_EQUAL_STRING("Window changed", strsignal(SIGWINCH));
+	TEST_ASSERT_EQUAL_STRING("User defined signal 1", strsignal(SIGUSR1));
+	TEST_ASSERT_EQUAL_STRING("User defined signal 2", strsignal(SIGUSR2));
+}
+
+
+TEST(string_errsingn, strsignal_invaild)
+{
+	int values[] = { INT_MIN, INT_MIN / 2, -64, -1, 0, 128, 1024, INT_MAX / 2, INT_MAX };
```
nalajcie, line 205: const

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+TEST_TEAR_DOWN(string_errsingn)
+{
+}
+
+
+TEST(string_errsingn, strerror_basic)
+{
+	for (int i = 0; i < (sizeof error_codes / sizeof *error_codes); i++) {
+		errno = 0;
+		TEST_ASSERT_NOT_NULL(strerror(error_codes[i - 1]));
```
nalajcie, line 57: buffer underflow

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+		TEST_ASSERT_NOT_NULL(strerror(error_codes[i]));
+		TEST_ASSERT_LESS_OR_EQUAL_INT(MAX_LEN_STRING, strlen(strerror(error_codes[i - 1])));
```
nalajcie, line 59: buffer underflow, IMHO checking strlen of returned string doesn't prove anything. Does POSIX specify that all results of `strerror` can't exceed 60 bytes?

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+		TEST_ASSERT_LESS_OR_EQUAL_INT(MAX_LEN_STRING, strlen(strerror(error_codes[i])));
+		TEST_ASSERT_TRUE(strerror(error_codes[i - 1]) != strerror(error_codes[i]));
```
nalajcie, line 61: always true, comparing pointers not values

libc/string_errsig.c
```diff
@@ -0,0 +1,247 @@
...
+#else
+static const int error_codes[] = { E2BIG, EACCES, EADDRINUSE, EADDRNOTAVAIL, EAFNOSUPPORT, EAGAIN, EALREADY, EBADF, EBADMSG, EBUSY, ECANCELED, ECHILD, ECONNABORTED, ECONNREFUSED, ECONNRESET, EDEADLK, EDESTADDRREQ, EDOM, EDQUOT, EEXIST, EFAULT, EFBIG, EHOSTUNREACH, EIDRM, EILSEQ, EINPROGRESS, EINTR, EINVAL, EIO, EISCONN, EISDIR, ELOOP, EMFILE, EMLINK, EMSGSIZE, EMULTIHOP, ENAMETOOLONG, ENETDOWN, ENETRESET, ENETUNREACH, ENFILE, ENOBUFS, ENODATA, ENODEV, ENOENT, ENOEXEC, ENOLCK, ENOLINK, ENOMEM, ENOMSG, ENOPROTOOPT, ENOSPC, ENOSR, ENOSTR, ENOSYS, ENOTCONN, ENOTDIR, ENOTEMPTY, ENOTRECOVERABLE, ENOTSOCK, ENOTSUP, ENOTTY, ENXIO, EOPNOTSUPP, EOVERFLOW, EOWNERDEAD, EPERM, EPIPE, EPROTO, EPROTONOSUPPORT, EPROTOTYPE, ERANGE, EROFS, ESPIPE, ESRCH, ESTALE, ETIME, ETIMEDOUT, ETXTBSY, EWOULDBLOCK, EXDEV, ENOTBLK };
```
nalajcie, line 36: if linux is a superset of phoenix error codes - don't bother (test just these supported by phoenix).

Also: add newlines for the table to be more readable (probably around column 120?)

nalajcie (changes requested):

libc/string_errsig.c
mateusz-bloch: strerror_r will return ERANGE instead 0, if insufficient storage was supplied via strerrbuf and/or buflen. I'm testing it in test  `strerror_r_overflow` .

libc/string_errsig.c
```diff
@@ -0,0 +1,268 @@
...
+
+	TEST_ASSERT_EQUAL_STRING("Real-time signal 30", strsignal(SIGRTMAX));
+	TEST_ASSERT_EQUAL_STRING("Real-time signal 29", strsignal(SIGRTMAX - 1));
+	TEST_ASSERT_EQUAL_STRING("Real-time signal 14", strsignal(SIGRTMAX - RTSIG_MAX / 2));
+	TEST_ASSERT_EQUAL_STRING("Unknown signal 32", strsignal(SIGRTMAX - RTSIG_MAX));
+#endif
+}
+
+
+TEST(string_errsingn, strsignal_psignal)
```
damianloew, line 218: I think we can split psignal tests to a few cases, like for the functions before (not null checks and so on) and then add a case where we compare outputs of strerror and psignal. Now it would be hard to investigate what was the reason of fail, the wrong output/error of psignal or the fact that outputs for strsignal and psignal are different. This way we will assure that there are no basic errors with those functions and then we will comapre their messages in more detailed test case.

damianloew (changes requested):

libc/string_errsig.c
```diff
@@ -0,0 +1,268 @@
...
+	}
+#endif
+}
+
+
+TEST(string_errsingn, strerror_r_zero)
+{
+/* strerror_r() not yet implemented in Phoenix-RTOS */
+#ifdef __phoenix__
+	TEST_IGNORE();
```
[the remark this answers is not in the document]
mateusz-bloch: If we delete `#else`, there will be warnings about undeclared function in this case

libc/string_errsig.c
```diff
@@ -0,0 +1,312 @@
...
+
+
+TEST(string_errsign, strsignal_real_time)
+{
+/* Disabled because of #687 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/687 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#687 issue");
+#else
+	TEST_ASSERT_NOT_NULL(strsignal(SIGRTMIN));
+	TEST_ASSERT_NOT_NULL(strstr(strsignal(SIGRTMIN), "0"));
```
damianloew, line 206: As we decided to not check the exact content of an error message I would suggest checking whether it's not NULL.

Alternatively we could leave this approach, but then it would be good to check those numbers dynamically, so not 30, but `SIGRTMAX`, because it's not specified that SIGRTMAX should equal 30. Still the more flexible solution with asserting not NULL seems to be better. Above that, we could check all the possible `RT` values (SIGRTMIN..SIGRTMAX)

libc/string_errsig.c
```diff
@@ -0,0 +1,312 @@
...
+	char buff[MAX_LEN_STRING];
+
+	for (int i = 0; i < signal_codes_len; i++) {
+		char *signal_output = strsignal(signal_codes[i]);
+
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
+		TEST_ASSERT_NOT_NULL(freopen("stderr.log", "w", stderr));
+		psignal(signal_codes[i], "");
```
damianloew, line 281: We could also check `NULL` and other edge strings, I think the whole ascii set passed in the string would be sufficient. So we will have four `psignal` cases. We could consider creating the separate test group then and place initialization and deinitialization of `err_file` in setup/teardown.

libc/string_errsig.c
```diff
@@ -0,0 +1,312 @@
...
+#else
+	for (int i = 0; i < error_codes_len; i++) {
+		char prev_buf[MAX_LEN_STRING] = { 0 };
+		char buf[MAX_LEN_STRING] = { 0 };
+
+		errno = 0;
+		TEST_ASSERT_EQUAL_INT(0, strerror_r(error_codes[i], buf, sizeof(buf)));
+
+		if (i != 0) {
+			TEST_ASSERT_EQUAL_INT(0, strerror_r(error_codes[i - 1], prev_buf, sizeof(prev_buf)));
```
damianloew, line 112: I suppose you wanted to do:
```suggestion
			TEST_ASSERT_EQUAL_INT(0, strerror_r(error_codes[i - 1], buf, sizeof(buf)));
```

libc/string_errsig.c
```diff
@@ -0,0 +1,312 @@
...
+
+
+TEST(string_errsign, strerror_r_basic)
+{
+/* Disabled because of #694 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/694 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#694 issue");
+#else
+	for (int i = 0; i < error_codes_len; i++) {
+		char prev_buf[MAX_LEN_STRING] = { 0 };
```
damianloew, line 105: We have to establish some max length of the message anyway. So providing we don't want to assert that it's shorter or equal to some value we would increase MAX_LEN_STRING to some larger value, I think 100 will be fine.

damianloew (changes requested):

libc/string_errsig.c
mateusz-bloch: This is for check if the previous message is different from the currently checked one

libc/main.c
```diff
@@ -54,6 +54,8 @@ void runner(void)
 	RUN_TEST_GROUP(stdio_scanf_squareBrackets);
 	RUN_TEST_GROUP(stdio_scanf_rest);
 	RUN_TEST_GROUP(string_chr);
+	RUN_TEST_GROUP(string_errsign);
+	RUN_TEST_GROUP(signal);
```
damianloew, line 58:
```suggestion
	RUN_TEST_GROUP(signal_psignal);
```

damianloew (changes requested):

damianloew (dismissed):

libc/signal.c
```diff
@@ -0,0 +1,196 @@
...
+
+TEST_SETUP(signal_psignal)
+{
+	err_file = fopen(PATH, "a+");
+}
+
+
+TEST_TEAR_DOWN(signal_psignal)
+{
+	remove(PATH);
```
nalajcie, line 52: `fclose(err_file)` should be moved here for symmetry with `TEST_SETUP`

libc/signal.c
```diff
@@ -0,0 +1,196 @@
...
+	TEST_IGNORE_MESSAGE("#695 issue");
+#else
+	/* Since these function shall not return a value, we check only errno */
+	char buff[MAX_LEN_STRING];
+
+	for (int i = 0; i < signal_codes_len; i++) {
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
+		TEST_ASSERT_NOT_NULL(freopen("stderr.log", "w", stderr));
```
nalajcie, line 69: please check if it's possible to restore the `stderr` to the state before the test (right now all tests after this one would have `stderr` redirected to file) - probably to be done in `TEST_TEAR_DOWN`

Also: use define (currently `PATH`)

libc/signal.c
```diff
@@ -0,0 +1,196 @@
...
+/* Disabled because of #695 issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/695 */
+#ifdef __phoenix__
+	TEST_IGNORE_MESSAGE("#695 issue");
+#else
+	char buff[MAX_LEN_STRING];
+
+	char ascii_str[2 * MAX_LEN_STRING];
+
+	for (int i = 0; i < 128; i++) {
+		ascii_str[i] = i;
```
nalajcie, line 97: ascii[0] = 0 so `strlen(ascii) = 0`, please fix the test data.

libc/signal.c
```diff
@@ -0,0 +1,196 @@
...
+		int length;
+		char *signal_output = strsignal(signal_codes[i]);
+
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
+		TEST_ASSERT_NOT_NULL(freopen("stderr.log", "w", stderr));
+
+		errno = 0;
+		psignal(signal_codes[i], "Simple Message");
```
nalajcie, line 140: maybe make it a define instead of repeating it 3 times?

eg. 
```c
#define SIMPLE_MSG "Simple Message"

psignal(signal_codes[i], SIMPLE_MSG)
length = strlen(signal_output) + strlen(SIMPLE_MSG ": ") + 1;  // TODO: explain why + 1
```

libc/signal.c
```diff
@@ -0,0 +1,196 @@
...
+#include <string.h>
+#include <errno.h>
+#include <signal.h>
+#include <limits.h>
+#include <unistd.h>
+#include <unity_fixture.h>
+
+/* Typical error message does not exceed ~60 characters, that's why we expect a maximum value a little bit bigger */
+#define MAX_LEN_STRING 100
+#define PATH           "stderr.log"
```
nalajcie, line 29: `PATH` is a too-generic name for define, use maybe `STDERR_PATH`?

libc/signal.c
```diff
@@ -0,0 +1,196 @@
...
+	for (int i = 0; i < 128; i++) {
+		ascii_str[i] = i;
+	}
+	ascii_str[128] = '\0';
+
+	for (int i = 0; i < signal_codes_len; i++) {
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
+		TEST_ASSERT_NOT_NULL(freopen("stderr.log", "w", stderr));
```
nalajcie, line 105: use define

libc/signal.c
```diff
@@ -0,0 +1,196 @@
...
+	char buff[MAX_LEN_STRING];
+
+	for (int i = 0; i < signal_codes_len; i++) {
+		int length;
+		char *signal_output = strsignal(signal_codes[i]);
+
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
+		TEST_ASSERT_NOT_NULL(freopen("stderr.log", "w", stderr));
```
nalajcie, line 137: use define

libc/signal.c
```diff
@@ -0,0 +1,196 @@
...
+	TEST_IGNORE_MESSAGE("#695 issue");
+#else
+	char buff[MAX_LEN_STRING];
+
+	for (int i = 0; i < signal_codes_len; i++) {
+		char *signal_output = strsignal(signal_codes[i]);
+
+		rewind(err_file);
+		memset(buff, 0, sizeof(buff));
+
```
nalajcie, line 171: use define

libc/signal.c
```diff
@@ -0,0 +1,196 @@
...
+		TEST_ASSERT_NOT_NULL(freopen("stderr.log", "w", stderr));
+
+		errno = 0;
+		psignal(signal_codes[i], NULL);
+		TEST_ASSERT_EQUAL_INT(0, errno);
+
+		fflush(stderr);
+
+		TEST_ASSERT_EQUAL_STRING(signal_output, fgets(buff, strlen(signal_output) + 1, err_file));
```
nalajcie, line 180: try to take more bytes or You would miss potential garbage data after the signal name

libc/string_errsig.c
```diff
@@ -0,0 +1,228 @@
...
+
+
+TEST(string_errsign, strerror_basic)
+{
+	for (int i = 0; i < error_codes_len; i++) {
+		errno = 0;
+		TEST_ASSERT_NOT_NULL(strerror(error_codes[i]));
+
+		if (i != 0) {
+			TEST_ASSERT_NOT_NULL(strerror(error_codes[i - 1]));
```
nalajcie, line 71: not needed, was tested in previous loop iteration

libc/string_errsig.c
```diff
@@ -0,0 +1,228 @@
...
+	TEST_ASSERT_EQUAL_INT(0, errno);
+}
+
+
+TEST(string_errsign, strerror_invaild)
+{
+	const int values[] = { INT_MIN, INT_MIN / 2, INT_MIN / 4, -1024, -256, 256, 1024, INT_MAX / 4, INT_MAX / 2, INT_MAX };
+	for (int i = 0; i < (sizeof(values) / sizeof(values[0])); i++) {
+		TEST_ASSERT_NOT_NULL(strerror(values[i]));
+		TEST_ASSERT_NOT_NULL(strstr(strerror(values[i]), "Unknown error"));
```
nalajcie, line 93: not guaranteed, dependent on locale - please remove

libc/string_errsig.c
```diff
@@ -0,0 +1,228 @@
...
+#else
+	for (int i = 0; i < error_codes_len; i++) {
+		char prev_buf[MAX_LEN_STRING] = { 0 };
+		char buf[MAX_LEN_STRING] = { 0 };
+
+		errno = 0;
+		TEST_ASSERT_EQUAL_INT(0, strerror_r(error_codes[i], buf, sizeof(buf)));
+
+		if (i != 0) {
+			TEST_ASSERT_EQUAL_INT(0, strerror_r(error_codes[i - 1], prev_buf, sizeof(prev_buf)));
```
nalajcie, line 112: nitpicking: could be optimized (by alternatively using 2 buffers instead of calling strerror_r twice every loop)

libc/string_errsig.c
```diff
@@ -0,0 +1,228 @@
...
+
+
+TEST(string_errsign, strsignal_basic)
+{
+	for (int i = 0; i < signal_codes_len; i++) {
+		errno = 0;
+		TEST_ASSERT_NOT_NULL(strsignal(signal_codes[i]));
+
+		if (i != 0) {
+			TEST_ASSERT_NOT_NULL(strsignal(signal_codes[i - 1]));
```
nalajcie, line 178: already tested in previous loop iteration

libc/string_errsig.c
```diff
@@ -0,0 +1,228 @@
...
+}
+
+
+TEST(string_errsign, strsignal_invaild)
+{
+	const int values[] = { INT_MIN, INT_MIN / 2, INT_MIN / 4, -1024, -256, -128, -64, -1, 0, 128, 256, 1024, INT_MAX / 4, INT_MAX / 2, INT_MAX };
+	for (int i = 0; i < (sizeof(values) / sizeof(values[0])); i++) {
+		errno = 0;
+		TEST_ASSERT_NOT_NULL(strsignal(signal_codes[i]));
+		TEST_ASSERT_NOT_NULL(strstr(strsignal(values[i]), "Unknown signal"));
```
nalajcie, line 193: implementation / locale dependant - please remove

libc/string_errsig.c
```diff
@@ -0,0 +1,228 @@
...
+		errno = 0;
+		TEST_ASSERT_EQUAL_INT(ERANGE, strerror_r(error_codes[i], buf, sizeof(buf)));
+		TEST_ASSERT_EQUAL_INT(ERANGE, strerror_r(error_codes[i], buf2, 2));
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+#endif
+}
+
+
+TEST(string_errsign, strsignal_basic)
```
nalajcie, line 171: IMHO the strsignal tests should be moved to the other file (and the array with signals above removed)

libc/string_errsig.c
```diff
@@ -0,0 +1,228 @@
...
+			TEST_ASSERT_NOT_NULL(strsignal(signal_codes[i - 1]));
+			TEST_ASSERT_NOT_EQUAL_INT(0, strcmp(strsignal(signal_codes[i - 1]), strsignal(signal_codes[i])));
+		}
+
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+}
+
+
+TEST(string_errsign, strsignal_invaild)
```
nalajcie, line 187: Invalid test, please remove.
image

libc/string_errsig.c
```diff
@@ -0,0 +1,228 @@
...
+	for (int i = 0; i < (sizeof(values) / sizeof(values[0])); i++) {
+		errno = 0;
+		TEST_ASSERT_NOT_NULL(strsignal(signal_codes[i]));
+		TEST_ASSERT_NOT_NULL(strstr(strsignal(values[i]), "Unknown signal"));
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+}
+
+
+TEST(string_errsign, strsignal_real_time)
```
nalajcie, line 199: maybe do this test only `#if defined(SIGRTMIN) && defined(SIGRTMAX)`

nalajcie (changes requested):

libc/string_errsig.c
damianloew: We would rather name files using specific header prefixes. `strsignal` is from string.h, so it's in string_errsig.c, `psignal` from signal.h so it's tested in signal.c, (or some other file with signal prefix)

nalajcie: If you feel the need to, than do so - but please keep only one copy of `signal_codes` array (and other arrays of this type in the future) as any changes to signal definitions would need to be introduced in multiple places now (and this obviously saves some `.rodata` size ;))

libc/signal.c
```diff
@@ -0,0 +1,199 @@
...
+#else
+	char buff[MAX_LEN_STRING];
+
+	char ascii_str[MAX_LEN_STRING];
+
+	/*
+	 * In ASCII table from number 33 starts printable characters
+	 * in this case we want to dodge terminating ASCII signs
+	 */
+	for (int i = 33; i < 128; i++) {
```
nalajcie, line 107: better, but `'0' + i` is still a bad idea. Also - try avoiding repeated magic numbers.

Maybe:
```c
char ascii_str[MAX_LEN_STRING] = {0};

/* use safely-printable chars */
const char first_char = '!';
for (char i = first_char; i < 120; i++) {
  ascii_str[i - first_char] = i;
}
// note: \0 is taken care of because of zeroing the array beforehand
```

OR:
```c
char c = 0, count = 0;
while (c < 128) {
  if (isprint(c)) {
   ascii_str[count++] = c;
 }
 c += 1;
}

ascii_str[count++] = '\0';
```

libc/signal.c
```diff
@@ -0,0 +1,199 @@
...
+TEST_SETUP(signal_psignal)
+{
+	err_file = fopen(STDERR_PATH, "a+");
+	stderr_fd = dup(STDERR_FILENO);
+}
+
+
+TEST_TEAR_DOWN(signal_psignal)
+{
+	TEST_ASSERT_NOT_NULL(freopen("/dev/stderr", "w", stderr));
```
nalajcie, line 55: not available on phoenix-rtos?

libc/signal.c
```diff
@@ -0,0 +1,199 @@
...
+	close(STDERR_FILENO);
```
nalajcie, line 56: `stderr` should be unbuffered but still calling `fflush(stderr)` beforehand is a good idea.

You may want to write some comment to future yourself here or in TEST_SETUP regarding what you're doing (or why)  ;)

nalajcie (commented):
thanks for the fixes, I've pointed out some small remaining nitpicks.

damianloew (dismissed):
All requested changes seems to be applied. Thanks for upgrading this test, LGTM.

libc/string_errsig.c
nalajcie: my comment was to the line 100 which is always true regardless of any other test - so testing it is pointless. Doesn't matter as this code was removed.

nalajcie (approved):
Mixing unbuffered and buffered file access is usually not a good idea but I'm not enforcing any more changes - I'm happy that we have more and more features tested :).

PR #182, opened 2023-03-24
trunner: add printing logs

damianloew:
## Description

 - trunner: add --logfile argument
 - trunner: add printing logs for github-actions

## Motivation and Context

JIRA: CI-224

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic-qemu, armv7a9-zynq7000-zedboard).

## Checklist:

## Special treatment

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -53,6 +53,7 @@ def parse_args(targets: Dict[str, TargetBase], hosts: Dict[str, Host]):
         hosts: Dictionary of available hosts.
 
     """
+
     parser = argparse.ArgumentParser()
 
     parser.add_argument(
@@ -136,6 +137,23 @@ def parse_args(targets: Dict[str, TargetBase], hosts: Dict[str, Host]):
         default=False,
         action="store_true",
         help="Nightly tests will be run",
+    )
+
+    def is_dir(dirpath):
+        """Check whether the provided path is a directory (if exists)"""
+        if os.path.exists(dirpath) and not os.path.isdir(dirpath):
+            raise argparse.ArgumentTypeError(f"{dirpath} is not a directory!")
+        return dirpath
+
+    parser.add_argument(
+        "--logdir",
+        default=None,
+        const="./phoenix_test_campaign_logs",
+        help="Directory path where log files from the test campaign will be stored. "
+        "By default log files will not be created. "
+        "When only --logdir without a value is set, uses %(const)s",
+        nargs="?",
+        type=is_dir,
     )
 
     class keyValue(argparse.Action):
@@ -223,6 +241,7 @@ def main():
         baudrate=args.baudrate,
         project_path=resolve_project_path(),
         nightly=args.nightly,
+        logdir=args.logdir,
         should_flash=not args.no_flash,
         should_test=not args.no_test,
         verbosity=args.verbose,
--- a/trunner/dut.py
+++ b/trunner/dut.py
@@ -37,6 +37,7 @@ class Dut(ABC):
 
     def __init__(self):
         self.pexpect_proc = None
+        self._logfiles = None
 
     def __getattr__(self, __name: str) -> Any:
         return getattr(self.pexpect_proc, __name)
@@ -46,6 +47,21 @@ class Dut(ABC):
             super().__setattr__(__name, __value)
         except AttributeError:
             setattr(self.pexpect_proc, __name, __value)
+
+    def set_logfiles(self, rd, wr, all):
+        self._logfiles = rd, wr, all
+
+        if self.pexpect_proc:
+            self._set_logfiles()
+
+    def get_logfiles(self):
+        return self._logfiles
+
+    def _set_logfiles(self):
+        if self._logfiles is not None:
+            self.pexpect_proc.logfile_read = self._logfiles[0]
+            self.pexpect_proc.logfile_send = self._logfiles[1]
+            self.pexpect_proc.logfile = self._logfiles[2]
 
     def clear_buffer(self):
         """
@@ -111,6 +127,7 @@ class ProcessDut(Dut):
 
     def open(self):
         self.pexpect_proc = pexpect.spawn(*self.args, **self.kwargs)
+        self._set_logfiles()
 
     def set_args(self, *args, **kwargs):
         self.args = args
--- a/trunner/test_runner.py
+++ b/trunner/test_runner.py
@@ -9,7 +9,7 @@ from trunner.config import ConfigParser
 from trunner.dut import Dut
 from trunner.harness import HarnessError, FlashError
 from trunner.text import bold, green, red, yellow
-from trunner.types import Status, TestOptions, TestResult
+from trunner.types import Status, TestOptions, TestResult, is_github_actions
 
 
 def _add_tests_module_to_syspath(project_path: Path):
@@ -41,30 +41,50 @@ def init_logdir(logdir: str):
 def set_logfiles(dut: Dut, logdir: str):
     """Sets dut logfiles associated with the pexpect process if needed."""
 
-    if not logdir:
+    if not logdir and not is_github_actions():
         return
 
     logfile_r, logfile_w, logfile_a = StringIO(""), StringIO(""), StringIO("")
     dut.set_logfiles(logfile_r, logfile_w, logfile_a)
 
 
+def format_logs_for_gh(logs):
+    """Prepares drop-down list for github actions workflows based on provided raw logs"""
+
+    # skip esc codes intended to clear window and double cr to avoid printing additional newline
+    logs = logs.replace("\r\r", "\r")
+    logs = logs.replace("\033[2J", "")
+    logs = logs.replace("\033c", "")
+    logs = "::group::show logs\n" + logs + "\n::endgroup::"
+
+    return logs
+
+
 def dump_logfiles(dut: Dut, dirname: str, logdir: str):
     """Dump logfiles to log directory or to gh actions if needed."""
 
-    if not logdir:
+    if not logdir and not is_github_actions():
         return
 
-    for logfile_name, log in zip(("out", "in", "inout"), dut.get_logfiles()):
-        logs = log.getvalue()
-        # empty string -> do not dump logs
-        if not logs:
-            return
-        if not os.path.isdir(f"{logdir}/{dirname}"):
-            os.mkdir(f"{logdir}/{dirname}")
-        with open(f"{logdir}/{dirname}/{logfile_name}.log", "w") as logfile:
-            logfile.write(logs)
-        with open(f"{logdir}/test_campaign/{logfile_name}.log", "a") as logfile:
-            logfile.write(logs)
+    # we want to dump logs in the given directory
+    if logdir:
+        for logfile_name, log in zip(("out", "in", "inout"), dut.get_logfiles()):
+            logs = log.getvalue()
+            # empty string -> do not dump logs
+            if not logs:
+                return
+            if not os.path.isdir(f"{logdir}/{dirname}"):
+                os.mkdir(f"{logdir}/{dirname}")
+            with open(f"{logdir}/{dirname}/{logfile_name}.log", "w") as logfile:
+                logfile.write(logs)
+            with open(f"{logdir}/test_campaign/{logfile_name}.log", "a") as logfile:
+                logfile.write(logs)
+    # we want to dump logs on Github Actions workflow
+    if is_github_actions():
+        logfiles = dut.get_logfiles()
+        # use only out logfile
+        logs = logfiles[0].getvalue()
+        print(format_logs_for_gh(logs))
 
 
 class TestRunner:
--- a/trunner/test_runner.py
+++ b/trunner/test_runner.py
@@ -1,8 +1,12 @@
+import os
+import shutil
 import sys
+from io import StringIO
 from pathlib import Path
 from typing import List, Sequence
 
 from trunner.config import ConfigParser
+from trunner.dut import Dut
 from trunner.harness import HarnessError, FlashError
 from trunner.text import bold, green, red, yellow
 from trunner.types import Status, TestOptions, TestResult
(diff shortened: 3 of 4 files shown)


trunner/test_runner.py
```diff
@@ -111,10 +127,20 @@ def run_tests(self, tests: Sequence[TestOptions]):
...
+                    fail += 1
+                if is_github_actions():
+                    # New logfile is created after emulator reboot and when running a new test on host-generic-pc
+                    if 'host' in self.target.name or (tests[idx].should_reboot and 'emu' in self.target.name):
+                        pos = 0
+                    logs = self.target.dut.pexpect_proc.logfile_read.getvalue()[pos:]
+                    pos = pos + len(logs)
+
+                    logs = prepare_logs(logs)
+                    print(logs)
```
jsarzy, line 143: Let's pack this code into a function. We should not touch the guts of the `Dut` class. Instead, the `Dut` class should have `get_logfiles` method that returns the triplet (logfile, logfile_read, logfile_write). Here's a possible implementation:
```suggestion
                    dump_logfiles()
                    if is_github_actions():
                        print_dut_logs(self.target.dut)
```

We don't have to keep track of the last position in the buffer. Instead, we could create a new `StringIO` instance before executing the test (not one for all tests). The workflow could look like this:
1. Set logfile to `StringIO`.
2. Execute the harness.
3. Print the content of the `StringIO`.
4. Repeat.

Additionaly, I'm not sure if it's good idea to get `logfile_read` as logs. I think logfile would fit better here. If the input is not echoed, then there is no input context in the logs.

jsarzy (changes requested):
Thanks for making the changes! I have a few comments.

Perhaps we could keep logs for each executed test as well? It would make navigation much easier. To do that ,we could use `StringIO` as a logfile and then dump its contents to the file for the test case and to the file that keeps all logs. This approach may be slightly less efficient, as we iterate over these logs a few times (first saving them to StringIO and then to the file) but I don't think it's crucial in this case.

trunner/dut.py
```diff
@@ -95,6 +121,10 @@ def open(self):
             raise PortError(e) from e
 
         self.pexpect_proc = pexpect.fdpexpect.fdspawn(self.serial, *self.args, **self.kwargs)
+        if self.logdir:
```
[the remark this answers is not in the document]
damianloew: It was more reasonable for me to init pexpect logfile just next to `spawn/fdspawn`.

trunner/dut.py
```diff
@@ -61,6 +70,23 @@ def clear_buffer(self):
...
 
+    def prepare_pexpect_logfiles(self):
+        """Creates or clears pexpect logfiles."""
+
+        if not os.path.isdir(self.logdir):
+            os.mkdir(self.logdir)
+
+        for logfile_name in self.lognames.values():
+            with open(f'{self.logdir}/{logfile_name}', "w"):
+                pass
```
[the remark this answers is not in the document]
damianloew: This is done to clear a file.

runner.py
```diff
@@ -233,6 +244,9 @@ def main():
     host = host_cls.from_context(ctx)
     ctx = dataclasses.replace(ctx, host=host)
 
+    # Set global context
+    trunner.ctx = ctx
+
```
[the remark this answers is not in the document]
damianloew: Probably not needed as logging won't be done in DUT.

trunner/test_runner.py
damianloew: I don't think that `logfile` is better than `logfile_read`, in most of the cases sent commands are echoed and usually it's very difficult to read. Ok, maybe you are right, I will think about a different solution.

trunner/test_runner.py
```diff
@@ -20,6 +24,69 @@ def resolve_project_path():
...
+    internal_logdir_names = ("out", "in", "inout")
+    if not os.path.isdir(logdir):
+        os.mkdir(logdir)
+
+    for dirname in internal_logdir_names:
+        if not os.path.isdir(f"{logdir}/{dirname}"):
+            os.mkdir(f"{logdir}/{dirname}")
+        # clear logfiles from the whole test campaign
+        else:
+            with open(f"{trunner.ctx.logdir}/{dirname}/test_campaign.log", "w"):
```
jsarzy, line 42: I would go with a simple open().close() as we don't need to use the benefits of a context manager here

trunner/test_runner.py
```diff
@@ -20,6 +24,69 @@ def resolve_project_path():
...
+                break
```
jsarzy, line 43: I think you meant `pass` instead of `break`.

trunner/test_runner.py
```diff
@@ -20,6 +24,69 @@ def resolve_project_path():
...
+        return
+
+    logfile_r, logfile_w, logfile_a = StringIO(""), StringIO(""), StringIO("")
+    dut.set_logfiles(logfile_r, logfile_w, logfile_a)
+
+
+def format_logs_for_gh(logs_in):
+    """Prepares drop-down list for github actions workflows based on provided raw logs"""
+
+    logs = logs_in
```
jsarzy, line 59: I suggest reaming `logs_in` to just `logs`, since we are doing this anyway in the first statement.

trunner/test_runner.py
```diff
@@ -20,6 +24,69 @@ def resolve_project_path():
     return project_dir
 
 
+def init_logdir():
+    """Inits log directories if it's needed."""
+    if not trunner.ctx.logdir:
+        return
+
+    logdir = trunner.ctx.logdir
+    internal_logdir_names = ("out", "in", "inout")
```
jsarzy, line 33: I'm wondering if this is a good directory structure for logs. I think it would make more sense to group these logs by test, with directories named after each tests, and files inside named log.out, log.in and log.inout. Alternatively, we could avoid creating additional directories and simply save the logs as name.in, name.out, name.inout.

With this approach, when a test fails, we can simply go into the directory named after that test, or look for name.*, rather than having to navigate into the in, out and inout directories. This should also simplify the code.

What do you think?

trunner/test_runner.py
```diff
@@ -20,6 +24,69 @@ def resolve_project_path():
...
+    if not trunner.ctx.logdir and not is_github_actions():
+        return
+
+    test_basename = test.name.replace("phoenix-rtos-tests/", "")
+    test_basename = test_basename.replace("/", ".")
+
+    for idx, internal_dirname in enumerate(("out", "in", "inout")):
+        logfiles = dut.get_logfiles()
+        logs = logfiles[idx].getvalue()
+        if trunner.ctx.logdir:
```
jsarzy, line 80: We did this check in the line 71

trunner/test_runner.py
```diff
@@ -20,6 +24,69 @@ def resolve_project_path():
...
+
+def dump_logfiles(dut: Dut, test):
+    """Dump logfiles to log directory or to gh actions if needed."""
+    if not trunner.ctx.logdir and not is_github_actions():
+        return
+
+    test_basename = test.name.replace("phoenix-rtos-tests/", "")
+    test_basename = test_basename.replace("/", ".")
+
+    for idx, internal_dirname in enumerate(("out", "in", "inout")):
```
jsarzy, line 77:
```suggestion
    for dir, log in zip(("out", "in", "inout"), dut.get_logfiles()):
```
I think this should simplify things! dut.get_logfiles is called once and we don't have to touch any indices.

trunner/test_runner.py
```diff
@@ -20,6 +24,69 @@ def resolve_project_path():
...
+    logs = logs.replace("\033[2J", "")
+    logs = logs.replace("\033c", "")
+    logs = "::group::show logs\n" + logs + "\n::endgroup::"
+
+    return logs
+
+
+def dump_logfiles(dut: Dut, test):
+    """Dump logfiles to log directory or to gh actions if needed."""
+    if not trunner.ctx.logdir and not is_github_actions():
```
jsarzy, line 71: Let's pass logdir as an argument here. That way, we will be able to avoid importing context.

trunner/test_runner.py
jsarzy: I think maybe we should also clean the entire logdir. If I only do a few tests using `-t` flag then older logs will persist and this could be confusing.

jsarzy (changes requested):

trunner/test_runner.py
damianloew: yeah thanks, as the `with` solution is replaced it's outdated.

trunner/test_runner.py
damianloew: Yeah, we discussed it and your solution seems to be more suitable. It will be available in the next force push.

trunner/test_runner.py
damianloew: Yeah, but there were also condition with gh_actions, we rather don't want to create logfiles by default when using gh actions. Please see the last changes.

runner.py
```diff
@@ -53,6 +53,15 @@ def parse_args(targets: Dict[str, TargetBase], hosts: Dict[str, Host]):
...
     """
+
+    def_logdir = "./phoenix_test_campaign_logs"
+
+    def is_dir(dirpath):
+        """Check whether the provided path is a directory (if exists) """
+        if os.path.exists(dirpath) and not os.path.isdir(dirpath):
+            raise argparse.ArgumentTypeError(f"{dirpath} is not a directory!")
+        return dirpath
+
```
jsarzy, line 64: I suggest moving this code before line 150.

trunner/test_runner.py
```diff
@@ -20,6 +25,69 @@ def resolve_project_path():
...
+        shutil.rmtree(logdir)
+
+    # test campaign directory will always be needed
+    os.makedirs(f"{logdir}/test_campaign")
+
+
+def set_logfiles(dut: Dut):
+    """Sets dut logfiles associated with the pexpect process if needed."""
+
+    if not trunner.ctx.logdir and not is_github_actions():
```
jsarzy, line 44: Please pass logdir as an argument to the `set_logfiles` function. Let's try to be consistent in all function headers.

trunner/test_runner.py
```diff
@@ -20,6 +25,69 @@ def resolve_project_path():
     return project_dir
 
 
+def init_logdir(logdir):
```
jsarzy, line 28: It would be nice to have type hint for logdir

trunner/test_runner.py
```diff
@@ -20,6 +25,69 @@ def resolve_project_path():
...
+            with open(f"{logdir}/{dirname}/{logfile_name}.log", "w") as logfile:
+                logfile.write(logs)
+            with open(f"{logdir}/test_campaign/{logfile_name}.log", "a") as logfile:
+                logfile.write(logs)
+    # we want to dump logs on Github Actions workflow
+    if is_github_actions():
+        logfiles = dut.get_logfiles()
+        # use only out logfile
+        logs = logfiles[0].getvalue()
+        if logs:
```
jsarzy, line 87: ditto

trunner/test_runner.py
```diff
@@ -154,8 +227,12 @@ def run(self) -> bool:
 
         tests = self.parse_tests()
 
+        init_logdir(trunner.ctx.logdir)
+
         if self.ctx.should_flash:
+            set_logfiles(self.target.dut)
             self.flash()
+            dump_logfiles(self.target.dut, "flash", trunner.ctx.logdir)
```
jsarzy, line 235:
```suggestion
            dump_logfiles(self.target.dut, "flash", self.ctx.logdir)
```

trunner/test_runner.py
```diff
@@ -154,8 +227,12 @@ def run(self) -> bool:
 
         tests = self.parse_tests()
 
+        init_logdir(trunner.ctx.logdir)
```
jsarzy, line 230:
```suggestion
        init_logdir(self.ctx.logdir)
```

trunner/test_runner.py
```diff
@@ -20,6 +25,69 @@ def resolve_project_path():
...
+    # skip esc codes intended to clear window and double cr to avoid printing additional newline
+    logs = logs.replace("\r\r", "\r")
+    logs = logs.replace("\033[2J", "")
+    logs = logs.replace("\033c", "")
+    logs = "::group::show logs\n" + logs + "\n::endgroup::"
+
+    return logs
+
+
+def dump_logfiles(dut: Dut, dirname, logdir):
```
jsarzy, line 63: Please add type hints for dirname and logdir

trunner/test_runner.py
```diff
@@ -1,11 +1,16 @@
+import os
+import shutil
 import sys
+from io import StringIO
 from pathlib import Path
 from typing import List, Sequence
 
+import trunner
```
jsarzy, line 8: This import should not be needed

jsarzy (changes requested):
I suggested small changes, overall looks good

trunner/test_runner.py
damianloew: Yeah, I see it now, it will be also corrected in other functions.

trunner/test_runner.py
```diff
@@ -20,6 +25,69 @@ def resolve_project_path():
...
+    if not logdir and not is_github_actions():
+        return
+
+    # we want to dump logs in the given directory
+    if logdir:
+        for logfile_name, log in zip(("out", "in", "inout"), dut.get_logfiles()):
+            logs = log.getvalue()
+            # empty string -> do not dump logs
+            if not logs:
+                return
```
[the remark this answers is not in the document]
damianloew: The drop-down list without a content on github may be misleading. I don't know whether we want change it (if yes, it's a task for another PR), but now the log `Flashing the system image...` is printed even on targets, where there is no flashing operations (qemu, host), there are no logs then. I would opt for leaving this solution as it is, empty directory could also be confusing IMHO

jsarzy (approved):

PR #184, opened 2023-03-28
trunner: fix handling TEST_IGNORE_MESSAGE

mateusz-bloch:
JIRA CI-244

## Description

Fixed TEST_IGNORE_MESSAGE handling

Before changes:
image

After changes:
image

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/harness/unity.py
+++ b/trunner/harness/unity.py
@@ -20,6 +20,8 @@ def _format_result_output(results: Sequence[Dict[str, Any]], ctx: TestContext) -
             out += green("OK")
         elif res["status"] == "IGNORE":
             out += yellow("IGNORE")
+            if "msg" in res:
+                out += ": " + res["msg"]
         elif res["status"] == "INFO":
             out += blue("INFO")
 
@@ -31,13 +33,13 @@ def _format_result_output(results: Sequence[Dict[str, Any]], ctx: TestContext) -
 
 
 def unity_harness(dut: Dut, ctx: TestContext) -> Optional[TestResult]:
-    assert_re = r"ASSERTION (?P<path>.*?):(?P<line>\d+):(?P<status>FAIL|INFO|IGNORE): (?P<msg>.*?)\r"
+    assert_re = r"ASSERTION (?P<path>[\S]+):(?P<line>\d+):(?P<status>FAIL|INFO|IGNORE): (?P<msg>.*?)\r"
     result_re = r"TEST\((?P<group>\w+), (?P<name>\w+)\) (?P<status>PASS|IGNORE)"
     # Fail need to have its own regex due to greedy matching
     result_final_re = r"TEST\((?P<group>\w+), (?P<name>\w+)\) (?P<status>FAIL) at (?P<path>.*?):(?P<line>\d+)\r"
     final_re = r"(?P<total>\d+) Tests (?P<fail>\d+) Failures (?P<ignore>\d+) Ignored"
 
-    last_fail = {"path": None, "line": None}
+    last_assertion = {}
     stats = {"FAIL": 0, "IGNORE": 0, "PASS": 0}
     results = []
 
@@ -46,13 +48,13 @@ def unity_harness(dut: Dut, ctx: TestContext) -> Optional[TestResult]:
         parsed = dut.match.groupdict()
 
         if idx == 0:
-            if parsed["status"] == "FAIL":
-                last_fail = parsed
+            if parsed["status"] in ["FAIL", "IGNORE"]:
+                last_assertion = parsed
         elif idx in (1, 2):
-            # TODO we do not consider INFO/IGNORE messages
-            if parsed["status"] == "FAIL":
-                if last_fail["path"] == parsed["path"] and last_fail["line"] == parsed["line"]:
-                    parsed["msg"] = last_fail["msg"]
+            # TODO we do not consider INFO messages
+            if last_assertion and last_assertion["status"] != "INFO":
+                parsed["msg"] = last_assertion["msg"]
+                last_assertion = {}
 
             stats[parsed["status"]] += 1
             results.append(parsed)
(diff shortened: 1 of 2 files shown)


nalajcie:
please use imperative mood when writing commit messages (so `fix` instead of `fixed`)

jsarzy (changes requested):
Thanks for changes, I'd appreciate some description to pull request to know what we want to achieve. I made proof of concept and verified that indeed we do not catch ignored test cases when we use `TEST_IGNORE` (but not in case when it's defined as a test header).

This is a mix of bug and lack of feature.
We customized the way unity prints output to make parsing easier. So the bug is the omission of the newline character- after line 1825 in unity.c we have to add `UNITY_PRINT_EOL();`. For this reason, the ignore messages are not the prettiest:
```
ASSERTION phoenix-rtos-tests/sample/test/test-unity.c:51:IGNORETEST(unity_example, example_5) IGNORE
ASSERTION phoenix-rtos-tests/sample/test/test-unity.c:56:IGNORE: one two threeTEST(unity_example, example_6) IGNORE
```
It should look like this:
```
ASSERTION phoenix-rtos-tests/sample/test/test-unity.c:51:IGNORE
TEST(unity_example, example_5) IGNORE
ASSERTION phoenix-rtos-tests/sample/test/test-unity.c:56:IGNORE: one two three
TEST(unity_example, example_6) IGNORE
```
What we want to do here is:
1. Fix the bug in unity.c
2. Catch the assertion message for ignored test (similar what we do with failed test cases)
3. Catch the header message for ignored test (also similar to what we have with fails)
4. Add message to result output for ignored test cases if the assertion message was present.

trunner/harness/unity.py
```diff
@@ -46,13 +48,17 @@ def unity_harness(dut: Dut, ctx: TestContext) -> Optional[TestResult]:
...
+            # TODO we do not consider INFO messages
             if parsed["status"] == "FAIL":
-                if last_fail["path"] == parsed["path"] and last_fail["line"] == parsed["line"]:
-                    parsed["msg"] = last_fail["msg"]
+                if last_assertion["path"] == parsed["path"] and last_assertion["line"] == parsed["line"]:
+                    parsed["msg"] = last_assertion["msg"]
+            if parsed["status"] == "IGNORE":
+                if last_assertion["status"] == parsed["status"]:
+                    parsed["msg"] = last_assertion["msg"]
+                    last_assertion["status"] = None
```
jsarzy, line 61: Let's set `last_assertion` to an empty dict before line 63. I think is more readable, it clearly says that we've read the test final message and we clear the last assertion because it should not be re-used later.

unity/unity.c
```diff
@@ -1823,7 +1823,10 @@ void UnityIgnore(const char* msg, const UNITY_LINE_TYPE line)
         UNITY_OUTPUT_CHAR(' ');
         UnityPrint(msg);
     }
-    UNITY_IGNORE_AND_BAIL;
+
+	UNITY_PRINT_EOL();
+
+	UNITY_IGNORE_AND_BAIL;
```
jsarzy, line 1829: User 4 spaces instead of tabs. We would like to keep the same formatting as it is external library. You can ignore github action formatter comments here.

trunner/harness/unity.py
```diff
@@ -31,13 +33,13 @@ def _format_result_output(results: Sequence[Dict[str, Any]], ctx: TestContext) -
...
 def unity_harness(dut: Dut, ctx: TestContext) -> Optional[TestResult]:
-    assert_re = r"ASSERTION (?P<path>.*?):(?P<line>\d+):(?P<status>FAIL|INFO|IGNORE): (?P<msg>.*?)\r"
+    assert_re = r"ASSERTION (?P<path>[\S]+):(?P<line>\d+):(?P<status>FAIL|INFO|IGNORE): (?P<msg>.*?)\r"
     result_re = r"TEST\((?P<group>\w+), (?P<name>\w+)\) (?P<status>PASS|IGNORE)"
     # Fail need to have its own regex due to greedy matching
     result_final_re = r"TEST\((?P<group>\w+), (?P<name>\w+)\) (?P<status>FAIL) at (?P<path>.*?):(?P<line>\d+)\r"
     final_re = r"(?P<total>\d+) Tests (?P<fail>\d+) Failures (?P<ignore>\d+) Ignored"
 
-    last_fail = {"path": None, "line": None}
+    last_assertion = {"path": None, "line": None, "status": None}
```
jsarzy, line 42: I would set it to an empty dict and later check if there is last_assertion.

trunner/harness/unity.py
```diff
@@ -46,13 +48,17 @@ def unity_harness(dut: Dut, ctx: TestContext) -> Optional[TestResult]:
         parsed = dut.match.groupdict()
 
         if idx == 0:
-            if parsed["status"] == "FAIL":
-                last_fail = parsed
+            if parsed["status"] in ["FAIL", "IGNORE"]:
+                last_assertion = parsed
         elif idx in (1, 2):
-            # TODO we do not consider INFO/IGNORE messages
+            # TODO we do not consider INFO messages
```
jsarzy, line 54: Maybe do check here `if last_assertion and last_assertion["status"] != "INFO"`. Then we know that last assertion exists and it's not INFO.

trunner/harness/unity.py
```diff
@@ -46,13 +48,17 @@ def unity_harness(dut: Dut, ctx: TestContext) -> Optional[TestResult]:
...
             if parsed["status"] == "FAIL":
-                if last_fail["path"] == parsed["path"] and last_fail["line"] == parsed["line"]:
-                    parsed["msg"] = last_fail["msg"]
+                if last_assertion["path"] == parsed["path"] and last_assertion["line"] == parsed["line"]:
+                    parsed["msg"] = last_assertion["msg"]
+            if parsed["status"] == "IGNORE":
+                if last_assertion["status"] == parsed["status"]:
```
jsarzy, line 59: Do the check I mentioned above and then this if should not be needed.

jsarzy (commented):
Thanks for changes, I verified and it works well. I made a few comments you could apply.

jsarzy (approved):

PR #192, opened 2023-04-11
mateusz-bloch/libc: printf tests

mateusz-bloch:
JIRA CI-260

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/stdio_printf.c
+++ b/libc/stdio_printf.c
@@ -1,0 +1,3677 @@
+/*
+ * Phoenix-RTOS
+ *
+ * POSIX.1-2017 standard library functions tests
+ * HEADER:
+ *    - stdio.h
+ *
+ * TESTED:
+ *    - dprintf()
+ *    - fprintf()
+ *    - printf()
+ *    - snprintf()
+ *    - sprintf()
+ *    - vdprintf()
+ *    - vfprintf()
+ *    - vprintf()
+ *    - vsnprintf()
+ *    - vsprintf()
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Mateusz Bloch
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+#include <stdarg.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <float.h>
+#include <limits.h>
+#include <stdbool.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <unity_fixture.h>
+#include <signal.h>
+
+static FILE *output_file;
+static char buff[700];
+
+
+#define PATH "stdio_printf_test"
+
+/* comma delimiter macro ',' for multiple arguments parsing to *printf() function via macros */
+#define COMMA ,
+
+/* macro assertions*/
+#define dtest(format, value, expect) \
+	{ \
+		memset(buff, 0, sizeof(buff)); \
+		rewind(output_file); \
+		dprintf(fileno(output_file), format, value); \
+		fseek(output_file, 0, SEEK_SET); \
+		fflush(output_file); \
+		TEST_ASSERT_EQUAL_STRING(expect, fgets(buff, strlen(expect) + 1, output_file)); \
+		TEST_ASSERT_EQUAL_STRING(expect, buff); \
+	}
+
+#define ftest(format, value, expect) \
+	{ \
+		memset(buff, 0, sizeof(buff)); \
+		rewind(output_file); \
+		fprintf(output_file, format, value); \
+		fseek(output_file, 0, SEEK_SET); \
+		fflush(output_file); \
+		TEST_ASSERT_EQUAL_STRING(expect, fgets(buff, strlen(expect) + 1, output_file)); \
+		TEST_ASSERT_EQUAL_STRING(expect, buff); \
+	}
+
+#define stdtest(format, value, expect) \
+	{ \
+		rewind(output_file); \
+		memset(buff, 0, sizeof(buff)); \
+		int stdout_fd = dup(STDOUT_FILENO); \
+		int fd = open(PATH, O_WRONLY | O_CREAT, 0666); \
+		dup2(fd, STDOUT_FILENO); \
+		printf(format, value); \
+		fflush(stdout); \
+		fseek(output_file, 0, SEEK_SET); \
+		fflush(output_file); \
+		/* redirect back to stdout */ \
+		dup2(stdout_fd, STDOUT_FILENO); \
+		close(fd); \
+		close(stdout_fd); \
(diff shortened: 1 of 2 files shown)


damianloew (changes requested):

libc/stdio_printf.c
```diff
@@ -0,0 +1,3323 @@
...
+		vsntest(expect, format, values[i]);
+		vstest(expect, format, values[i]);
+	}
+}
+
+
+TEST(stdio_printf_u, tu)
+{
+	const char *format = "%tu";
+	const ptrdiff_t values[] = { PTRDIFF_MAX, PTRDIFF_MAX / 2, 0 };
```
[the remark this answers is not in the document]
mateusz-bloch: I based on scanf test, but main reason is that for specifier o, u, x we use unsigned argument
image

libc/stdio_printf.c
```diff
@@ -0,0 +1,3323 @@
...
+}
+
+
+TEST(stdio_printf_cspn, c_non_ascii)
+{
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmultichar"
+#pragma GCC diagnostic ignored "-Woverflow"
+	const char *format = "%c";
+	const char values[6] = { 'é', 'ć', 'ñ', 'ж', 'न', 'げ' };
```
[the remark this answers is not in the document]
damianloew: I meant testing all possible non ascii values in for.

libc/stdio_printf.c
```diff
@@ -0,0 +1,3336 @@
...
+
+	memset(buff, 0, sizeof(buff));
+	snprintf(buff, 7, "%6s", "abc");
+	TEST_ASSERT_EQUAL_STRING("   abc", buff);
+
+#pragma GCC diagnostic pop
+}
+
+
+TEST(stdio_printf_rest, errnos)
```
damianloew, line 3174: Please write some short comments why the rest of errnos are not checked - for example: `/* EILSEQ not tested, because locales are not yet supported on Phoenix-RTOS`


I think that catching `EOVERFLOW` could be possible.

damianloew (changes requested):
Really good job, the code is clear and seems to cover most of the requirements from documentation. Is the `#` modifier tested somewhere there?
Screenshot from 2023-05-25 17-07-10

damianloew (changes requested):
Really good job, the code is clear and seems to cover most of the requirements from documentation. Is the `#` modifier tested somewhere there?
Screenshot from 2023-05-25 17-07-10

libc/stdio_printf.c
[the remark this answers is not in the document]
mateusz-bloch: All take unsigned argument, but i see we can test also signed values to see if its correctly converted

libc/stdio_printf.c
```diff
@@ -0,0 +1,3336 @@
...
+#endif
+
+#ifndef DBL_MIN
+#define DBL_MIN 2.225074e-308
+#endif
+
+#define PATH "stdio_printf_test"
+
+/* comma delimiter macro ',' for multiple arguments parsing to *printf() function via macros */
+#define COMMA ,
```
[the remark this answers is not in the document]
mateusz-bloch: It is for multiple arguments parsing to *printf() function via macros

mateusz-bloch:
> Really good job, the code is clear and seems to cover most of the requirements from documentation. Is the `#` modifier tested somewhere there? Screenshot from 2023-05-25 17-07-10

Tested in `TEST(stdio_printf_rest, mods_o_x)`

libc/stdio_printf.c
```diff
@@ -0,0 +1,3322 @@
...
+		vstest(expect[i], format, values[i], values[i], values[i], values[i], values[i], values[i]);
+	}
+}
+
+
+TEST(stdio_printf_rest, mods_float)
+{
+	int i;
+	const char *format = "%2f %10f %010f %+f %.5f %-10f";
+	const float values[] = { FLT_MAX, 0, FLT_MIN, -1.23, 1.23, 0.23456789 };
```
damianloew, line 2970: I think adding there DBL_MAX and DBL_MIN values could be valuable and certainly asserting the significant figures in edge min values.

libc/stdio_printf.c
```diff
@@ -0,0 +1,3322 @@
...
+}
+
+
+TEST(stdio_printf_cspn, c_ascii)
+{
+#ifdef __phoenix__
+	int i = 0;
+#else
+	int i = 1;
+#endif
```
damianloew, line 2592: why like this?

damianloew:
> 

It could be good to test it also with a, A, e, E, f, F as it's mentioned.

damianloew (changes requested):
Please test also values out of range (I think 1 value for each case would be sufficient, for example testing `INT_MAX + 10` and try to print it using `%d`)

damianloew:
> > Really good job, the code is clear and seems to cover most of the requirements from documentation. Is the `#` modifier tested somewhere there? Screenshot from 2023-05-25 17-07-10
> 
> Tested in `TEST(stdio_printf_rest, mods_o_x)`

It could be good to test it also with a, A, e, E, f, F as it's mentioned.

libc/stdio_printf.c
```diff
@@ -0,0 +1,3322 @@
...
+	remove(PATH);
+}
+
+
+TEST(stdio_printf_o, o)
+{
+	int i;
+	char expect[256] = { 0 };
+	const char *format = "%o %o %o %o %o %o %o";
+	const unsigned int values[] = { UINT_MAX, UINT_MAX / 2, INT_MAX, INT_MAX / 2, 0, INT_MIN / 2, INT_MIN };
```
damianloew, line 797: Why ints are passed to unisgned int array? You may want to tests values out of bonds, but I think we can do it in separate cases, because as you can see you would need the another type.

damianloew: Applies to the other places like this one.

libc/stdio_printf.c
```diff
@@ -0,0 +1,3582 @@
...
+
+	vsprintf(buff, format, arg_ptr);
+	TEST_ASSERT_EQUAL_STRING(expect, buff);
+
+	va_end(arg_ptr);
+}
+
+
+/* other functions */
+static char *signed_to_string(long long int value, int base)
```
damianloew, line 221: Let's prefix functions names like that with `test_` and use camelCase as it's mentioned in our coding convention. I am not sure about renaming macros above, maybe let's leave them as they are to not increase binary size so much.

```suggestion
static char *test_signedToStr(long long int value, int base)
```

We could do it for the rest of functions analogically.

damianloew (changes requested):
IMHO it's really good test, I know that the size is very big and there is a problem with `stm32l4x6-nucleo` target, but please check the idea I pasted you today, maybe it will be ok as the separate binary.

Above that please leave a comment next to all pragmas, which disables a warning

damianloew (approved):

PR #194, opened 2023-04-25
mateusz-bloch/libc: add test for strto functions

mateusz-bloch:
added tests for strtod(), strtof(), strtold(), strtol(), strtoll(), strtoul(), strtoull()

JIRA CI-238

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


nalajcie:
We already have some `strtod` tests in separate file - maybe we should include them here? (or move strtod tests there?)

damianloew:
> We already have some `strtod` tests in separate file - maybe we should include them here? (or move strtod tests there?)

Yeah, but we want to cover this function in more detailed way with the rest of strto* tests. Not seen those tests yet, but there was an assumption of using the existing strtod tests or covering the same utilities in a different way.

nalajcie:
> > We already have some `strtod` tests in separate file - maybe we should include them here? (or move strtod tests there?)
> 
> Yeah, but we want to cover this function in more detailed way with the rest of strto* tests. Not seen those tests yet, but there was an assumption of using the existing strtod tests or covering the same utilities in a different way (then removing it).

As you've written, this PR should be about extending existing `strtod` tests, so the original tests should probably be moved ot the new destination and reused in this PR (not in some unspecified future). 

Please note that the currently-existing tests were designed to explicitly show errors in previous `strtod` libphoenix implementation so it's crucial to include them (`strtod` weaknesses were originally found while running libstdc++ test suite)

libc/stdlib_strto.c
```diff
@@ -0,0 +1,645 @@
...
+		-.75e+8, -.75E+8, 0, 3140.0, 0.0314, 3140, -0.0314, 2.22507386e-308, 1.79769313e+308 };
+
+	for (int i = 0; i < (sizeof(str) / sizeof(str[0])); i++) {
+		errno = 0;
+		TEST_ASSERT_EQUAL_DOUBLE(expected[i], strtod(str[i], NULL));
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+}
+
+//
```
damianloew, line 61:
```suggestion
```

libc/stdlib_strto.c
```diff
@@ -0,0 +1,645 @@
...
+
+
+TEST_TEAR_DOWN(stdlib_strto)
+{
+}
+
+
+TEST(stdlib_strto, strtod_basic)
+{
+	const char *str[] = { "0", "-0", "-1", "1", "+1", ".1", "3.1415", "-3.1415", "0.1234567891234567",
```
damianloew, line 48: I would also prefer to define DBL_MAX and DBL_MIN and test them too. Applies to the other cases.

libc/stdlib_strto.c
```diff
@@ -0,0 +1,645 @@
...
+		TEST_ASSERT_EQUAL_DOUBLE(expected[i], strtold(str[i], NULL));
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+}
+
+
+TEST(stdlib_strto, strtol_basic)
+{
+	const char *str[] = { "0", "1", "-1234567890", "1234567890", "2147483647", "-2147483648" };
+	const long expected[] = { 0, 1, -1234567890, 1234567890, INT_MAX, INT_MIN };
```
damianloew, line 148: There are LONG_MAX, LONG_MIN macros too.

libc/stdlib_strto.c
```diff
@@ -0,0 +1,671 @@
...
+
+	for (int i = 0; i < (sizeof(str) / sizeof(str[0])); i++) {
+		errno = 0;
+		TEST_ASSERT_EQUAL_INT(expected[i], strtoull(str[i], NULL, 16));
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+}
+
+
+TEST(stdlib_strto, strto_inf_nan)
```
damianloew, line 325: I know that you wanted to be consistent with other names, but IMHO this prefix is not needed and that suggests test case applies to all `strto` functions

```suggestion
TEST(stdlib_strto, inf_nan)
```

libc/stdlib_strto.c
```diff
@@ -0,0 +1,671 @@
...
+TEST(stdlib_strto, strto_empty)
+{
+	const char str[] = "";
+	char *end = NULL;
+
+	errno = 0;
+	TEST_ASSERT_NULL(end);
+	TEST_ASSERT_EQUAL_DOUBLE(0, strtod(str, &end));
+	TEST_ASSERT_EQUAL_PTR(str, end);
+	TEST_ASSERT_TRUE(errno == EINVAL || errno == 0);
```
damianloew, line 424: When does it return 0? I suppose that EINVAL should be setas errno.

damianloew (changes requested):
Thanks for those tests, I think we could also add a test case with checking the string with all possible white-space characters before placing the number.

libc/stdlib_strto.c
mateusz-bloch: According to POSIX: When "No conversion could be performed", these functions may set errno to EINVAL.  So errno in these situations can be either 0 or EINVAL. 
image

damianloew (changes requested):
IMHO now it's a good test, but it would be nice to have a little bit more checks to provide better coverage:

libc/stdlib_strto.c
mateusz-bloch: This is checked in the following test: ` TEST(stdlib_strto, invalid_base)`

damianloew (changes requested):
Please add more checks for various numbers (especially floating point ones) in decimal notation.

libc/stdlib/stdlib_strto.c
```diff
@@ -0,0 +1,1121 @@
...
+
+
+TEST(stdlib_strto, strtold_min_max)
+{
+	errno = 0;
+	TEST_ASSERT_EQUAL_DOUBLE(hugeDouble, strtold(hugeDoubleString, NULL));
+	TEST_ASSERT_EQUAL_INT(0, errno);
+
+/* LDBL_MIN_MAX not declared on Phoeix_RTOS */
+#ifdef __phoenix__
```
damianloew, line 257: Are you sure? I think it could be visible as other defines from float.h:
Screenshot from 2023-06-27 16-24-54

damianloew: This way the rest of code in this test case is disabled for Phoenix so the next ifdefs have no effect.

libc/stdlib/stdlib_strto.c
```diff
@@ -0,0 +1,1121 @@
...
+		errno = 0;
+		TEST_ASSERT_EQUAL_INT(1.23, strtoull(str_float[i], NULL, 10));
+		TEST_ASSERT_EQUAL_INT(0, errno);
+	}
+}
+
+
+TEST(stdlib_strto, truncate_whitespaces)
+{
+	const char *str_int[] = { "\t123", "\v123", "\f123", "\n123", "\r123", "\r\n123", " 123" };
```
damianloew, line 848: Thanks for adding the rest of white spaces, I meant adding also a one check with all of them in one string.

damianloew (changes requested):
I don't see a similar case in the new version, so please add `TEST(strtod_family, remaining_string)` from the previous one and add @DawidSzpejna to the authors. In my opinion it will be ready for merge then.

damianloew (approved):
LGTM

PR #195, opened 2023-04-26
libc: strchr tests: add test group for ASCII compare

maska989:
## Description
Added strcspn and strspn test cases

## Motivation and Context
JIRA: CI-232

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/string/string_lenchr.c
+++ b/libc/string/string_lenchr.c
@@ -3,13 +3,15 @@
  *
  * POSIX.1-2017 standard library functions tests
  * HEADER:
- * 	- string.h
+ *    - string.h
  * TESTED:
- * 	- strlen()
- * 	- strnlen()
+ *    - strlen()
+ *    - strnlen()
  * 	- strchr()
  * 	- strrchr()
- * 	- memchr()
+ *    - strchr()
+ *    - strrchr()
+ *    - memchr()
  *
  * Copyright 2023 Phoenix Systems
  * Author: Damian Modzelewski
@@ -21,15 +23,14 @@
 
 
 #include <stdio.h>
-#include <errno.h>
 #include <string.h>
-#include <limits.h>
 #include <stdlib.h>
-#include <sys/stat.h>
 #include <unistd.h>
-#include <limits.h>
 #include <unity_fixture.h>
 
+#include "testdata.h"
+
+#define BUFF_SIZE 128
 
 TEST_GROUP(string_len);
 TEST_GROUP(string_chr);
@@ -52,11 +53,12 @@ TEST(string_len, ascii)
 			   torn[] = "foo\0bar",
 			   doubleNul[] = "\0\0abc",
 			   specials[] = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~",
-			   whites[] = " \v\t\r\n";
-	char asciiSet[128] = { 0 };
-	int sz, i;
+			   whites[] = " \v\t\r\n",
+			   *asciiSet;
+	int sz;
 
 	TEST_ASSERT_EQUAL_INT(0, strlen(""));
+	asciiSet = testdata_createCharStr(BUFF_SIZE + 1);
 
 	/* Pangram with a whole alphabet set */
 	sz = sizeof(pangram) - 1;
@@ -96,63 +98,50 @@ TEST(string_len, ascii)
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(whites, sz + 1));
 
 	/* Checking ascii charset */
-	for (i = 1; i < sizeof(asciiSet); i++) {
-		asciiSet[i - 1] = i;
-	}
 
-	sz = sizeof(asciiSet) - 1;
+
+	sz = BUFF_SIZE;
 	TEST_ASSERT_EQUAL_INT(sz, strlen(asciiSet));
 	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(asciiSet, sz - 1));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(asciiSet, sz));
 	TEST_ASSERT_EQUAL_INT(sz, strnlen(asciiSet, sz + 1));
+
+	free((void *)asciiSet);
 }
 
 
 TEST(string_len, not_ascii)
 {
-	const char notAsciiString[] = "♦♥♣♠◊⊗こんにちは❉❉⌨⌨⌨⌨⌨⌨⌨⌨❉❉";
-	unsigned char notAsciiSet[129];
+	unsigned char charSet[BUFF_SIZE] = { 0 };
 	int sz, i;
 
-	/* Checking ability to read the chars out of ascii charset */
-	sz = sizeof(notAsciiString) - 1;
-	TEST_ASSERT_EQUAL_INT(sz, strlen(notAsciiString));
-	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(notAsciiString, sz - 1));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(notAsciiString, sz));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(notAsciiString, sz + 1));
-
-	/* Checking  out of ascii bytes */
-	for (i = 128; i <= 255; i++) {
-		notAsciiSet[i - 128] = i;
+	/* Checking  out of ASCII bytes */
+	for (i = sizeof(charSet); i < sizeof(charSet) * 2 - 1; i++) {
+		charSet[i - sizeof(charSet)] = i;
 	}
-	notAsciiSet[128] = 0;
 
-	sz = sizeof(notAsciiSet) - 1;
-	TEST_ASSERT_EQUAL_INT(sz, strlen((const char *)notAsciiSet));
-	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen((const char *)notAsciiSet, sz - 1));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen((const char *)notAsciiSet, sz));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen((const char *)notAsciiSet, sz + 1));
+	sz = sizeof(charSet) - 1;
+	TEST_ASSERT_EQUAL_INT(sz, strlen((const char *)charSet));
+	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen((const char *)charSet, sz - 1));
+	TEST_ASSERT_EQUAL_INT(sz, strnlen((const char *)charSet, sz));
+	TEST_ASSERT_EQUAL_INT(sz, strnlen((const char *)charSet, sz + 1));
 }
 
 
 TEST(string_len, big)
 {
-	char bigstr[PATH_MAX] = { 0 };
-	int sz, i;
+	char bigStr[PATH_MAX] = { 0 };
+	int sz;
 
-	/* The length of string is not clearly restricted, so we test one of bigger value, which may be used  */
-	for (i = 0; i < PATH_MAX - 1; i++) {
-		bigstr[i] = 'A';
-	}
-	bigstr[i] = '\0';
+	/* The length of the string is not restricted, so we test one of bigger value, which may be used  */
+	memset(bigStr, 1, sizeof(bigStr) - 1);
 
-	sz = sizeof(bigstr) - 1;
-	TEST_ASSERT_EQUAL_INT(sz, strlen(bigstr));
-	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(bigstr, sz - 1));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz + 1));
+	sz = sizeof(bigStr) - 1;
+	TEST_ASSERT_EQUAL_INT(sz, strlen(bigStr));
+	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(bigStr, sz - 1));
+	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigStr, sz));
+	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigStr, sz + 1));
 }
-
 
 /*
 ////////////////////////////////////////////////////////////////////////////////////
@@ -177,21 +166,21 @@ TEST(string_chr, basic)
 	sz = strlen(lorem);
 
 	/* Check of proper working base */
-	TEST_ASSERT_EQUAL_STRING(lorem, strchr(lorem, 'L'));
-	TEST_ASSERT_EQUAL_STRING(lorem, strrchr(lorem, 'L'));
-	TEST_ASSERT_EQUAL_STRING(lorem, memchr(lorem, 'L', sz));
+	TEST_ASSERT_EQUAL_PTR(lorem, strchr(lorem, 'L'));
+	TEST_ASSERT_EQUAL_PTR(lorem, strrchr(lorem, 'L'));
+	TEST_ASSERT_EQUAL_PTR(lorem, memchr(lorem, 'L', sz));
 
-	TEST_ASSERT_EQUAL_STRING("lor sit amet, consectetur adipiscing elit y", strchr(lorem, 'l'));
-	TEST_ASSERT_EQUAL_STRING("lit y", strrchr(lorem, 'l'));
-	TEST_ASSERT_EQUAL_STRING("lor sit amet, consectetur adipiscing elit y", memchr(lorem, 'l', sz));
+	TEST_ASSERT_EQUAL_PTR(&lorem[14], strchr(lorem, 'l'));
+	TEST_ASSERT_EQUAL_PTR(&lorem[52], strrchr(lorem, 'l'));
+	TEST_ASSERT_EQUAL_PTR(&lorem[14], memchr(lorem, 'l', sz));
 
-	TEST_ASSERT_EQUAL_STRING(NULL, strchr(lorem, 'x'));
-	TEST_ASSERT_EQUAL_STRING(NULL, strrchr(lorem, 'x'));
-	TEST_ASSERT_EQUAL_STRING(NULL, memchr(lorem, 'x', sz));
+	TEST_ASSERT_EQUAL_PTR(NULL, strchr(lorem, 'x'));
+	TEST_ASSERT_EQUAL_PTR(NULL, strrchr(lorem, 'x'));
+	TEST_ASSERT_EQUAL_PTR(NULL, memchr(lorem, 'x', sz));
 
-	TEST_ASSERT_EQUAL_STRING(&lorem[sz], strchr(lorem, lorem[sz]));
-	TEST_ASSERT_EQUAL_STRING(&lorem[sz], strrchr(lorem, lorem[sz]));
-	TEST_ASSERT_EQUAL_STRING(&lorem[sz], memchr(lorem, lorem[sz], sz + 1));
+	TEST_ASSERT_EQUAL_PTR(&lorem[sz], strchr(lorem, lorem[sz]));
+	TEST_ASSERT_EQUAL_PTR(&lorem[sz], strrchr(lorem, lorem[sz]));
+	TEST_ASSERT_EQUAL_PTR(&lorem[sz], memchr(lorem, lorem[sz], sz + 1));
 }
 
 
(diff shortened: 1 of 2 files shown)


libc/string_lenchr.c
```diff
@@ -150,9 +153,41 @@ TEST(string_len, big)
...
+TEST_SETUP(string_spn)
+{
+}
+
+
+TEST_TEAR_DOWN(string_spn)
+{
+}
+
+TEST(string_spn, ascii)
```
damianloew, line 165: Please think about more cases, for example my first thought is that string s2 which is not present in s1 is not checked (when strspn would return 0). I suppose there won't be much more edge cases, but things like empty strings, bigger buffers, pointing the same string in the same place in memory as you did it in strcspn), pointing to the same strings, which are in different variables and maybe some other things.

damianloew (changes requested):

mateusz-bloch (commented):
Thanks for this tests. I've pointed out some small nitpicks.

Also I would consider renaming the commit to more closely match the changes, and change beginning of the file:
```
 * TESTED:
 * 	- strlen()
 * 	- strnlen()
 *     -  strcspn()
 *     -  strspn()
 * 	- strchr()
 * 	- strrchr()
 * 	- memchr()
```

libc/string_lenchr.c
```diff
@@ -404,11 +546,21 @@ TEST(string_chr, memchr_size)
...
 
 
+TEST_GROUP_RUNNER(string_spn)
+{
+	RUN_TEST_CASE(string_spn, ascii);
+	RUN_TEST_CASE(string_spn, diff_chars);
+	RUN_TEST_CASE(string_spn, big);
+	RUN_TEST_CASE(string_spn, empty_output);
+	RUN_TEST_CASE(string_spn, out_of_set);
+}
```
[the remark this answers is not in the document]
maska989: This case was tested in full scale in TEST(string_spn,ascii) where we counting length based on generated array.

libc/string_lenchr.c
```diff
@@ -108,30 +111,30 @@ TEST(string_len, ascii)
 }
 
 
-TEST(string_len, not_ascii)
+TEST(string_len, diff_chars)
 {
-	const char notAsciiString[] = "♦♥♣♠◊⊗こんにちは❉❉⌨⌨⌨⌨⌨⌨⌨⌨❉❉";
-	unsigned char notAsciiSet[129];
+	const char testArray[] = "♦♥♣♠◊⊗こんにちは❉❉⌨⌨⌨⌨⌨⌨⌨⌨❉❉";
```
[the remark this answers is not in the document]
nalajcie: **Don't use UTF-8 chars for initializing non-widechars array**. 

This was already discussed in other PR.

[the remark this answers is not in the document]
damianloew: Yeah, I have already told about it on our internal chat - it will be changed soon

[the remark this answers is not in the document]
maska989: After rethink deleted all cases where this string was used.

mateusz-bloch (approved):
LGTM

damianloew (changes requested):
@mateusz-bloch thanks for previous reviews and @maska989 for much cleaner code, a few things more to add/change.

libc/string_lenchr.c
```diff
@@ -177,21 +306,21 @@ TEST(string_chr, basic)
 	sz = strlen(lorem);
 
 	/* Check of proper working base */
-	TEST_ASSERT_EQUAL_STRING(lorem, strchr(lorem, 'L'));
-	TEST_ASSERT_EQUAL_STRING(lorem, strrchr(lorem, 'L'));
-	TEST_ASSERT_EQUAL_STRING(lorem, memchr(lorem, 'L', sz));
+	TEST_ASSERT_EQUAL_PTR(lorem, strchr(lorem, 'L'));
```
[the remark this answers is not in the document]
damianloew: It's a `const char *`, so it's not possible.

libc/string_lenchr.c
```diff
@@ -96,61 +99,255 @@ TEST(string_len, ascii)
...
+	TEST_ASSERT_EQUAL_INT(0, strspn(emptySet, ""));
+
+	TEST_ASSERT_EQUAL_INT(0, strcspn(emptySet, "abc"));
+	TEST_ASSERT_EQUAL_INT(0, strspn(emptySet, "abc"));
+
+	TEST_ASSERT_EQUAL_INT(0, strcspn("", "abc"));
+	TEST_ASSERT_EQUAL_INT(0, strspn("", "abc"));
+
+	TEST_ASSERT_EQUAL_INT(0, strcspn("", ""));
+	TEST_ASSERT_EQUAL_INT(0, strspn("", ""));
```
damianloew, line 239: As `s1` is a string it does not matter, whether there is one or more NUL terminators.

```suggestion
	TEST_ASSERT_EQUAL_INT(0, strcspn("", "abc"));
	TEST_ASSERT_EQUAL_INT(0, strspn("", "abc"));

	TEST_ASSERT_EQUAL_INT(0, strcspn("", ""));
	TEST_ASSERT_EQUAL_INT(0, strspn("", ""));
```

libc/string_lenchr.c
```diff
@@ -96,61 +99,255 @@ TEST(string_len, ascii)
...
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz + 1));
+	TEST_ASSERT_EQUAL_INT(sz, strcspn(bigstr, ""));
+	TEST_ASSERT_EQUAL_INT(sz, strspn(bigstr, bigstr));
+}
+
+
+TEST(string_spn, empty_output)
+{
+	char emptySet[BUFF_SIZE] = { 0 };
```
damianloew, line 224:
```suggestion
```

libc/string_lenchr.c
```diff
@@ -96,61 +97,189 @@ TEST(string_len, ascii)
...
+	 */
+	memset(bigstr, 'a', sizeof(bigstr) - 1);
 
 	sz = sizeof(bigstr) - 1;
-	TEST_ASSERT_EQUAL_INT(sz, strlen(bigstr));
-	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(bigstr, sz - 1));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz + 1));
+	TEST_ASSERT_EQUAL_INT(sz, strcspn(bigstr, ""));
+	TEST_ASSERT_EQUAL_INT(sz, strspn(bigstr, bigstr));
```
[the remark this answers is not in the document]
damianloew: Did you add such check? If not please do it (it may be done within this test case).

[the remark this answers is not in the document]
damianloew: Did you add such check? If not please do it (it may be done within this test case).

libc/string_lenchr.c
```diff
@@ -148,6 +149,213 @@ TEST(string_len, big)
...
+		TEST_ASSERT_EQUAL_INT(0, strcspn(asciiSet, supportSet));
+
+		/* Same case as before */
+		if (i < (BUFF_SIZE - 1) / 2) {
+			TEST_ASSERT_EQUAL_INT(BUFF_SIZE - 2 - i, strspn(asciiSet, supportSet));
+		}
+		else {
+			TEST_ASSERT_EQUAL_INT(i + 1, strspn(asciiSet, supportSet));
+		}
+	}
```
damianloew, line 323: The question is what do you want to test there. Please try to name the utility you test and create a separate test case for it. Above that there are some debug printfs.

libc/string_lenchr.c
```diff
@@ -148,6 +149,213 @@ TEST(string_len, big)
...
+	for (i = 0; i < BUFF_SIZE - 1; i++) {
+
+		/*
+		 * Checking if strcspn and strspn will search
+		 * the whole string before returning value and
+		 * he found an element of impact on the output
+		 */
+
+		/* Setting only the first element of set to values from asciiSet */
+		supportSet[0] = asciiSet[i];
```
damianloew, line 297: From what I see it's checking whether 2 same bytes are not a problem. You could check it, but it should be done in the separate test case and I would place there the strings like "aa", "aabbcc", "aaaaaa" as s2 and "abc" as s1 then. Not so many iterations and cases are needed in my opinion. I don't quite understand what you meant in the comment above, we could talk about it in the office.

libc/string_lenchr.c
```diff
@@ -96,61 +99,255 @@ TEST(string_len, ascii)
...
+		}
+	}
+
+	free((void *)asciiSet);
+}
+
+
+TEST(string_spn, out_of_set)
+{
+	char pangram[45] = " The quick brown fox jumps over the lazy dog",
```
damianloew, line 331: Why 45? Yoo could declare it this way:
```suggestion
	const char pangram[] = " The quick brown fox jumps over the lazy dog",
```

Please remember about `const`.

libc/string_lenchr.c
```diff
@@ -96,61 +99,255 @@ TEST(string_len, ascii)
...
+		else {
+			TEST_ASSERT_EQUAL_INT(i + 1, strspn(asciiSet, supportSet));
+		}
+	}
+
+	free((void *)asciiSet);
+}
+
+
+TEST(string_spn, out_of_set)
```
damianloew, line 329: I don't quite understand the need of this test case, what is out of set there?

damianloew (approved):

damianloew (changes requested):
A few things more to improve, please look at the suggestions above.

libc/string/string_lenchr.c
```diff
@@ -96,61 +99,215 @@ TEST(string_len, ascii)
...
+
+	/*
+	 * The length of the string is not restricted,
+	 * so we test one of the bigger values, which may be used
+	 * with remembering the last element must be a null term
+	 * zero to collect the data as the string from
+	 * declared space, that's why we memset one place less than its size
+	 */
 
 	sz = sizeof(bigstr) - 1;
```
damianloew, line 249: I think it could be better to name it `len` to avoid misunderstanding. Size is usually the return value of `sizeof` and len is the length of the string (return value of strlen).

```suggestion
	len = sizeof(bigstr) - 1;
```

libc/string/string_lenchr.c
```diff
@@ -96,61 +99,215 @@ TEST(string_len, ascii)
...
-	TEST_ASSERT_EQUAL_INT(sz, strlen(bigstr));
-	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(bigstr, sz - 1));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz + 1));
+	memset(bigstr, 'a', sz);
+
+	TEST_ASSERT_EQUAL_INT(sz, strcspn(bigstr, ""));
+	TEST_ASSERT_EQUAL_INT(sz, strspn(bigstr, bigstr));
```
damianloew, line 253: If we have buffers provided we can add there a little bit more checks (I know that some of them were placed in empty_output case, but when using big string as s1 it could behave in a different way). Above that the utility of placing more than one same byte in s2 should be tested in the separate case (and probably is), that's why "a" would be sufficient there.

```suggestion
	TEST_ASSERT_EQUAL_INT(sz, strcspn(bigstr, ""));
	TEST_ASSERT_EQUAL_INT(0, strcspn(bigstr, "a"));
	TEST_ASSERT_EQUAL_INT(sz, strspn(bigstr, "a"));
	TEST_ASSERT_EQUAL_INT(0, strspn(bigstr, ""));
```

libc/string_lenchr.c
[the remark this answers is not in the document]
damianloew: Oh, sorry there was a mistake, I meant `s1` instead of `s2`.

libc/string/string_lenchr.c
```diff
@@ -96,61 +99,215 @@ TEST(string_len, ascii)
...
+
+	/* Creating bounds for strings operations remembering we start counting from 0 */
+	testStr[BUFF_SIZE - 1] = 0;
+	reversStr[BUFF_SIZE - 1] = 0;
+
+	TEST_ASSERT_EQUAL_INT(0, strcspn("abc", "cba"));
+	TEST_ASSERT_EQUAL_INT(3, strspn("cba", "abc"));
+
+	TEST_ASSERT_EQUAL_INT(0, strcspn("abc", "bac"));
+	TEST_ASSERT_EQUAL_INT(3, strspn("bac", "abc"));
```
damianloew, line 300: The goal in this case is to test mixed order of `s2` with the same `s1` and assert that a behavior is proper. Mixing order in `s1` checks it too, but it's less readable in my opinion.

```suggestion
	TEST_ASSERT_EQUAL_INT(0, strcspn("abc", "cba"));
	TEST_ASSERT_EQUAL_INT(3, strspn("abc", "cba"));

	TEST_ASSERT_EQUAL_INT(0, strcspn("abc", "bac"));
	TEST_ASSERT_EQUAL_INT(3, strspn("abc", "bac"));

	TEST_ASSERT_EQUAL_INT(0, strcspn("abc", "bca"));
	TEST_ASSERT_EQUAL_INT(3, strspn("abc", "bca"));
```

libc/string/string_lenchr.c
```diff
@@ -96,61 +99,215 @@ TEST(string_len, ascii)
...
+{
+	int i;
+	char *reversStr,
+		*testStr;
+
+	testStr = testdata_createCharStr(BUFF_SIZE);
+	reversStr = malloc(BUFF_SIZE + 1);
+
+	for (i = 1; i < BUFF_SIZE; i++) {
+		reversStr[i - 1] = testStr[BUFF_SIZE - i];
```
damianloew, line 289: If a reverse string has a NUL term at the beginning it's in fact empty str, you want to start from the last valid character of testStr.

libc/string/string_lenchr.c
```diff
@@ -96,61 +99,215 @@ TEST(string_len, ascii)
...
+
+	TEST_ASSERT_EQUAL_INT(0, strcspn("abc", "bac"));
+	TEST_ASSERT_EQUAL_INT(3, strspn("bac", "abc"));
+
+	TEST_ASSERT_EQUAL_INT(BUFF_SIZE - 1, strcspn(testStr, reversStr));
+	TEST_ASSERT_EQUAL_INT(0, strspn(testStr, reversStr));
+
+	/* Last element of set testStr is null terminator zero which means we need to start from one place further */
+	TEST_ASSERT_EQUAL_INT(0, strcspn(&reversStr[1], testStr));
+	TEST_ASSERT_EQUAL_INT(BUFF_SIZE - 2, strspn(&reversStr[1], testStr));
```
damianloew, line 307: As I mentioned above we don't need to check s1 in reversed order there.

```suggestion
```

libc/string/string_lenchr.c
```diff
@@ -409,6 +568,17 @@ TEST_GROUP_RUNNER(string_len)
...
 
 
+TEST_GROUP_RUNNER(string_spn)
+{
+	RUN_TEST_CASE(string_spn, ascii);
+	RUN_TEST_CASE(string_spn, not_ascii);
+	RUN_TEST_CASE(string_spn, empty_output);
+	RUN_TEST_CASE(string_spn, mixed_order);
+	RUN_TEST_CASE(string_spn, basic);
+	RUN_TEST_CASE(string_spn, big);
```
damianloew, line 578: The basic case should always be tested at the beginning.

```suggestion
	RUN_TEST_CASE(string_spn, basic);
	RUN_TEST_CASE(string_spn, ascii);
	RUN_TEST_CASE(string_spn, not_ascii);
	RUN_TEST_CASE(string_spn, empty_output);
	RUN_TEST_CASE(string_spn, mixed_order);
	RUN_TEST_CASE(string_spn, big);
```

libc/string_lenchr.c
damianloew: Did you add it somewhere?

libc/string/string_lenchr.c
```diff
@@ -96,61 +99,215 @@ TEST(string_len, ascii)
...
+
+
+TEST(string_spn, empty_output)
+{
+	TEST_ASSERT_EQUAL_INT(0, strcspn("", "abc"));
+	TEST_ASSERT_EQUAL_INT(0, strspn("", "abc"));
+
+	TEST_ASSERT_EQUAL_INT(0, strcspn("", ""));
+	TEST_ASSERT_EQUAL_INT(0, strspn("", ""));
+}
```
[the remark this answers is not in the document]
maska989:
```
strcspn("", "abc")
```
This element is present on commented lines.

libc/string/string_lenchr.c
```diff
@@ -52,105 +54,303 @@ TEST(string_len, ascii)
...
+
+	TEST_ASSERT_EQUAL_INT(0, strcspn("", ""));
+	TEST_ASSERT_EQUAL_INT(0, strspn("", ""));
+
+	TEST_ASSERT_EQUAL_INT(3, strcspn("abc", ""));
+	TEST_ASSERT_EQUAL_INT(0, strspn("abc", ""));
+
+	TEST_ASSERT_EQUAL_INT(1, strcspn("abc", "bc"));
+	TEST_ASSERT_EQUAL_INT(0, strspn("abc", "bc"));
+
```
damianloew, line 297: As you can see, there are no empty arguments, but IMHO it's a good case to check. Let's create another test case and name it `first_byte`, or sth like that, then I'd suggest: `strcspn("abc", "a")` instead of `strcspn("abc", "bc")` to check scenarios where initial byte shouldn't be counted.

It would be good to add a comment to this test case, sth like: `This case is intended to check the ability of stopping at first byte`

You can add more checks, like:
 - `strspn("abc", "c")`
 - `strcspn("abc", "ab")`
 - `strspn("aaaabc", "bc")`
 - `strcspn("aaaabc", "a")`

libc/string/string_lenchr.c
```diff
@@ -96,61 +99,215 @@ TEST(string_len, ascii)
...
+	support[sz - 3] = 'b';
+	TEST_ASSERT_EQUAL_INT(sz, strspn(bigstr, support));
+	support[sz - 3] = 'a';
+	TEST_ASSERT_EQUAL_INT(sz, strspn(bigstr, support));
+
+	memset(support, 'b', sz);
+	support[sz - 3] = 'a';
+	TEST_ASSERT_EQUAL_INT(0, strcspn(bigstr, support));
+	support[sz - 3] = 'b';
+	TEST_ASSERT_EQUAL_INT(sz, strcspn(bigstr, support));
```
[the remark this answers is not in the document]
damianloew: I meant changed supportStr, so with `b` within, please look at the newest suggestions.

libc/string/string_lenchr.c
```diff
@@ -52,105 +54,306 @@ TEST(string_len, ascii)
 			   torn[] = "foo\0bar",
 			   doubleNul[] = "\0\0abc",
 			   specials[] = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~",
-			   whites[] = " \v\t\r\n";
-	char asciiSet[128] = { 0 };
-	int sz, i;
+			   whites[] = " \v\t\r\n",
```
mateusz-bloch, line 57: Why not test every white space?

libc/string/string_lenchr.c
```diff
@@ -52,105 +54,306 @@ TEST(string_len, ascii)
...
+		}
 	}
-	bigstr[i] = '\0';
 
-	sz = sizeof(bigstr) - 1;
-	TEST_ASSERT_EQUAL_INT(sz, strlen(bigstr));
-	TEST_ASSERT_EQUAL_INT(sz - 1, strnlen(bigstr, sz - 1));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz));
-	TEST_ASSERT_EQUAL_INT(sz, strnlen(bigstr, sz + 1));
+	free((void *)asciiStr);
```
mateusz-bloch, line 215: (void*) not needed?
```suggestion
	free(asciiStr);
```

libc/string/string_lenchr.c
```diff
@@ -52,105 +54,306 @@ TEST(string_len, ascii)
...
+
+
+TEST(string_spn, mixed_order)
+{
+	int i;
+	char *reversStr,
+		*testStr;
+
+	testStr = testdata_createCharStr(BUFF_SIZE);
+	reversStr = malloc(BUFF_SIZE);
```
mateusz-bloch, line 313: I'm considering whether it is reasonable to use malloc with the current value of BUFF_SIZE.
```suggestion
	char *reversStr[BUFF_SIZE],
		*testStr;

	testStr = testdata_createCharStr(BUFF_SIZE);
```

libc/string/string_lenchr.c
```diff
@@ -52,105 +54,303 @@ TEST(string_len, ascii)
...
+	supportStr[len - 3] = 'b';
+	TEST_ASSERT_EQUAL_INT(len - 3, strspn(supportStr, bigstr));
+	supportStr[len - 3] = 'a';
+	TEST_ASSERT_EQUAL_INT(len, strspn(supportStr, bigstr));
+
+	memset(supportStr, 'b', len);
+	supportStr[len - 3] = 'a';
+	TEST_ASSERT_EQUAL_INT(len - 3, strcspn(supportStr, bigstr));
+	supportStr[len - 3] = 'b';
+	TEST_ASSERT_EQUAL_INT(len, strcspn(supportStr, bigstr));
```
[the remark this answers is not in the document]
damianloew: Yeah, but as you can see `bigstr[len - 3] = 'b';` was missing in my suggestion, without it there will be the same block of code two times.

libc/string/string_lenchr.c
damianloew: Most of the mentioned cases haven't been applied. All of checks there should return 0 as we check this utility (stopping at once).

damianloew (changes requested):

libc/string/string_lenchr.c
```diff
@@ -52,105 +54,306 @@ TEST(string_len, ascii)
...
+}
+
+
+TEST(string_spn, mixed_order)
+{
+	int i;
+	char *reversStr,
+		*testStr;
+
+	testStr = testdata_createCharStr(BUFF_SIZE);
```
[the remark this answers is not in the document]
maska989: It is done in #199

damianloew (approved):

PR #198, opened 2023-04-27
libc: add bsearch test

adamdebek:
JIRA: CI-239

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/stdlib_bsearch.c
+++ b/libc/stdlib_bsearch.c
@@ -1,0 +1,306 @@
+/*
+ * Phoenix-RTOS
+ *
+ *    POSIX.1-2017 standard library functions tests
+ *    HEADER:
+ *    - stdlib.h
+ *    TESTED:
+ *    - bsearch()
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Adam Debek
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <unity_fixture.h>
+
+
+struct node {
+	const char *string;
+	int length;
+};
+
+/* Special node with reversed attributes */
+struct node2 {
+	int length;
+	const char *string;
+};
+
+
+/* table sorted alphabetically */
+static const char *const test_names[] = { "adam",
+	"bartek",
+	"cezary",
+	"damian",
+	"edward",
+	"fryderyk",
+	"gracjan",
+	"henryk",
+	"iwo",
+	"jan" };
+
+static const char *const namesRepeat[] = { "adam",
+	"adam",
+	"adam",
+	"damian",
+	"edward",
+	"fryderyk",
+	"fryderyk",
+	"henryk",
+	"henryk",
+	"jan" };
+
+/* table sorted alphabetically */
+static const char *const names_nsort[] = { "henryk",
+	"bartek",
+	"cezary",
+	"damian",
+	"edward",
+	"fryderyk",
+	"gracjan",
+	"adam",
+	"iwo",
+	"jan" };
+
+#define TABSIZE (sizeof(test_names) / sizeof(test_names[0]))
+
+
+static void test_prepare(const char *const names_arg[], struct node table[])
+{
+	int i;
+
+	for (i = 0; i < TABSIZE; i++) {
+		table[i].string = names_arg[i];
+		table[i].length = strlen(names_arg[i]);
+	}
+}
+
+
+/* bsearch can't change content of table, but may reorder elements */
+static int test_checkTab(const char *const names_arg[], struct node table[])
+{
+	int i, j, found;
+	int found_tab[TABSIZE];
+	memset(found_tab, 0, sizeof(int) * TABSIZE);
+
+	for (i = 0; i < TABSIZE; i++) {
+		found = 0;
+		for (j = 0; j < TABSIZE && found_tab[i] == 0; j++) {
+			if (strcmp(names_arg[i], table[j].string) == 0 && table[i].length == table[j].length) {
+				found = 1;
+				found_tab[i] = 1;
+				break;
+			}
+		}
+		if (found == 0) {
+			return found;
+		}
+	}
+
+	return found;
+}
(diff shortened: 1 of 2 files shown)


damianloew (commented):
What about edge cases like passing 0 as `nel`? I suppose that there are a little bit more things to test (make sure that every utility is tested and use as many edge cases as possible). Maybe it's not crucial to have many of tests for this function, but still please look through it and think about possible scenarios.

libc/bsearch.c
```diff
@@ -0,0 +1,213 @@
...
+	for (i = 0; i < TABSIZE; i++) {
+		node.string = names[i];
+		node.length = strlen(names[i]);
+		rval = (struct node *)bsearch((void *)&node, (void *)table,
+			(size_t)TABSIZE, sizeof(struct node), node_cmp);
+		TEST_ASSERT_NOT_NULL(rval);
+		TEST_ASSERT_EQUAL_MEMORY(&table[i], rval, sizeof(*rval));
+	}
+
+	TEST_ASSERT_EQUAL_INT(1, check_tab(names));
```
damianloew, line 141: Why only name in each member is checked? The `length` element could potentially change.

damianloew: Above that, maybe it could be good to check whether members didn't change after each bsearch() call

libc/bsearch.c
```diff
@@ -0,0 +1,213 @@
...
+
+	/* check for odd number of elements */
+	for (i = 0; i < TABSIZE - 1; i++) {
```
damianloew, line 144: If I understand correctly the order of elements could be changed, so it's not certain that the last member equals `names[TABSIZE]`

libc/bsearch.c
```diff
@@ -0,0 +1,213 @@
...
+	TEST_ASSERT_EQUAL_INT(1, check_tab(names));
+}
+
+
+TEST(bsearch, find_node_not_in)
+{
+	struct node *rval, node;
+	char name[8];
+
+	TEST_ASSERT_EQUAL_MEMORY(name, strcpy(name, "phoenix"), sizeof(name));
```
damianloew, line 162: We could just pass "notExistentMember" as key.

libc/bsearch.c
```diff
@@ -0,0 +1,213 @@
...
+	/* check for odd number of elements */
+	rval = (struct node *)bsearch((void *)&node, (void *)table,
+		(size_t)TABSIZE - 1, sizeof(struct node), node_cmp);
+	TEST_ASSERT_NULL(rval);
+
+	TEST_ASSERT_EQUAL_INT(1, check_tab(names));
+}
+
+
+TEST(bsearch, not_sorted)
```
damianloew, line 181: We could also look for all the elements there.

libc/bsearch.c
```diff
@@ -0,0 +1,213 @@
...
+
+
+TEST_GROUP_RUNNER(bsearch)
+{
+	RUN_TEST_CASE(bsearch, zero_el);
+	prepare(names);
+	RUN_TEST_CASE(bsearch, find_node_in);
+	RUN_TEST_CASE(bsearch, find_node_not_in);
+	prepare(names_nsort);
+	RUN_TEST_CASE(bsearch, not_sorted);
```
damianloew, line 212: It's not checked:
Screenshot from 2023-06-07 18-29-45

Two or more members comapre equal - please add such case.

damianloew (changes requested):
We could certainly add cases, where key element meets the `compar` requirements partially, so for example when we decided to use simple strcmp() we would want to have such key: "bc\0d", when the members on `base` address are "abc", "def", and so on. We may want to make sure that members are correctly splitted and passed to the `compar` function.

libc/bsearch.c
adamdebek: To do this i would need to prepare test_table per every elements which is unnecessary effort in my opinion.

libc/stdlib_bsearch.c
```diff
@@ -0,0 +1,320 @@
...
+{
+	int i;
+
+	for (i = 0; i < TABSIZE; i++) {
+		test_table[i].string = (char *)malloc(strlen(names_arg[i]) + 1);
+
+		TEST_ASSERT_NOT_NULL(test_table[i].string);
+
+		strcpy(test_table[i].string, names_arg[i]);
+		test_table[i].length = strlen(names_arg[i]);
```
damianloew, line 85: Please call strlen one time in the iteration.

libc/stdlib_bsearch.c
```diff
@@ -0,0 +1,320 @@
...
+
+/* compare function used by bsearch */
+static int test_nodeCmp(const void *node1, const void *node2)
+{
+	return strcmp(((struct node *)node1)->string, ((struct node *)node2)->string);
+}
+
+
+/* compare function used by bsearch in special_node testcase */
+static int test_nodeCmp2(const void *node1, const void *node2)
```
damianloew, line 131: Please use more distinctive names.

```suggestion
static int test_nodeCmpStr(const void *node1, const void *node2)
{
	return strcmp(((struct node *)node1)->string, ((struct node *)node2)->string);
}


/* compare function used by bsearch in special_node testcase */
static int test_nodeCmpLen(const void *node1, const void *node2)
```

libc/stdlib_bsearch.c
```diff
@@ -0,0 +1,320 @@
...
+	char name[] = "adam";
+	node2.string = name;
+	node2.length = strlen(name);
+
+	/* Check for even number of elements */
+	test_prepare(names);
+
+	rval = (struct node *)bsearch((void *)&node2, (void *)test_table,
+		(size_t)TABSIZE, sizeof(struct node), test_nodeCmp2);
+	TEST_ASSERT_NULL(rval);
```
damianloew, line 305: Please explain this case a little bit in a comment, the best would be o provide and example, so sth like: 

```C
/* the table content is {str1len1str2len2str3len3}, so we assert that elements are splitted properly and len1str2 won't be found */
```

damianloew (changes requested):

damianloew (approved):

PR #199, opened 2023-04-28
libc: string.h: add cmp functions tests

maska989:
## Description
Added 
 *    memcmp(),
 *    strcmp(),
 *    strncmp(),
 *    strcoll()
 
function tests.

## Motivation and Context
JIRA: CI-231

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/string/string_cmp.c
+++ b/libc/string/string_cmp.c
@@ -1,0 +1,544 @@
+/*
+ * Phoenix-RTOS
+ *
+ * POSIX.1-2017 standard library functions tests
+ * HEADER:
+ *    - string.h
+ * TESTED:
+ *    - memcmp()
+ *    - strcmp()
+ *    - strncmp()
+ *    - strcoll()
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Damian Modzelewski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+
+#include <stdio.h>
+#include <errno.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <unity_fixture.h>
+
+#include "testdata.h"
+
+#define BUFF_SIZE 129
+#define BIG_SIZE  1024
+
+static char empty[BUFF_SIZE];
+
+TEST_GROUP(string_memcmp);
+TEST_GROUP(string_strncmp);
+TEST_GROUP(string_strcmp);
+TEST_GROUP(string_strcoll);
+
+
+/*
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+*/
+
+
+TEST_SETUP(string_memcmp)
+{
+}
+
+
+TEST_TEAR_DOWN(string_memcmp)
+{
+}
+
+
+TEST(string_memcmp, basic)
+{
+	const char *basicStr = "Test";
+
+	TEST_ASSERT_EQUAL_INT(0, memcmp(basicStr, basicStr, 5));
+	TEST_ASSERT_LESS_THAN_INT(0, memcmp(basicStr, "Tests", 5));
+	TEST_ASSERT_GREATER_THAN_INT(0, memcmp("Tests", basicStr, 5));
+}
+
+
+TEST(string_memcmp, unsigned_char_cast)
+{
+	const signed char charEdgeValue = SCHAR_MIN;
+
+	TEST_ASSERT_GREATER_THAN_INT(0, memcmp(&charEdgeValue, "\0", 1));
+	TEST_ASSERT_LESS_THAN_INT(0, memcmp("\0", &charEdgeValue, 1));
+}
+
+
+TEST(string_memcmp, emptyInput)
+{
+	char *asciiSet = testdata_createCharStr(BUFF_SIZE),
+		 separated[] = "\0\0\0\0\0TEST\0\0";
+
+	asciiSet[0] = 0;
+
+	TEST_ASSERT_LESS_THAN_INT(0, memcmp(empty, asciiSet, sizeof(empty)));
+	TEST_ASSERT_GREATER_THAN_INT(0, memcmp(asciiSet, empty, sizeof(empty)));
+	TEST_ASSERT_EQUAL_INT(0, memcmp(empty, empty, sizeof(empty)));
+
+	/* Memory cmp is not sensitive for NUL characters */
+	TEST_ASSERT_NOT_EQUAL_INT(0, memcmp(empty, separated, sizeof(separated)));
+
+	free(asciiSet);
+}
+
+
+TEST(string_memcmp, big)
+{
+	char *hugeStr = testdata_createCharStr(BIG_SIZE),
+		 hugeStr2[BIG_SIZE];
+
+	memcpy(hugeStr2, hugeStr, BIG_SIZE);
+
+	TEST_ASSERT_EQUAL_INT(0, memcmp(hugeStr, hugeStr, BIG_SIZE));
+	/* Comparing the same strings, that are placed in different location */
+	TEST_ASSERT_EQUAL_INT(0, memcmp(hugeStr, hugeStr2, BIG_SIZE));
+
+	hugeStr[BIG_SIZE - 2] = 1;
+	hugeStr2[sizeof(hugeStr2) - 2] = 2;
+
+	TEST_ASSERT_LESS_THAN_INT(0, memcmp(hugeStr, hugeStr2, BIG_SIZE));
+	TEST_ASSERT_GREATER_THAN_INT(0, memcmp(hugeStr2, hugeStr, BIG_SIZE));
+
+	free(hugeStr);
+}
+
+
+TEST(string_memcmp, various_sizes)
+{
+	int i;
+
+	char *asciiStr = testdata_createCharStr(BUFF_SIZE),
+		 asciiStr2[BUFF_SIZE];
+
+	memcpy(asciiStr2, asciiStr, sizeof(asciiStr2));
+	for (i = 1; i < BUFF_SIZE - 1; i++) {
+		asciiStr2[i] = asciiStr[i] - 1;
+		TEST_ASSERT_EQUAL_INT(0, memcmp(asciiStr, asciiStr2, i));
+		TEST_ASSERT_GREATER_THAN_INT(0, memcmp(asciiStr, asciiStr2, i + 1));
+		asciiStr2[i] = asciiStr[i];
+	}
+
+	free(asciiStr);
+}
+
+
+TEST(string_memcmp, offsets)
+{
+
+	char dataSet[4000] = { 0 },
+		 supportSet[4000] = { 0 };
+	int s1Offs, s2Offs, szOffset, sz = sizeof(dataSet);
+
+	for (s1Offs = 0; s1Offs < sz; s1Offs++) {
+		dataSet[s1Offs] = s1Offs;
+		supportSet[s1Offs] = s1Offs;
+	}
+
+	/*Testing different offset of data blocks with same space or different*/
+	for (s1Offs = 0; s1Offs < 8; s1Offs++) {
+		for (s2Offs = 0; s2Offs < 8; s2Offs++) {
+			for (szOffset = 0; szOffset < 8; szOffset++) {
+
+				if (s2Offs < s1Offs) {
+					TEST_ASSERT_GREATER_THAN_INT(0, memcmp(&dataSet[s1Offs], &dataSet[s2Offs], sz - (s1Offs + szOffset)));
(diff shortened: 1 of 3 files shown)


libc/string_cmp.c
```diff
@@ -0,0 +1,127 @@
...
+{
+}
+
+
+TEST_TEAR_DOWN(string_cmp)
+{
+}
+
+
+TEST(string_cmp, cmp)
```
damianloew, line 49: Only a quick look: please split this test to a few cases and make sure that there are no more edge cases/things to check.

mateusz-bloch (changes requested):

libc/string_cmp.c
```diff
@@ -0,0 +1,238 @@
...
+
+
+/*
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+*/
+
+
+TEST(string_cmp, strncmp_basic)
+{
+	TEST_ASSERT_EQUAL_INT(0, strncmp(asciiSet, asciiSet, BUFF_SIZE));
```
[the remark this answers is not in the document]
maska989: It was intentional as you can see in one line after this we used asciiSet and nonZeroAsciiSet.
This case supposes to force *cmp to compare elements that are the same and we try to catch if it will assign data correctly.

libc/string_cmp.c
```diff
@@ -0,0 +1,238 @@
...
+	RUN_TEST_CASE(string_cmp, strncmp_empty);
+	RUN_TEST_CASE(string_cmp, strncmp_sizes);
+
+	RUN_TEST_CASE(string_cmp, strcmp_basic);
+	RUN_TEST_CASE(string_cmp, strcmp_empty);
+	RUN_TEST_CASE(string_cmp, strcmp_sizes);
+
+	RUN_TEST_CASE(string_cmp, strcoll_basic);
+	RUN_TEST_CASE(string_cmp, strcoll_empty);
+	RUN_TEST_CASE(string_cmp, strcoll_sizes);
```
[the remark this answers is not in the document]
maska989: Those cases check asciiStr where are present null terms.

libc/string_cmp.c
```diff
@@ -0,0 +1,431 @@
...
+
+
+/*
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+*/
+
+
+TEST_SETUP(string_strcoll)
+{
+	errno = 0;
```
damianloew, line 323: Why only in tests of strcoll function? Unfortunately it's not certain that errno won't change during the execution of some internal functions:
Screenshot from 2023-06-27 16-03-55

damianloew (changes requested):
Thanks for correcting the test to this version, it looks clear, but let's improve it a little bit.

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+
+		/*
+		 * The testing compares elements that value are every time on a different spot
+		 * and will check if the placement of the dataset on s1 and s2 will not impact on this
+		 */
+		support[i] = 1;
+		TEST_ASSERT_LESS_THAN_INT(0, memcmp(empty, support, BUFF_SIZE));
+		TEST_ASSERT_GREATER_THAN_INT(0, memcmp(support, empty, BUFF_SIZE));
+		support[i] = 0;
+	}
```
damianloew, line 137: Not tested, but this approach should be more effective.

```suggestion
	/* +1 to provide all ascii characters (ChrStr contains double 1 at the beginning) */
	const char *asciiStr = testdata_createCharStr(BUFF_SIZE + 1); //not sure whether const is possible there
	char asciiStr[BUFF_SIZE +1];
		
	memcpy(asciiStr2, asciiStr, sizeof(asciiStr));

	for (i = 1; i < BUFF_SIZE; i++) {
		asciiStr2[i] = asciiStr[i] - 1;
		TEST_ASSERT_EQUAL_INT(0, memcmp(asciiStr, asciiStr2, i));
		TEST_ASSERT_GREATER_THAN_INT(0, memcmp(asciiStr, asciiStr2, i + 1));
		asciiStr2[i] = asciiStr[i];
	}
	
	free(asciiStr);
	
```

 - you want to test especially cases, where different size changes the result of memcmp, that's why the approach above had been suggested.
 - comment not needed as it should be obvious that we check various sizes in such case,
 - there is no problem to use ascii string with NUL term, the exact data does not matter there, so I'd name those variables asciiStr
 - local variables seems to be better there, because we see the whole flow of data within them.

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+TEST_GROUP(string_strncmp);
+TEST_GROUP(string_strcmp);
+TEST_GROUP(string_strcoll);
+
+/*
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+*/
+
+
+TEST_SETUP(string_memcmp)
```
damianloew, line 54: Thanks to the changes below you won't need most of the code there.

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+
+static char empty[BUFF_SIZE] = { 0 },
+			support[BUFF_SIZE] = { 0 },
+			hugeZeroString[BUFF_SIZE * 4] = { 0 },
+			separated[24] = { 0, 0, 0, 0, 0, 0, 'T', 'E', 'S', 'T', 0, 0, 0 },
+			charEdgeValue[1] = { 0 },
+			*asciiSet,
+			*hugeString,
+			*nonZeroAsciiSet,
+			basicStr[6] = "Basic";
```
damianloew, line 42: Thanks to the suggestions below you shouldn't need most of variables there.

libc/string_cmp.c
```diff
@@ -0,0 +1,431 @@
...
+TEST(string_memcmp, sizes)
+{
+	int i;
+
+	/* Memory cmp is not sensitive for NUL characters */
+	TEST_ASSERT_NOT_EQUAL_INT(0, memcmp(empty, separated, sizeof(separated)));
+
+
+	/* Loop which changes the size of compared space to place where asciSet ends */
+	for (i = 1; i < sizeof(asciiSet); i++) {
```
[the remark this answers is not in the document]
damianloew: It was not applied, I've suggested sth similar, but with the ascii data and with checking two sizes in one iteration.

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+	}
+}
+
+
+/*
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+*/
+
+
+TEST_SETUP(string_strncmp)
```
damianloew, line 181: Please apply the suggestions above to the rest of functions too, in case of doubts please let me know.

damianloew (changes requested):

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+TEST(string_memcmp, unsigned_char_cast)
+{
+	/*
+	 * Checking do memcmp convert elements of input to unsigned char
+	 * If we get a value less than 0 that means memcmp treats the value of s1 as
+	 * signed char and return diff between SCHAR_MIN and 0
+	 * otherwise it will calculate the difference between 254 and 0 and memcmp will return
+	 * negative value
+	 */
+	charEdgeValue[0] = SCHAR_MIN;
```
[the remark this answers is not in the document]
maska989: In memcmp it will pass but with str* it throws warnings about the wrong signature that's why i used only char without signed prefix.

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+		 *nonZeroAsciiStr = testdata_createCharStr(BUFF_SIZE);
+	asciiStr[0] = 0;
+
+	TEST_ASSERT_EQUAL_INT(0, strcmp(empty, asciiStr));
+	TEST_ASSERT_EQUAL_INT(0, strcmp(asciiStr, empty));
+	TEST_ASSERT_LESS_THAN_INT(0, strcmp(empty, nonZeroAsciiStr));
+	TEST_ASSERT_GREATER_THAN_INT(0, strcmp(nonZeroAsciiStr, empty));
+
+	free(asciiStr);
+	free(nonZeroAsciiStr);
```
damianloew, line 313:
```suggestion
	char *asciiStr = testdata_createCharStr(BUFF_SIZE),

	TEST_ASSERT_LESS_THAN_INT(0, strcmp(empty, asciiStr));
	TEST_ASSERT_GREATER_THAN_INT(0, strcmp(asciiStr, empty));

	free(asciiStr);
```

damianloew: What about `strcmp(empty, empty)`?

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+	TEST_ASSERT_EQUAL_INT(0, strcoll(hugeStr, hugeStr));
+	/* Comparing the same strings, that are placed in different location */
+	TEST_ASSERT_EQUAL_INT(0, strcoll(hugeStr, hugeStr2));
+	hugeStr[BIG_SIZE - 2] = 1;
+	hugeStr2[sizeof(hugeStr2) - 2] = 2;
+	TEST_ASSERT_LESS_THAN_INT(0, strcoll(hugeStr, hugeStr2));
+	TEST_ASSERT_GREATER_THAN_INT(0, strcoll(hugeStr2, hugeStr));
+	hugeStr[sizeof(hugeStr2) - 2] = 0;
+	TEST_ASSERT_LESS_THAN_INT(0, strcoll(hugeStr, hugeStr2));
+	TEST_ASSERT_GREATER_THAN_INT(0, strcoll(hugeStr2, hugeStr));
```
damianloew, line 396:
```suggestion
```

damianloew (changes requested):

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+}
+
+
+TEST(string_strncmp, basic)
+{
+	const char *basicStr = "Basic";
+
+	TEST_ASSERT_EQUAL_INT(0, strncmp(basicStr, basicStr, 6));
+	TEST_ASSERT_GREATER_THAN_INT(0, strncmp(basicStr, support, 6));
+	TEST_ASSERT_LESS_THAN_INT(0, strncmp(support, basicStr, 6));
```
damianloew, line 188: I think you can also use "Test" or sth like that and we could remove `support` buff.

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+
+#include "testdata.h"
+
+
+#define BUFF_SIZE 129
+#define BIG_SIZE  1024
+
+static char empty[BUFF_SIZE] = { 0 },
+			support[BUFF_SIZE] = { 0 },
+			separated[24] = { 0, 0, 0, 0, 0, 0, 'T', 'E', 'S', 'T', 0, 0, 0 };
```
damianloew, line 37: There are only 2/3 calls now, we can safely change them to using "\0\0\0\0\0TEST\0\0".

```suggestion
```

libc/string/string_cmp.c
```diff
@@ -0,0 +1,439 @@
...
+#include <unistd.h>
+#include <unity_fixture.h>
+
+#include "testdata.h"
+
+
+#define BUFF_SIZE 129
+#define BIG_SIZE  1024
+
+static char empty[BUFF_SIZE] = { 0 },
```
damianloew, line 35: It's y default initialized to 0, when it's global buffer.

```suggestion
static char empty[BUFF_SIZE];
```

damianloew (changes requested):

libc/string/string_cmp.c
```diff
@@ -0,0 +1,544 @@
...
+
+	TEST_ASSERT_GREATER_THAN_INT(0, memcmp(&charEdgeValue, "\0", 1));
+	TEST_ASSERT_LESS_THAN_INT(0, memcmp("\0", &charEdgeValue, 1));
+}
+
+
+TEST(string_memcmp, emptyInput)
+{
+	char *asciiSet = testdata_createCharStr(BUFF_SIZE),
+		 separated[24] = "\0\0\0\0\0TEST\0\0";
```
damianloew, line 79: IMHO we should choose one of the following options, because the current one may be misleading:

 - separated[] = "\0\0\0\0\0TEST\0\0\0\0\0\0\0\0\0\0\0\0\0\0"; // sttring with size 24 for init
 - separated[] = "\0\0\0\0\0TEST\0\0"; //current sttring for init
 
I think the second one would be sufficient.

damianloew (changes requested):
Small thing, but I think we should be consistent and start with a lowercase letter after suffix in commit message, so `testdata: extend description for createCharStr`

damianloew (approved):
LGTM, Let's wait with merge for https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/230

mateusz-bloch (approved):

PR #200, opened 2023-04-28
nightly tests: fix ports tests

jsarzy:
## Description
Activate tests that can be run on targets that are build using the "ports" argument. If specific ports are not built by default (e.g embtls) then target is not included in yaml config. Fix a few things in micopython harness.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: emulated targets

## Checklist:

## Special treatment

### DIFF
--- a/micropython/micropython.py
+++ b/micropython/micropython.py
@@ -1,12 +1,12 @@
 import re
 from pathlib import Path
 
+from trunner.ctx import TestContext
+from trunner.dut import Dut
 from trunner.types import Status, TestResult
 
-PROMPT = r"(\r+)\x1b\[0J" + r"\(psh\)% "
 EOL = r"\r+\n"
 
-UPYTH_PROMPT = ">>>"
 MICROPYTHON = "/bin/micropython"
 
 TESTS_WITH_REGEX_OUTPUT = {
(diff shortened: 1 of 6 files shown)


micropython/micropython.py
```diff
@@ -64,11 +66,17 @@ def is_test_with_regex_output(test_path: str):
...
-def get_test(dut):
-    dut.expect(PROMPT, timeout=45)
-    output = dut.before
+def remove_ansi_escape(text: str):
+    ansi_escape = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
+    return ansi_escape.sub("", text)
+
+
+def get_test(dut: Dut, ctx: TestContext):
+    dut.expect_exact(ctx.target.shell_prompt, timeout=10)
```
damianloew, line 75: There are no problems with such short timeout instead of 45s?

micropython/micropython.py
```diff
@@ -64,11 +66,17 @@ def is_test_with_regex_output(test_path: str):
...
+    output = remove_ansi_escape(dut.before)
```
damianloew, line 76: Why those escape sequences are so problematic?


Did you mean ascii?
```suggestion
    output = remove_ascii_escape(dut.before)
```

micropython/micropython.py
```diff
@@ -64,11 +66,17 @@ def is_test_with_regex_output(test_path: str):
     return p.parent.name + "/" + p.name in TESTS_WITH_REGEX_OUTPUT
 
 
-def get_test(dut):
-    dut.expect(PROMPT, timeout=45)
-    output = dut.before
+def remove_ansi_escape(text: str):
+    ansi_escape = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
```
damianloew, line 70: Maybe according to control sequence introducer pattern:

```suggestion
    ascii_escape = re.compile(r"(\x1b\[[\x30-\x3F]*[\x20-\x2F]*[\x40-\x7E])")
```

damianloew (changes requested):
A few small suggestions. Overall looks good, thanks for fixing it :+1:

damianloew (changes requested):
A few small suggestions. Overall looks good, thanks for fixing it :+1:

micropython/micropython.py
damianloew: Ok, let's leave it.

micropython/micropython.py
```diff
@@ -1,9 +1,11 @@
 import re
 from pathlib import Path
 
+from trunner.ctx import TestContext
+from trunner.dut import Dut
 from trunner.types import Status, TestResult
 
-PROMPT = r"(\r+)\x1b\[0J" + r"\(psh\)% "
+PROMPT = r"\r*\x1b\[0J" + r"\(psh\)% "
```
damianloew, line 8: Is it still needed?

micropython/micropython.py
jsarzy: Changed to removing only last characters

damianloew (approved):

PR #202, opened 2023-05-02
enable test-libc on armv7m4-stm32l4x6-nucleo

mateusz-bloch:
- Changed directory structure in libc to make it easier to split tests into separate binaries so that tests can be run on nucleo.

Before:
image

After:
image

- Enabled all libc nucleo tests

- JIRA CI-268

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


libc/rest/Makefile
```diff
@@ -0,0 +1,10 @@
+LOCAL_DIR := $(call my-dir)
+PARENT_DIR := $(shell dirname $(realpath $(LOCAL_DIR)))
```
damianloew, line 2: I think it could be better

```suggestion
PARENT_DIR = $(shell dirname $(realpath $(LOCAL_DIR)))
```

damianloew (commented):
IMHO looks good, I proposed such structure and I think it will be the best for the current and incoming tests. @nalajcie do you think this Makefile is ok, or could we do it in a simpler way?

libc/rest/Makefile
nalajcie: nah, always use `:=` in makefiles if possible (this makes variable value to be computed once instead at every usage).

Buut... the `Makefile` should not try compiling anything in `..`, the common files should probably be made into static lib to be reused in sub-components if needed.

mateusz-bloch: @nalajcie So, do you think the name `libtest` would be adequate for this static library?

damianloew (approved):
It seems to be be much simpler now, @nalajcie thanks for your suggestion. In my opinion we can merge this version

libc/test.yaml
```diff
@@ -1,10 +1,35 @@
...
-      execute: test-libc
-      #FIXME: symlink tests will fail on non-rootfs targets
-      #FIXME: unistd_uid doesn't work on zynq7000-qemu
+    - name: printf
+      execute: test-libc-printf
       targets:
         exclude: [armv7m4-stm32l4x6-nucleo]
         include: [host-generic-pc]
+
+    # Executing test-printf on 'armv7m4-stm32l4x6-nucleo' takes ~9 min, that's why we want to run it only during nightly campaigns
```
nalajcie, line 10: uh, could You elaborate what takes so much time at this target (i would like to be sure it's not a problem with `printf` implementation)

damianloew (dismissed):

libc/test.yaml
mateusz-bloch: Of course, I'm almost certain that it's not the fault of the printf implementation, the test itself is quite extensive, but I will try to check as soon as possible what exactly causes this test to take so long to run on stm32l4x6-nucleo.

libc/Makefile
```diff
@@ -1,9 +1,23 @@
...
+SRCS := $(wildcard $(MY_LOCAL_DIR)$(1)/*.c) $(COMMON_SRCS)
+DEP_LIBS := unity
+LOCAL_CFLAGS := -I$(MY_LOCAL_DIR)
+LOCAL_LDFLAGS := $(2)
 include $(binary.mk)
+endef
+
+$(eval $(call add_test_libc,printf, -z stack-size=12288))
+$(eval $(call add_test_libc,pthread, -lpthread -z stack-size=12288))
+$(eval $(call add_test_libc,rest))
```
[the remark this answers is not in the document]
mateusz-bloch: agree, `misc` seems to be a much better name

libc/test.yaml
```diff
@@ -1,10 +1,35 @@
...
-      execute: test-libc
-      #FIXME: symlink tests will fail on non-rootfs targets
-      #FIXME: unistd_uid doesn't work on zynq7000-qemu
+    - name: printf
+      execute: test-libc-printf
       targets:
         exclude: [armv7m4-stm32l4x6-nucleo]
         include: [host-generic-pc]
+
+    # Executing test-printf alone on 'armv7m4-stm32l4x6-nucleo' takes ~4 min, that's why we want to run it only during nightly campaigns
```
damianloew, line 10:
```suggestion
    # Executing test-printf alone on 'armv7m4-stm32l4x6-nucleo' takes ~4 min, because of custom pexpect send method used on this target and large amount of test cases. 
    # That's why we want to run it only during nightly campaigns 
```

damianloew (commented):
I hope it will be clear now

libc/test.yaml
damianloew: Ok, I see that, not so many commands are sent in this test, so it shouldn't be the main factor.

damianloew: Yeah, sth disturbing happens there, maybe the process of loading a binary using `restore` command in Jlink takes a long time. @mateusz-bloch please investigate it before a merge.

nalajcie: - the CI overhead is worrying and should be investigating further
- still, 4mins for the test is also quite a long time, we should probably investigate it further (can be done at a later time)

damianloew: Yeah, it's not a problem with this PR. In general it seems that all tests are executed much longer on stm target (for example comparing to imxrt106x), the results of investigation will be presented on Confluence.

damianloew (approved):

PR #206, opened 2023-05-12
trunner: pass not aligned size to plo alias cmd

damianloew:
## Description

trunner: pass not aligned size to plo alias cmd

## Motivation and Context

JIRA: CI-276

It's not needed, applied after the corresponding change in plo scripts used in project (https://github.com/phoenix-rtos/phoenix-rtos-project/commit/fb5e36f8cb0a80cb12e4e154ca9883bd0093d18a)

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/target/armv7m4.py
+++ b/trunner/target/armv7m4.py
@@ -65,7 +65,7 @@ class STM32L4x6PloAppLoader(TerminalHarness, PloInterface):
         self.ram_addr = 0x20000000
         self.page_sz = 0x200
 
-    def _app_size(self, path: Path):
+    def _aligned_app_size(self, path: Path):
         sz = path.stat().st_size
         # round up to the size of the page
         offset = (self.page_sz - sz) % self.page_sz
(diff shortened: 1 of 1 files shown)


trunner/target/armv7m4.py
```diff
@@ -93,7 +86,7 @@ def __call__(self):
         offset = self.load_offset
         for app in self.apps:
             path = self.gdb.cwd / Path(app.file)
-            sz = self._app_size(path)
+            sz = path.stat().st_size
```
anglov, line 89: in line 91  it;s good, but in line 94 you need to align to page

anglov (changes requested):

trunner/target/armv7m4.py
damianloew: Oh right, sorry for that, it's now fixed.

anglov: I can't add comment in place, because is too far in context, but still issue, now in line 101.

Offset needs to be page aligned. Only size in alias command shouldn't be aligned to provide valid binary size to kernel.

anglov (changes requested):

trunner/target/armv7m4.py
damianloew: Yeah, now I see, thanks for being watchful. I hope it won't be a big issue to get file size two times (lines 96 and 101)

anglov: It's good enough

anglov (approved):
Thank you for this correction

PR #207, opened 2023-05-17
Maska989/TestStrStrPbrkTok

maska989:
Added

- strtok(),
- strtok_r(),
- strstr(),
- strpbrk()

function tests.

JIRA: CI-234

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/string/string_tokpbrkstr.c
+++ b/libc/string/string_tokpbrkstr.c
@@ -1,0 +1,1037 @@
+/*
+ * Phoenix-RTOS
+ *
+ * POSIX.1-2017 standard library functions tests
+ * HEADER:
+ * 	- string.h
+ * TESTED:
+ * 	- strpbrk()
+ * 	- strstr()
+ * 	- strtok()
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Damian Modzelewski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <unity_fixture.h>
+#include "testdata.h"
+
+#define ASCII_LENGTH          128
+#define EXTENDED_ASCII_LENGTH 256
+#define BUFFSIZE              24
+#define LOREM_IPSUM           "Lorem Ipsum Dolor"
+
+TEST_GROUP(string_tok);
+TEST_GROUP(string_tok_r);
+TEST_GROUP(string_str);
+TEST_GROUP(string_pbrk);
+
+static char *create_extAscii_set(void)
+{
+	int i;
+	char *hold = malloc(EXTENDED_ASCII_LENGTH);
+
+	for (i = 1; i < EXTENDED_ASCII_LENGTH; i++) {
+		hold[i - 1] = i;
+	}
+	hold[i - 1] = 0;
+
+	return hold;
+}
+
+
+TEST_SETUP(string_tok)
+{
+}
+
+
+TEST_TEAR_DOWN(string_tok)
+{
+}
+
+
+TEST(string_tok, basic)
+{
+	char str[] = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas eleifend elementum tellum.";
+	const char *separators = ", .";
+	char *token;
+
+	TEST_ASSERT_EQUAL_PTR(str, (token = strtok(str, separators)));
+	TEST_ASSERT_EQUAL_STRING("Lorem", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[6], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("ipsum", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[12], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("dolor", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[18], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("sit", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[22], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("amet", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[28], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("consectetur", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[40], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("adipiscing", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[51], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("elit", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[57], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("Maecenas", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[66], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("eleifend", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[75], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("elementum", token);
+
+	TEST_ASSERT_EQUAL_PTR(&str[85], (token = strtok(NULL, separators)));
+	TEST_ASSERT_EQUAL_STRING("tellum", token);
+
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, separators));
+}
+
+
+/* assert that looking for tokens stops after encountering even one NUL term */
+TEST(string_tok, torn)
+{
+	char str[] = "Lor\0em ipsum";
+	const char *separators = " ";
+	char *tok;
+	TEST_ASSERT_EQUAL_PTR(str, (tok = strtok(str, separators)));
+	TEST_ASSERT_EQUAL_STRING("Lor", tok);
+	TEST_ASSERT_EQUAL_PTR(NULL, (tok = strtok(NULL, separators)));
+}
+
+
+TEST(string_tok, empty_args)
+{
+	/* Different scenarios of acquiring null pointer as output*/
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok("", "d"));
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, ""));
+
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok("\0", ""));
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, ""));
+
+	TEST_ASSERT_EQUAL_STRING("abc", strtok("abc", ""));
+}
+
+
+TEST(string_tok, multi_call)
+{
+	char loremStr[20] = LOREM_IPSUM;
+	char sep[5] = { 0 };
+	char multiCallStr[ASCII_LENGTH] = { 0 };
+	char *asciiStr = testdata_createCharStr(ASCII_LENGTH);
+	int i;
+
+	/*
+	 * In this case we are checking if it is possible to
+	 * pass different stop points in each call to
+	 * get desired tokens
+	 */
+
+	sprintf(sep, "%s", "ImDr");
+
+	TEST_ASSERT_EQUAL_PTR(loremStr, strtok(loremStr, "I"));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[7], strtok(NULL, "m"));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[11], strtok(NULL, "D"));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[13], strtok(NULL, "r"));
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, sep));
+
+	/*
+	 * strtok() changes the content of a string,
+	 * that's why we have to set it once again
+	 */
+	memcpy(loremStr, LOREM_IPSUM, sizeof(LOREM_IPSUM));
+
+	/*
+	 * After passing all delimiters we cannot avoid misplaced tokens because
+	 * strtok looks for all elements in "ImDr" and breaks a string
+	 * in points where the first occurrence found
+	 */
+	TEST_ASSERT_EQUAL_PTR(loremStr, strtok(loremStr, sep));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[3], strtok(NULL, sep));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[5], strtok(NULL, sep));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[7], strtok(NULL, sep));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[11], strtok(NULL, sep));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[13], strtok(NULL, sep));
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, sep));
+
+	/*
+	 * In this case, we tokenize the whole ASCII set using jump-over by
+	 * 2 elements. We are forced to jump over 2 places because strtok set
+	 *  divider placed as a null term and in this case, we are unable to
+	 * cover the whole set. Jumping over 2 places guarantees us output in the form
+	 * of one element form ascii set to do an assert with the place where
+	 * strtok has saved the stopping place
+	 */
+
+	memset(sep, 0, sizeof(sep));
+
+	TEST_ASSERT_EQUAL_PTR(&asciiStr[1], strtok(&asciiStr[1], "\2"));
+	for (i = 4; i < ASCII_LENGTH; i += 2) {
+		sep[0] = i;
+		TEST_ASSERT_EQUAL_PTR(&asciiStr[i - 1], strtok(NULL, &sep[0]));
+	}
+
+	/* Checking if set is empty */
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, &sep[0]));
+
+	/* Creating string with similar pattern */
+	for (i = 0; i < sizeof(multiCallStr) - 1; i++) {
+		multiCallStr[i] = i % 3 + 1;
+	}
+
+	/* Preparing separator with a bunch of same elements*/
+	memset(sep, 1, sizeof(sep) - 1);
+
+	/* Initial tokenization */
+	TEST_ASSERT_EQUAL_PTR(&multiCallStr[1], strtok(multiCallStr, sep));
+
+
+	/* Checking if tokenization is in the correct places*/
+	for (i = 4; i < sizeof(multiCallStr) - 3; i = i + 3) {
+		TEST_ASSERT_EQUAL_PTR(&multiCallStr[i], strtok(NULL, sep));
+	}
+
+	memset(multiCallStr, 2, sizeof(multiCallStr));
+	memset(&multiCallStr[sizeof(multiCallStr) / 2], 1, 5);
+	multiCallStr[sizeof(multiCallStr) - 1] = 0;
+
+	/* Checking ability to tokenize element with multiple same characters treated as a chain of the same element from sep*/
+	TEST_ASSERT_EQUAL_PTR(multiCallStr, strtok(multiCallStr, sep));
+	TEST_ASSERT_EQUAL_PTR(&multiCallStr[sizeof(multiCallStr) / 2 + 5], strtok(NULL, sep));
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, sep));
+
+
+	free(asciiStr);
+}
+
+
+TEST(string_tok, out_of_ascii)
+{
+	char divider[6] = { 0 };
+	char separator[ASCII_LENGTH];
+	char *extAsciiStr = create_extAscii_set();
+	int i;
+
+	divider[0] = extAsciiStr[0];
+	divider[1] = extAsciiStr[64];
+	divider[2] = extAsciiStr[128];
+	divider[3] = extAsciiStr[192];
+	divider[4] = extAsciiStr[254];
+
+
+	/* Checking ability for tokenize element above standard charset */
+	TEST_ASSERT_EQUAL_PTR(&extAsciiStr[1], strtok(extAsciiStr, divider));
+	TEST_ASSERT_EQUAL_PTR(&extAsciiStr[65], strtok(NULL, divider));
+	TEST_ASSERT_EQUAL_PTR(&extAsciiStr[129], strtok(NULL, divider));
+	TEST_ASSERT_EQUAL_PTR(&extAsciiStr[193], strtok(NULL, divider));
+
+	/* 256 is not taken as a token because on element after it is a null term and it is recognized as the end of set */
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, divider));
+
+	free(extAsciiStr);
+	extAsciiStr = create_extAscii_set();
+	/*
+	 * Checking available tokenization on all elements
+	 * from outside of ASCII charset one by one, with
+	 * the first separator set as the first element of the set
+	 */
+	TEST_ASSERT_EQUAL_PTR(extAsciiStr, strtok(extAsciiStr, "\128"));
+	for (i = ASCII_LENGTH + 1; i < sizeof(separator); i++) {
+		separator[i - ASCII_LENGTH] = i;
+		TEST_ASSERT_EQUAL_PTR(&extAsciiStr[i - ASCII_LENGTH], strtok(NULL, &separator[i - ASCII_LENGTH]));
+	}
+
+	free(extAsciiStr);
+}
+
+
+TEST(string_tok, big_string)
+{
+	char str[PATH_MAX];
+	char supp[PATH_MAX];
+
+	memset(str, 1, sizeof(str));
+
+	str[sizeof(str) - 2] = 2;
+	str[sizeof(str) - 1] = 0;
+
+	/* We must use supp in this place to suppress tokenization on this element*/
+	memcpy(supp, str, sizeof(str));
+
+	TEST_ASSERT_EQUAL_PTR(str, strtok(str, "\2"));
+	TEST_ASSERT_EQUAL_STRING("\2", strtok("\2", str));
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(str, supp));
+}
+
+
+/*
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
+*/
+
+
+TEST_SETUP(string_tok_r)
+{
+}
+
+
+TEST_TEAR_DOWN(string_tok_r)
+{
+}
+
+
+TEST(string_tok_r, basic)
+{
+	/* Ifdef used because lack of function 'strtok_r' */
+#ifdef __phoenix__
+	TEST_IGNORE();
+#else
+
+	/*
+	 * There needed to be two copies of the same element as strtok_r works
+	 * on the same space where the variable is settled when even using restrict
+	 */
+	char str1[] = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas eleifend elementum tellum.";
+	char *restState1 = NULL;
+	char *restState2 = NULL;
+	char *token;
+	char str2[sizeof(str1)];
+
+	const char *separators = ", .";
+
+	memcpy(str2, str1, sizeof(str1));
+
+	TEST_ASSERT_EQUAL_PTR(str1, (token = strtok_r(str1, separators, &restState1)));
+	TEST_ASSERT_EQUAL_STRING("Lorem", token);
+	TEST_ASSERT_EQUAL_PTR(str2, (token = strtok_r(str2, separators, &restState2)));
+	TEST_ASSERT_EQUAL_STRING("Lorem", token);
+	TEST_ASSERT_NOT_NULL(restState1);
+	TEST_ASSERT_NOT_NULL(restState2);
+
+	TEST_ASSERT_EQUAL_PTR(&str1[6], (token = strtok_r(NULL, separators, &restState1)));
+	TEST_ASSERT_EQUAL_STRING("ipsum", token);
+	TEST_ASSERT_EQUAL_PTR(&str2[6], (token = strtok_r(NULL, separators, &restState2)));
+	TEST_ASSERT_EQUAL_STRING("ipsum", token);
+	TEST_ASSERT_NOT_NULL(restState1);
+	TEST_ASSERT_NOT_NULL(restState2);
+
+	TEST_ASSERT_EQUAL_PTR(&str1[12], (token = strtok_r(NULL, separators, &restState1)));
+	TEST_ASSERT_EQUAL_STRING("dolor", token);
+	TEST_ASSERT_EQUAL_PTR(&str2[12], (token = strtok_r(NULL, separators, &restState2)));
+	TEST_ASSERT_EQUAL_STRING("dolor", token);
+	TEST_ASSERT_NOT_NULL(restState1);
+	TEST_ASSERT_NOT_NULL(restState2);
+
+	TEST_ASSERT_EQUAL_PTR(&str1[18], (token = strtok_r(NULL, separators, &restState1)));
+	TEST_ASSERT_EQUAL_STRING("sit", token);
+	TEST_ASSERT_EQUAL_PTR(&str2[18], (token = strtok_r(NULL, separators, &restState2)));
+	TEST_ASSERT_EQUAL_STRING("sit", token);
+	TEST_ASSERT_NOT_NULL(restState1);
+	TEST_ASSERT_NOT_NULL(restState2);
+
+	TEST_ASSERT_EQUAL_PTR(&str1[22], (token = strtok_r(NULL, separators, &restState1)));
+	TEST_ASSERT_EQUAL_STRING("amet", token);
+	TEST_ASSERT_EQUAL_PTR(&str2[22], (token = strtok_r(NULL, separators, &restState2)));
+	TEST_ASSERT_EQUAL_STRING("amet", token);
+	TEST_ASSERT_NOT_NULL(restState1);
+	TEST_ASSERT_NOT_NULL(restState2);
+
+	TEST_ASSERT_EQUAL_PTR(&str1[28], (token = strtok_r(NULL, separators, &restState1)));
+	TEST_ASSERT_EQUAL_STRING("consectetur", token);
+	TEST_ASSERT_EQUAL_PTR(&str2[28], (token = strtok_r(NULL, separators, &restState2)));
+	TEST_ASSERT_EQUAL_STRING("consectetur", token);
+	TEST_ASSERT_NOT_NULL(restState1);
+	TEST_ASSERT_NOT_NULL(restState2);
+
+	TEST_ASSERT_EQUAL_PTR(&str1[40], (token = strtok_r(NULL, separators, &restState1)));
+	TEST_ASSERT_EQUAL_STRING("adipiscing", token);
+	TEST_ASSERT_EQUAL_PTR(&str2[40], (token = strtok_r(NULL, separators, &restState2)));
+	TEST_ASSERT_EQUAL_STRING("adipiscing", token);
+	TEST_ASSERT_NOT_NULL(restState1);
+	TEST_ASSERT_NOT_NULL(restState2);
+
+	TEST_ASSERT_EQUAL_PTR(&str1[51], (token = strtok_r(NULL, separators, &restState1)));
+	TEST_ASSERT_EQUAL_STRING("elit", token);
+	TEST_ASSERT_EQUAL_PTR(&str2[51], (token = strtok_r(NULL, separators, &restState2)));
+	TEST_ASSERT_EQUAL_STRING("elit", token);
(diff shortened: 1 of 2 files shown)


libc/string_tokpbrkstr.c
```diff
@@ -0,0 +1,633 @@
...
+	TEST_ASSERT_EQUAL_CHAR_ARRAY(holder, asciiStr, ASCII_LENGTH / 2);
+	TEST_ASSERT_NOT_EQUAL_INT(0, memcmp(holder, asciiStr, ASCII_LENGTH));
+}
+
+
+TEST(string_tok, empty_input_output)
+{
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wrestrict"
+
```
mateusz-bloch, line 139: Describe why this pragma is needed. You can also do it in different way to not use pragma, but I don't think it's worth to do it.

libc/string_tokpbrkstr.c
```diff
@@ -0,0 +1,633 @@
...
+{
+}
+
+
+TEST_TEAR_DOWN(string_tok_r)
+{
+}
+
+
+TEST(string_tok_r, basic)
```
mateusz-bloch, line 260: Warnings for all test cases with strtok_r

You can omit it with:
```
#ifdef __phoenix__
	TEST_IGNORE();
#else
	/SOME CODE/
#endif
```

libc/string_tokpbrkstr.c
```diff
@@ -0,0 +1,633 @@
...
+	TEST_IGNORE();
+#endif
+
+	create_ascii_set(1);
+	asciiStr[ASCII_LENGTH / 2] = '\0';
+	char holder[ASCII_LENGTH],
+		*rest = NULL;
+
+	/* Last element of ascii string is \127 and between 0 and 127 was written 0*/
+	memcpy(holder, strtok_r(asciiStr, "\127", &rest), ASCII_LENGTH);
```
mateusz-bloch, line 324: Warnings: 
image

This warning is misleading, it's caused by lack of `strtok_r `on Phoenix-RTOS, the code itself is fine :)

libc/string_tokpbrkstr.c
```diff
@@ -0,0 +1,633 @@
...
+	RUN_TEST_CASE(string_str, out_of_ascii);
+}
+
+
+TEST_GROUP_RUNNER(string_pbrk)
+{
+	RUN_TEST_CASE(string_pbrk, basic);
+	RUN_TEST_CASE(string_pbrk, empty_input_output);
+	RUN_TEST_CASE(string_pbrk, out_of_ascii);
+}
```
mateusz-bloch, line 633: missing new-line

mateusz-bloch (changes requested):
I just took a quick look, this is not a full review

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,645 @@
...
+	TEST_ASSERT_EQUAL_PTR(loremStr, strtok(loremStr, "I"));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[7], strtok(NULL, "m"));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[11], strtok(NULL, "D"));
+	TEST_ASSERT_EQUAL_PTR(&loremStr[13], strtok(NULL, "r"));
+
+	/*
+	 * After using strtok on some elements we must
+	 * again write it to the array if we want to use
+	 * the same set because strtok breaks the elements in the string
+	 */
```
damianloew, line 182: Tip for the future, you can leave it now. We want to have concise comments, so in this case if you could write for example `strtok() changes the content of a string, that's why we have to set it once again`.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,645 @@
...
+{
+	char *holder, dividers[4];
+
+	/* Create an Ascii array that starts from 1*/
+	create_ascii_set(1);
+
+	sprintf(dividers, "%c%c%c", asciiStr[1], asciiStr[sizeof(asciiStr) / 2], asciiStr[sizeof(asciiStr) - 1]);
+
+	/* Checking if the return of strtok is a pointer equal to the input string pointer in the first tokenization */
+	TEST_ASSERT_EQUAL_PTR(asciiStr, strtok(asciiStr, &asciiStr[sizeof(asciiStr) - 1]));
```
damianloew, line 80: I am afraid it will be necessary to assert both a pointer and a content of returned string. This way we are sure that it returned a pointer to source string in the proper position, but you don't know whether characters didn't change there.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,645 @@
...
+	 * 2 elements. We are forced to jump over 2 places because strtok set
+	 *  divider placed as a null term and in this case, we are unable to
+	 * cover the whole set. Jumping over 2 places guarantees us output in the form
+	 * of one element form ascii set to do an assert with the place where
+	 * strtok has saved the stopping place
+	 */
+	TEST_ASSERT_EQUAL_PTR(asciiStr, strtok(asciiStr, "\2"));
+	for (i = 4; i < sizeof(asciiStr) - 1; i += 2) {
+		separator[0] = i;
+		TEST_ASSERT_EQUAL_PTR(&asciiStr[i - 2], strtok(NULL, &separator[0]));
```
damianloew, line 212: We could also make sure that next call would return NULL.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,645 @@
...
+
+TEST(string_tok, out_of_ascii)
+{
+	create_extAscii_set();
+
+	char divider[BUFFSIZE],
+		separator[ASCII_LENGTH];
+	int i;
+
+	sprintf(divider, "%c%c%c%c%c", extAsciiStr[0], extAsciiStr[64], extAsciiStr[128], extAsciiStr[192], extAsciiStr[255]);
```
damianloew, line 225: Please initialize it without using sprintf as I mentioned before.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,645 @@
...
+{
+}
+
+
+TEST_TEAR_DOWN(string_tok_r)
+{
+}
+
+
+TEST(string_tok_r, basic)
```
damianloew, line 264: Please apply suggestions from `strtok` tests.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,645 @@
...
+	 * cover the whole set. Jumping over 2 places guarantees us output in the form
+	 * of one element form ascii set to do an assert with the place where
+	 * strtok has saved the stopping place
+	 */
+	TEST_ASSERT_EQUAL_PTR(asciiStr, strtok(asciiStr, "\2"));
+	for (i = 4; i < sizeof(asciiStr) - 1; i += 2) {
+		separator[0] = i;
+		TEST_ASSERT_EQUAL_PTR(&asciiStr[i - 2], strtok(NULL, &separator[0]));
+	}
+}
```
damianloew, line 214: We could also add a check, where separator is a character from the beginning part of a string, which shouldn't be available in next calls, so for example, when you call:

```
char str[] = "abc";
TEST_ASSERT_EQUAL_PTR(str, strtok(str, "b"));
```
you want to:
```
TEST_ASSERT_EQUAL_PTR(&str[2], strtok(str, "a"));
```

libc/string/string_tokpbrkstr.c
damianloew: Above that I would check in the separate line whether in the specified position of str NUL character were placed.

damianloew: **Applies to the rest of strtok tests.**

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,645 @@
...
+#endif
+}
+
+
+/*
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
+*/
+
+
+TEST_SETUP(string_str)
```
damianloew, line 481: Please look at the `strstr` and `strpbrk` tests after applying the suggestions mentioned before and make sure that there is nothing to improve in a similar way.

damianloew (changes requested):
Let's improve these tests a little bit.

libc/string/string_tokpbrkstr.c
maska989: Those cases are checked with previous checks

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,695 @@
...
+
+TEST(string_tok_r, empty_args)
+{
+#ifdef __phoenix__
+	TEST_IGNORE();
+#else
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wrestrict"
+
```
mateusz-bloch, line 365: Why is this pragma necessary?

mateusz-bloch (changes requested):
I just took a quick look

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,695 @@
...
+	free(asciiStr);
+}
+
+
+TEST(string_tok, out_of_ascii)
+{
+	char *extAsciiStr = create_extAscii_set();
+
+	char divider[6] = { 0 },
+		 separator[ASCII_LENGTH];
```
mateusz-bloch, line 205: not sure if it's worth it, but if you want:
```suggestion
	char divider[6] = { 0 },
		 separator[ASCII_LENGTH];
		 *extAsciiStr = create_extAscii_set();
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,695 @@
...
+	/*
+	 * There needed to be two copies of the same element as strtok_r works
+	 * on the same space where the variable is settled when even using restrict
+	 */
+	char *restState1 = NULL,
+		 *restState2 = NULL;
+	char str1[] = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas eleifend elementum tellum.",
+		 str2[sizeof(str1)];
+	const char *separators = ", .";
+	char *token;
```
mateusz-bloch, line 270:
```suggestion
	char *restState1 = NULL,
		 *restState2 = NULL;
		  *token;
		  str1[] = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas eleifend elementum tellum.",
		  str2[sizeof(str1)];
	const char *separators = ", .";
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,695 @@
...
+#ifdef __phoenix__
+	TEST_IGNORE();
+#else
+
+	char *rest[ASCII_LENGTH] = { 0 },
+		 *tok;
+
+	int i;
+
+	char *asciiStr;
```
mateusz-bloch, line 510:
```suggestion
	char *rest[ASCII_LENGTH] = { 0 },
		 *tok;
		 *asciiStr;

	int i;
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+		separator[i - ASCII_LENGTH] = i;
+		TEST_ASSERT_EQUAL_PTR(&extAsciiStr[i - ASCII_LENGTH], strtok_r(NULL, &separator[i - ASCII_LENGTH], &rest));
+	}
+
+	free(extAsciiStr);
+#endif
+}
+
+
+TEST(string_tok_r, restriction)
```
damianloew, line 493: What is the intention of this test? Please leave a comment (the test case name does not show it sufficiently)

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+
+	char *rest[ASCII_LENGTH] = { 0 },
+		 *tok,
+		 *asciiStr = testdata_createCharStr(ASCII_LENGTH);
+
+	int i;
+
+	for (i = 1; i < ASCII_LENGTH - 1; i++) {
+		TEST_ASSERT_EQUAL_PTR(&asciiStr[1], tok = strtok_r(&asciiStr[1], &asciiStr[i + 1], &rest[i]));
+		TEST_ASSERT_EQUAL_CHAR_ARRAY(&asciiStr[1], tok, i);
```
damianloew, line 507: Maybe just:
```suggestion
		TEST_ASSERT_EQUAL_STRING(&asciiStr[1], tok);
```

This way you could make sure that NUL term is set correctly

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+		 *asciiStr = testdata_createCharStr(ASCII_LENGTH);
+
+	/* Standard use of strstr on arrays */
+	TEST_ASSERT_EQUAL_PTR(loremIpsum, strstr(loremIpsum, "Lorem"));
+	TEST_ASSERT_EQUAL_PTR(&loremIpsum[6], strstr(loremIpsum, "Ipsum"));
+	TEST_ASSERT_EQUAL_PTR(&loremIpsum[12], strstr(loremIpsum, "Dolor"));
+
+	/* Simple usage of strstr with ascii set */
+	TEST_ASSERT_EQUAL_PTR(asciiStr, strstr(asciiStr, asciiStr));
+	TEST_ASSERT_EQUAL_STRING(&asciiStr[ASCII_LENGTH / 4], strstr(asciiStr, &asciiStr[ASCII_LENGTH / 4]));
```
damianloew, line 543: Why EQUAL_STRING not EQUAL_PTR? Asserting pointer could be better, because it could return a same string in different location, which would be an error.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+
+	free(asciiStr);
+}
+
+
+TEST(string_str, empty_args)
+{
+	/* Different scenarios of using empty input or output*/
+	TEST_ASSERT_NOT_NULL(strstr("", ""));
+	TEST_ASSERT_NOT_NULL(strstr("abc", "\0"));
```
damianloew, line 558: You can assert the exact value (you would have to declare those strings)

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+	TEST_ASSERT_EQUAL_PTR(&loremIpsum[12], strstr(loremIpsum, "Dolor"));
+
+	/* Simple usage of strstr with ascii set */
+	TEST_ASSERT_EQUAL_PTR(asciiStr, strstr(asciiStr, asciiStr));
+	TEST_ASSERT_EQUAL_STRING(&asciiStr[ASCII_LENGTH / 4], strstr(asciiStr, &asciiStr[ASCII_LENGTH / 4]));
+	TEST_ASSERT_EQUAL_STRING(&asciiStr[ASCII_LENGTH / 3], strstr(asciiStr, &asciiStr[ASCII_LENGTH / 3]));
+	TEST_ASSERT_EQUAL_STRING(&asciiStr[ASCII_LENGTH / 2], strstr(asciiStr, &asciiStr[ASCII_LENGTH / 2]));
+
+	/* Passing an empty string as the second parameter to get first */
+	TEST_ASSERT_EQUAL_STRING(&asciiStr[ASCII_LENGTH / 2], strstr(&asciiStr[ASCII_LENGTH / 2], ""));
```
damianloew, line 548: It should be tested in empty case.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+}
+
+
+TEST(string_str, basic)
+{
+	char loremIpsum[BUFFSIZE] = LOREM_IPSUM,
+		 *asciiStr = testdata_createCharStr(ASCII_LENGTH);
+
+	/* Standard use of strstr on arrays */
+	TEST_ASSERT_EQUAL_PTR(loremIpsum, strstr(loremIpsum, "Lorem"));
```
damianloew, line 537: You could add passing the same string in both arguments.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+	int i;
+
+	char *extAsciiStr = create_extAscii_set();
+
+	for (i = 1; i < sizeof(extAsciiStr); i++) {
+		TEST_ASSERT_EQUAL_PTR(&extAsciiStr[i], strstr(extAsciiStr, &extAsciiStr[i]));
+	}
+
+	free(extAsciiStr);
+}
```
damianloew, line 577: You could also add cases, where `s2` is present partially in `s1` - from the left and from the right, so 
 
 - strstr("abcde", "def"))
 - strstr("abcde", "xab"))
 
or sth similar

Moreover, maybe it would be good to test big buffers passed to this function. Do you remember approach with PATH_MAX? You can also think about it for the rest of functions.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+{
+}
+
+
+TEST_TEAR_DOWN(string_pbrk)
+{
+}
+
+
+TEST(string_pbrk, basic)
```
damianloew, line 595: You can certainly add more cases to cover this funciton. First example would be strbrk("abc", "") For example:

- passing bytes in the `s2` in various order, so strpbrk("abcde", "dce")
- passing more bytes (than available in `s1`) in `s2`, so sth like: strpbrk("abcde", "cdefgh")
- big buffers
- a few same bytes in s2
- maybe some other cases

damianloew (changes requested):

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+
+
+TEST_TEAR_DOWN(string_tok)
+{
+}
+
+
+TEST(string_tok, basic)
+{
+	char str[] = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas eleifend elementum tellum.";
```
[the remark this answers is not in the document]
maska989: Sadly, this solution is incorrect because strtok works on those strings and edits them with the passed desired credentials.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+{
+#ifdef __phoenix__
+	TEST_IGNORE();
+#else
+
+	/*
+	 * There needed to be two copies of the same element as strtok_r works
+	 * on the same space where the variable is settled when even using restrict
+	 */
+	char *restState1 = NULL,
```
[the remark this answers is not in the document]
maska989: Added in cases not related with strtok and strtok_r

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+		 *restState2 = NULL,
+		 *token,
+		 str1[] = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas eleifend elementum tellum.",
+		 str2[sizeof(str1)];
+
+	const char *separators = ", .";
+
+	memcpy(str2, str1, sizeof(str1));
+
+	TEST_ASSERT_EQUAL_PTR(str1, (token = strtok_r(str1, separators, &restState1)));
```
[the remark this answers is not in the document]
maska989: Those variables are restrictions for strtok_r to do thread-safe tokenization and to don't mix up tokenizations between two elements similar to themselves. Asserting this value is pointless because it is possible to assert this as not null, but we discard this option using the next tokenization without a mixup.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,687 @@
...
+TEST(string_tok_r, empty_args)
+{
+#ifdef __phoenix__
+	TEST_IGNORE();
+#else
+	char empty[] = { 0 },
+		 *rest;
+
+	/* Different scenarios of acquiring null pointer as output*/
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok_r("", "d", &rest));
```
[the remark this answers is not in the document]
maska989: The definition of strtok_r says it must be a pointer that is not initialized to ensure it will be assigned to the correct user-defined place:
image
Application use also says:
image

libc/string/string_tokpbrkstr.c
maska989: This test case was created to check state restrictions about managing which tokenization has a place at this particular moment.

libc/string/string_tokpbrkstr.c
maska989: I tried different approaches in those cases, I will wait for a response about them.

libc/string/string_tokpbrkstr.c
damianloew: I don't quite understand, first call puts a NULL pointer in asciiStr[2], so each of next calls will return only "1" as a token, I don't see a purpose of asserting it 255 times.

I suppose you wanted to test mainly the rest there, which in fact is the same for all iterations (I don't see why only two of the last rests have to be equal)

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,875 @@
...
+TEST(string_str, empty_args)
+{
+	const char empty[] = "",
+			   str[] = "Lorem";
+
+
+	char *asciiStr = testdata_createCharStr(ASCII_LENGTH);
+
+	/* Different scenarios of using empty input or output*/
+	TEST_ASSERT_EQUAL_PTR(NULL, strstr(empty, str));
```
damianloew, line 627: `empty` not needed.

```suggestion
	TEST_ASSERT_EQUAL_PTR(NULL, strstr("", str));
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,875 @@
...
+	/* Passing an empty string as the second parameter to get first */
+	TEST_ASSERT_EQUAL_PTR(asciiStr, strstr(asciiStr, ""));
+
+	TEST_ASSERT_EQUAL_PTR(str, strstr(str, str));
```
damianloew, line 631: How is it related with empty args?

damianloew: Moreover I see that the similar case is already added in `basic`, so it's not needed to leave it.

libc/string/string_tokpbrkstr.c
damianloew: Where is this partially case? Could you add it if it's not added yet?

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,875 @@
...
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "cc"));
+	TEST_ASSERT_EQUAL_STRING("abc", strstr("abc", "ab"));
+	TEST_ASSERT_EQUAL_STRING("bc", strstr("abc", "bc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "ac"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "aabbcc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "aaaaaa"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bbbbbb"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "cccccc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "ccaabb"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bbaacc"));
```
damianloew, line 681: Those checks are not about order, I'd suggest adding an another test case for it.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,875 @@
...
+	free(asciiStr);
+}
+
+
+TEST(string_pbrk, empty_args)
+{
+	/* Different scenarios to acquire null pointer as output*/
+	TEST_ASSERT_NULL(strpbrk("", "abc"));
+	TEST_ASSERT_NULL(strpbrk("\0", "abc"));
+	TEST_ASSERT_NULL(strpbrk("\0", "\0"));
```
damianloew, line 755: Would be sufficient, if the first NUL term is detected, checking the next ones is not necessary.
```suggestion
	TEST_ASSERT_NULL(strpbrk("", ""));
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,875 @@
...
+	int i;
+	char str[PATH_MAX];
+
+	memset(str, 1, sizeof(str));
+
+	for (i = 1; i < PATH_MAX - 1; i++) {
+		str[i] = 2;
+		TEST_ASSERT_EQUAL_PTR(&str[i], strstr(str, "\2"));
+		TEST_ASSERT_EQUAL_PTR(str, strstr(str, "\1"));
+
```
damianloew, line 826: strstr in pbrk tests?

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,875 @@
...
+	 * the first separator set as the first element of the set
+	 */
+	TEST_ASSERT_EQUAL_PTR(extAsciiStr, strtok(extAsciiStr, "\128"));
+	for (i = ASCII_LENGTH + 1; i < sizeof(separator); i++) {
+		separator[i - ASCII_LENGTH] = i;
+		TEST_ASSERT_EQUAL_PTR(&extAsciiStr[i - ASCII_LENGTH], strtok(NULL, &separator[i - ASCII_LENGTH]));
+	}
+
+	free(extAsciiStr);
+}
```
damianloew, line 232: We could also create a case, where the separator string contains the same character a few times, same with `str`, it could contain the same characters multiple times. There is no `big` case, we can do it at once in such one.

damianloew (changes requested):

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,875 @@
...
+
+	/* Different scenarios of acquiring null pointer as output*/
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok("", "d"));
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, ""));
+
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(empty, ""));
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, ""));
+
+	TEST_ASSERT_EQUAL_PTR(str, strtok(str, empty));
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, empty));
```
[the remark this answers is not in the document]
maska989: I applied those changes but in suggestions is an error because, with the string passed by hand, we can't assert PTR and one line above instead of passing string as `""` i was forced to use `"\0"` because of restrictions in variables passed to strtok(GCC)

image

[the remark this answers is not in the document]
damianloew: oh, ok there is `restrict` in the declaration. Using "\0" seems to be fine in this case.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1019 @@
...
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "xbc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "xab"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "aabbcc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "aaaaaa"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bbbbbb"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "cccccc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "ccaabb"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bbaacc"));
+
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abcdefg", "hij"));
```
damianloew, line 808: They are just different strings, it does not fit partially.

```suggestion
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1019 @@
...
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("cdefg", "abc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abefg", "cde"));
```
damianloew, line 810: I'd remove it, `strstr("abc", "xab")` checks this scenario.

```suggestion
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1019 @@
...
+}
+
+
+TEST(string_str, big_string)
+{
+	int i;
+	char str[PATH_MAX];
+
+	memset(str, 1, sizeof(str));
```
damianloew, line 819: There is no NUL term!

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1019 @@
...
+
+	for (i = 1; i < PATH_MAX - 1; i++) {
```
damianloew, line 821: For is even not needed there. You could leave only looking for "2" in the big str. Placing some big str as s2 could be also a good idea, so maybe

1) only s1 as big str
2) only s2 as big str
3) two of them as big str

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1019 @@
...
+	TEST_ASSERT_EQUAL_STRING("bc", strstr("abc", "bc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "ac"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "xbc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "xab"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "aabbcc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "aaaaaa"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bbbbbb"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "cccccc"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "ccaabb"));
+	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bbaacc"));
```
damianloew, line 806: - TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bb")); already asserted by a previous "aa" check, so we can remove it and "cc"
- TEST_ASSERT_EQUAL_STRING("abc", strstr("abc", "ab")); checks nothing more than cases in `basic` - to remove, same with "bc"

TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "aabbcc"));
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "aaaaaa"));
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bbbbbb"));
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "cccccc"));
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "ccaabb"));
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bbaacc"));
I think those cases are not necessary too. If it does not catch 2 same bytes it rather won't catch more.

Also I'd add `TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bcx"));`, because there is some potential for an issue.

```suggestion
TEST(string_str, partially_included)
{
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "aa"));
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "ac"));
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "xbc"));
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "bcx"));
	TEST_ASSERT_EQUAL_STRING(NULL, strstr("abc", "xab"));
```

libc/string/string_tokpbrkstr.c
damianloew: Thanks for adding new cases, some remarks left.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1019 @@
...
+	/* Asserting reversed string */
+	for (i = 0; i < ASCII_LENGTH - 1; i++) {
+		TEST_ASSERT_EQUAL_PTR(testStr, strpbrk(testStr, &reversStr[i]));
+	}
+
+	free(testStr);
+}
+
+
+TEST(string_pbrk, part_of_str)
```
damianloew, line 927: I think we can split this case to `multiple` one, where you could check passing same bytes multiple time and to `additional_bytes` that could check passing some bytes not present in s1 to s2 (you did it here). Moreover, you can add a third case called `not_present` or sth like that, where any of bytes in s2 is not present in s1.

libc/string/string_tokpbrkstr.c
damianloew: I think you can apply the similar approach for the rest of functions.

libc/string/string_tokpbrkstr.c
damianloew: Where is such case? I can't see it for `strtok()` function.

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1019 @@
...
+		TEST_ASSERT_EQUAL_PTR(&extAsciiStr[i - ASCII_LENGTH], strtok_r(NULL, &separator[i - ASCII_LENGTH], &rest));
+		TEST_ASSERT_NOT_NULL(rest);
+	}
+
+	free(extAsciiStr);
+#endif
+}
+
+
+TEST(string_tok_r, state_restriction)
```
damianloew, line 635: So maybe:

```suggestion
TEST(string_tok_r, same_state)
```

libc/string/string_tokpbrkstr.c
maska989: Those cases are checked in `multi_call` in strtok and strtok_r

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1029 @@
...
+	TEST_ASSERT_EQUAL_STRING("abc", strtok("abc", ""));
+}
+
+
+TEST(string_tok, multi_call)
+{
+	char loremStr[20] = LOREM_IPSUM,
+		 sep[5] = { 0 },
+		 multiCallStr[ASCII_LENGTH] = { 0 },
+		 *asciiStr = testdata_createCharStr(ASCII_LENGTH);
```
damianloew, line 137: Please switch to the following approach in this test and prepare separate PR with the changes in the rest of files.

```suggestion
	char loremStr[20] = LOREM_IPSUM;
	char sep[5] = { 0 };
	char multiCallStr[ASCII_LENGTH] = { 0 };
	char *asciiStr = testdata_createCharStr(ASCII_LENGTH);
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1029 @@
...
+
+	/* Checking if set is empty */
+	TEST_ASSERT_EQUAL_PTR(NULL, strtok(NULL, &sep[0]));
+
+	/* Creating string with similar pattern */
+	for (i = 0; i < sizeof(multiCallStr) - 1; i++) {
+		multiCallStr[i] = i % 3 + 1;
+	}
+
+	/* Preparing separator with a bunch of similar elements*/
```
damianloew, line 198:
```suggestion
	/* Preparing separator with a bunch of same elements*/
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1029 @@
...
+	/* Checking if tokenization is in the correct places*/
+	for (i = 4; i < sizeof(multiCallStr) - 3; i = i + 3) {
+		TEST_ASSERT_EQUAL_PTR(&multiCallStr[i], strtok(NULL, sep));
+	}
+
+	memset(multiCallStr, 2, sizeof(multiCallStr));
+	memset(&multiCallStr[sizeof(multiCallStr) / 2], 1, 5);
+	multiCallStr[sizeof(multiCallStr) - 1] = 0;
+
+	/* Checking ability to tokenize element with multiple same signs treated as a chain of the same element from sep*/
```
damianloew, line 214:
```suggestion
	/* Checking ability to tokenize element with multiple same characters treated as a chain of the same element from sep*/
```

libc/string/string_tokpbrkstr.c
```diff
@@ -0,0 +1,1029 @@
...
+	char *asciiStr = testdata_createCharStr(ASCII_LENGTH);
+	char holder[2] = { 0 };
+
+	for (i = 1; i < ASCII_LENGTH - 1; i++) {
+		holder[0] = asciiStr[i];
+		TEST_ASSERT_EQUAL_STRING(holder, strpbrk(holder, asciiStr));
+	}
+
+	free(asciiStr);
+}
```
damianloew, line 963: In this case one of the bytes passed in `s2` was still present in `s1`, we have the similar cases before. I'd like to have sth like this:

```suggestion
TEST(string_pbrk, not_present)
{
	int i;
	char *asciiStr = testdata_createCharStr(ASCII_LENGTH + 1);
	char holder[2] = { 0 };
	char temp;

	for (i = 1; i < ASCII_LENGTH; i++) {
		holder[0] = i;
		asciiStr[i] = asciiStr[i - 1]
		TEST_ASSERT_EQUAL_PTR(NULL, strpbrk(holder, asciiStr));
		asciiStr[i] = i;
	}

	free(asciiStr);
}
```

As you can see I've increased a size by 1 to have the full ASCII range (1 to 126 was here before)

damianloew (changes requested):
Not so much left to change :)

libc/string/string_tokpbrkstr.c
damianloew: As we discussed please apply the following approach (note that for i == 1 we should take i + 1)

damianloew (approved):

damianloew (approved):

mateusz-bloch (approved):

PR #208, opened 2023-05-23
Adamdebek/exit test

adamdebek:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+	}
+	/* parent */
+	else {
+		ret = wait(&status);
+		TEST_ASSERT_EQUAL_INT(pid, ret);
+
+		/* Wait until handler end execution */
+		sleep(1);
+		errno = 0;
+		ret = open(PATH, O_RDWR);
```
niewim19, line 359: Why file reopening without close (in parent)? What is this used for?

exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+		TEST_ASSERT_NOT_EQUAL_INT(-1, ret);
+		TEST_ASSERT_NOT_EQUAL_INT(ENOENT, errno);
```
niewim19, line 361: `errno` should not be asserted if function call was succesful

exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+#define PATH       "/tmp/testFile"
+#define CREATEMODE 0666
+#define TEST_STR   "test123"
+
+static int fd;
+static int val = 0;
+
+
+/* SIGCHLD signal handler */
+static void handler(int signum)
```
niewim19, line 39: Please move this handler definition closer to SIGCHILD test as it is used only there

exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+{
+}
+
+
+TEST_TEAR_DOWN(exit)
+{
+}
+
+
+TEST(exit, stream_flush)
```
niewim19, line 384: Please force buffering rules via `setvbuf` to be sure that file writing is unbuffered.

exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+		TEST_ASSERT_EQUAL_INT(0, atexit(fun1));
+		TEST_ASSERT_EQUAL_INT(0, atexit(fun1));
+		TEST_ASSERT_EQUAL_INT(0, atexit(fun2));
+		exit(EXIT_SUCCESS);
+	}
+	/* parent */
+	else {
+		int ret = wait(&status);
+		TEST_ASSERT_EQUAL_INT(pid, ret);
+		int buf;
```
niewim19, line 444: Variables declaration should be done at the function beginning

exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+	}
+}
+
+
+TEST(exit, atexit_over32_calls)
+{
+	/* Test that exit() invokes funcs registered by atexit */
+	pid_t pid;
+	int status;
+	int atexit_max = sysconf(_SC_ATEXIT_MAX);
```
niewim19, line 463:
```suggestion
	int atexitMax = sysconf(_SC_ATEXIT_MAX);
```

exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+	fd = open(PATH, O_RDWR | O_CREAT | O_TRUNC);
```
niewim19, line 464:
```suggestion
	
	fd = open(PATH, O_RDWR | O_CREAT | O_TRUNC);
```

exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+		TEST_ASSERT_LESS_THAN_INT(0, pid);
+	}
+	/* child */
+	if (pid == 0) {
+		int i;
+		for (i = 0; i < atexit_max - 1; i++) {
+			TEST_ASSERT_EQUAL_INT(0, atexit(fun1));
+		}
+		TEST_ASSERT_EQUAL_INT(0, atexit(fun2));
+		TEST_ASSERT_EQUAL_INT(0, atexit(fun1));
```
niewim19, line 478: Why is `fun2` used? Can`t we try to register `atexitMax + 1` of `fun1` and assert only `atexitMax` invocations?

niewim19 (changes requested):
General, repeating remarks:
 - `if` clause on `fork` exit has no effect and is used in many testcases
 - variable declaration in the middle of the code
 - `TEST_ASSERT`s in `fork`-ed processes may produce strange unity behavior and give unclear output in test runner. If you need to assert something in child process then do it via `if` clause and pass failure information as exit status of that process.

exit/test_exit.c
adamdebek: Last register function is invoked first, so I created fun2 to make distinguish between function calls and ensure they are called in proper order.

exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+
+TEST(_exit, status_vals)
+{
+	/* Return all possible least significant byte values */
+	pid_t pid;
+	int status, i;
+	int val[3];
+	val[0] = 0x1 << 8;
+	val[1] = (0x1 << 16) + 1;
+	val[2] = (0x1 << 24) + 2;
```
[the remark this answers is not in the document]
adamdebek: This is done because only the least significant 8 bits shall be available to a waiting parent process. 3 most significant bytes should be cut off so returning val[0] should resolve to returning 0, val[1] to 1 and val[2] to 2.

exit/test_exit.c
```diff
@@ -0,0 +1,518 @@
...
+	gpid = fork();
+	if (gpid < 0) {
+		TEST_ASSERT_LESS_THAN_INT(0, gpid);
+	}
+	if (gpid == 0) {
+		pid = fork();
+		if (pid < 0) {
+			TEST_ASSERT_LESS_THAN_INT(0, pid);
+		}
+		/* grandchild */
```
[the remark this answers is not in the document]
adamdebek: I think grandchild directly indicates that it is child of parent and grandchild of grandparent.

exit/test_exit.c
adamdebek: I think we shouldn't place function definitions among testcase definitions.

exit/test_exit.c
adamdebek: I think it is unnecessary to use setvbuf, because fopen with file grants fully buffered stream.

exit/test_exit.c
```diff
@@ -0,0 +1,513 @@
...
+	int fd, val = 1234;
+
+	fd = open(PATH2, O_WRONLY | O_APPEND);
+	TEST_ASSERT_NOT_EQUAL_INT(-1, fd);
+	TEST_ASSERT_EQUAL_INT(sizeof(int), write(fd, &val, sizeof(int)));
+	close(fd);
+}
+
+
+TEST_GROUP(_exit);
```
niewim19, line 71: Function name is used as a label in macro. I would recommend to change it. It might backfire in the most unpredictable way.

exit/test_exit.c
```diff
@@ -0,0 +1,513 @@
...
+#include <unity_fixture.h>
+
+#define PATH1    "/tmp/testFile1"
+#define PATH2    "/tmp/testFile2"
+#define PATH3    "/tmp/testFile3"
+#define PATH4    "/tmp/testFile4"
+#define TEST_STR "test123"
+
+/* Set to random value */
+static int test_creatRes = 0x12345678;
```
niewim19, line 38: Just a suggestion, imho more descriptive when found in the code. 
```suggestion
static int sigHandleRes = 0x12345678;
```
Why should it be some random number? `-1` wouldn`t be enough?

exit/test_exit.c
```diff
@@ -0,0 +1,513 @@
...
+			TEST_ASSERT_EQUAL_INT(pid, ret);
+			TEST_ASSERT_EQUAL_INT(i, WEXITSTATUS(status));
+		}
+	}
+	/* Some exceeding values */
+	for (i = 0; i < 3; i++) {
+		pid = fork();
+		if (pid < 0) {
+			TEST_ASSERT_LESS_THAN_INT(0, pid);
+		}
```
niewim19, line 119: Such `if`s were pointed in previous review. All are corrected  so i think this flew unchanged under the radar :)

exit/test_exit.c
```diff
@@ -0,0 +1,513 @@
...
+		/* child */
+		if (pid == 0) {
+			_exit(val[i]);
+		}
+		/* parent */
+		else {
+			int status, ret;
```
niewim19, line 126: All tests below (this included) declare some variables deep inside the test body. Please follow the coding convention of variable declaration in the beginning of the body.

exit/test_exit.c
niewim19: Yeah, you`re right. My remark is invalid.

exit/test_exit.c
```diff
@@ -0,0 +1,513 @@
...
+TEST(_exit, orphaned_child)
+{
+	/* Test if parent _exit affect child process */
+	pid_t pid;
+
+	pid = fork();
+	TEST_ASSERT_GREATER_OR_EQUAL(0, pid);
+	if (pid == 0) {
+		pid = fork();
+		TEST_ASSERT_GREATER_OR_EQUAL(0, pid);
```
niewim19, line 227: Assert within the child. I know that this was proposed as a desirable way of checking `fork` status, but not when you double-`fork` in the test. Now fail of the second `fork` throws off the test framework.

An instance is also present in other tests that double-`fork`, please take care of that. Maybe consult @damianloew on that matter.

exit/test_exit.c
```diff
@@ -0,0 +1,513 @@
...
+			close(pipefd[1]);
+			/* Sleep until child gets his parent ID */
+			sleep(1);
+			_exit(EXIT_SUCCESS);
+		}
+	}
+	/* grandparent */
+	else {
+		pid_t old_ppid, new_ppid;
+		int status;
```
niewim19, line 300: variable declaration deep in test body, not in the beginning. Also names do not follow coding convention.

niewim19 (changes requested):

exit/test_exit.c
adamdebek: sigHandleRes indicates that the value is returned by handler when its not. If it would be `-1` then when creat() failed then the value is not changed.

exit/test_exit.c
adamdebek: I think this approach is more readable and was used in other files in the past: https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/4334a5c63a0d66c7b884f34259c76b60ccdd50d5/socket/test_socket.c?plain=1#L623-L670

exit/test_exit.c
adamdebek: I think this approach is more readable and was used in other files in the past: https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/4334a5c63a0d66c7b884f34259c76b60ccdd50d5/socket/test_socket.c?plain=1#L623-L670

exit/test_exit.c
niewim19: Yes, this was used in the past. This code is pushed now and it should conform to current coding convention.

exit/test_exit.c
niewim19: Yes, this was used in the past. This code is pushed now and it should conform to current coding convention.

exit/test_exit.c
damianloew: Is this order asserted somewhere?

exit/test_exit.c
damianloew: We will discuss it during the meeting weekly meeting.

exit/test_exit.c
damianloew: @adamdebek still the variable names are not correct, please use camelCase.

exit/test_exit.c
```diff
@@ -0,0 +1,528 @@
...
+	if (pid == 0) {
+		pid = fork();
+		/* grandchild */
+		if (pid == 0) {
+			pid_t old_ppid, new_ppid;
+
+			close(pipefd[0]);
+			old_ppid = getppid();
+			/* Sleep until parents exits */
+			sleep(2);
```
damianloew, line 293: I think we can use much lower sleep values, you can consider using usleep - it may make this test much faster.

exit/test_exit.c
```diff
@@ -0,0 +1,528 @@
...
+
+
+TEST(unistd_exit, SIGCHLD_sent)
+{
+	/* Test that SIGCHILD signal is sent after child exits */
+	pid_t pid;
+	int status, ret;
+
+	struct sigaction sa;
+	sa.sa_handler = test_handler;
```
damianloew, line 338: Did you want to test this paragraph? If so, handler should be SIG_IGN:
```suggestion
	sa.sa_handler = SIG_IGN;
```
Screenshot from 2023-08-10 18-40-40

exit/test_exit.c
```diff
@@ -0,0 +1,528 @@
...
+
+		close(pipefd[1]);
+		wait(&status);
+		TEST_ASSERT_EQUAL_INT(EXIT_SUCCESS, WEXITSTATUS(status));
+
+		read(pipefd[0], &pid, sizeof(pid_t));
+		TEST_ASSERT_NOT_EQUAL_INT(-1, pid);
+		close(pipefd[0]);
+
+		/* Check if file exists, if so child sleep was interrupted */
```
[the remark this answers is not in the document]
adamdebek: Present comment is correct.

exit/test_exit.c
adamdebek: image

exit/test_exit.c
adamdebek: I got rid of sleeps

exit/test_exit.c
```diff
@@ -0,0 +1,665 @@
...
+		  -> https://github.com/phoenix-rtos/phoenix-rtos-project/issues/806
+
+		- threads terminated by a call to _Exit() or _exit() shall not invoke their cancellation cleanup handlers
+		  -> https://github.com/phoenix-rtos/phoenix-rtos-project/issues/827
+
+		- if the exit of the process causes a process group to become orphaned, and if any member of the newly-orphaned
+		  process group is stopped, then a SIGHUP signal followed by a SIGCONT signal shall be sent to each process in
+		  the newly-orphaned process group -> https://github.com/phoenix-rtos/phoenix-rtos-project/issues/809
+ */
+
```
damianloew, line 52: - `Open streams shall not be flushed.` - Is it checked anywhere?

damianloew, line 52: - directory streams, conversion descriptors, and message catalog descriptors open in the calling process shall be closed - I think director streams (*DIR) may be asserted, I am not sure about conversion descriptors, and message catalog descriptors - to be checked. If sth is not possible please leave a comment.

damianloew, line 52: - `If the parent process of the calling process has set its SA_NOCLDWAIT flag or has set the action for the SIGCHLD signal to SIG_IGN:` - I don't see setting the `SIG_IGN` action for parent - please correct me if I am wrong, I also don't see `SA_NOCLDWAIT`

damianloew, line 52: And the rest of points - I haven't verified the rest yet.
Screenshot from 2023-09-18 18-32-06

adamdebek: This quote refers to `X/OPEN` functionality.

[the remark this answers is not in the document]
adamdebek: siginfo_t can be accessed only when SA_SIGINFO flag is set, but it is not present in PHOENIX-RTOS right now.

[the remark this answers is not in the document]
adamdebek: Yes waitid() not implemented.

adamdebek: AFAIK it is not possible to check closure of directory stream.

exit/test_exit.c
```diff
@@ -0,0 +1,811 @@
...
+		test_common.test_exitPtr(EXIT_SUCCESS);
+	}
+	/* parent */
+	else {
+		int status, ret;
+
+		ret = waitpid(pid, &status, 0);
+		TEST_ASSERT_EQUAL_INT(pid, ret);
+		TEST_ASSERT_EQUAL_INT(EXIT_SUCCESS, WEXITSTATUS(status));
+	}
```
damianloew, line 219: We could create a case, where a few threads would be created in parent process. They could all wait using waitpid/wait (it could be good to use both) and write status to a file if a status has been read. Finally only one status should be written.
Screenshot from 2023-09-29 11-08-00

exit/test_exit.c
```diff
@@ -0,0 +1,811 @@
...
+		  If the process is a controlling process, the controlling terminal associated with the session shall be
+		  disassociated from the session, allowing it to be acquired by a new controlling process
+		  -> ???
+
+		- The full value of status shall be available from waitid() and in the siginfo_t passed to a signal handler
+		  for SIGCHLD -> https://github.com/phoenix-rtos/phoenix-rtos-project/issues/844,
+		  https://github.com/phoenix-rtos/phoenix-rtos-project/issues/845
+ */
+
+
```
damianloew, line 60: What about memory mapping?
Screenshot from 2023-09-29 11-34-27
Please research it and leave a comment/issue about it.

damianloew (changes requested):
Some ideas for more cases + suggested information about not covered requirements

exit/test_exit.c
```diff
@@ -0,0 +1,811 @@
...
+	if (pid == 0) {
+		test_common.test_exitPtr(EXIT_SUCCESS);
+	}
+	/* parent */
+	else {
+		int status, ret;
+
+		ret = waitpid(pid, &status, 0);
+		TEST_ASSERT_EQUAL_INT(pid, ret);
+		TEST_ASSERT_EQUAL_INT(EXIT_SUCCESS, WEXITSTATUS(status));
```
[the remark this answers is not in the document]
adamdebek: I added second waitpid() in exit_status_waitpid testcase to check if process was reaped.

exit/test_exit.c
```diff
@@ -0,0 +1,794 @@
...
+		  process group of the controlling terminal belonging to the calling process.
+		  If the process is a controlling process, the controlling terminal associated with the session shall be
+		  disassociated from the session, allowing it to be acquired by a new controlling process
+		  -> ???
+
+		- The full value of status shall be available from waitid() and in the siginfo_t passed to a signal handler
+		  for SIGCHLD -> https://github.com/phoenix-rtos/phoenix-rtos-project/issues/844,
+		  https://github.com/phoenix-rtos/phoenix-rtos-project/issues/845
+ */
+
```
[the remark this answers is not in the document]
adamdebek: Process memory locking is optional in `POSIX`, I will skip this comment.

exit/test_exit.c
```diff
@@ -0,0 +1,794 @@
...
+		- if the exit of the process causes a process group to become orphaned, and if any member of the newly-orphaned
+		  process group is stopped, then a SIGHUP signal followed by a SIGCONT signal shall be sent to each process in
+		  the newly-orphaned process group -> https://github.com/phoenix-rtos/phoenix-rtos-project/issues/809
+
+		- If the process is a controlling process, the SIGHUP signal shall be sent to each process in the foreground
+		  process group of the controlling terminal belonging to the calling process.
+		  If the process is a controlling process, the controlling terminal associated with the session shall be
+		  disassociated from the session, allowing it to be acquired by a new controlling process
+		  -> ???
```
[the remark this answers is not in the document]
adamdebek: I added: Phoenix-RTOS doesn't implement control of terminal by session leader with that being said controlling proccess is not supported

exit/test-exit.c
```diff
@@ -0,0 +1,1071 @@
...
+		remove(TEST_EXIT_PATH);
+	}
+}
+
+
+TEST(stdlib_exit, atexit_over32_calls)
+{
+	/* Test that exit() invokes over 32 funcs registered by atexit, 32 is number of function slots in 1 node */
+#ifndef __phoenix__
+	TEST_IGNORE();
```
damianloew, line 949:
```suggestion
	/* Phoenix-specific test case based on the implementation */
	TEST_IGNORE();
```

exit/test-exit.c
```diff
@@ -0,0 +1,1071 @@
...
+		ret = wait(&status);
+		TEST_ASSERT_EQUAL_INT(pid, ret);
+		TEST_ASSERT_EQUAL_INT(EXIT_SUCCESS, WEXITSTATUS(status));
+
+		TEST_ASSERT_EQUAL_INT(0, test_common.test_handlerFlag);
+	}
+}
+
+
+TEST_GROUP(stdlib_exit);
```
damianloew, line 854: From `exit` doc:

If a function registered by a call to atexit() fails to return, the remaining registered functions shall not be called and the rest of the exit() processing shall not be completed

IMHO it may also be tested.

damianloew (changes requested):

exit/test-exit.c
```diff
@@ -0,0 +1,1071 @@
...
+		sigaction(SIGUSR1, &sa, NULL);
+
+		pause();
+		test_common.test_exitPtr(EXIT_SUCCESS);
+	}
+	/* parent */
+	else {
+		volatile struct test_ThreadArgs argsThr1, argsThr2;
+		time_t start, curr;
+		double timeout = 3.0;
```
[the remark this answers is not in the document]
adamdebek: This timeout only occurs when test fail so hopefully very rare so I would stick to 3.0s.

exit/test-exit.c
```diff
@@ -0,0 +1,1071 @@
...
+		if (argsThr1.retWaitThr == pid) {
+			/* Wait() in second thread should -1 */
+			while (difftime(curr, start) <= timeout) {
+				curr = time(NULL);
+				if (argsThr2.retWaitThr == -1 && argsThr2.errnoThr == ECHILD) {
+					TEST_PASS();
+				}
+				else if (argsThr2.retWaitThr != 0 && argsThr2.errnoThr == 0) {
+					TEST_FAIL_MESSAGE("Error: More than 1 thread unblocked");
+				}
```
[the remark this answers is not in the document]
adamdebek: There is only 1 more `errno` value, which is EINTR but that is impossible to happen.

exit/test-exit.c
adamdebek: IMHO it is pointless.

exit/test-exit.c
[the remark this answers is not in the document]
damianloew: Even if it's hardly impossible I'd like to handle it (the potentially may be some issue with a system).

exit/test-exit.c
```diff
@@ -0,0 +1,1071 @@
...
+
+
+TEST_TEAR_DOWN(stdlib_exit)
+{
+}
+
+
+TEST(stdlib_exit, stream_flush)
+{
+	/* Test that exit() force unwritten buffered data to be flushed */
```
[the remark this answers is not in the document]
damianloew: Not applied.

exit/test-exit.c
```diff
@@ -0,0 +1,1071 @@
...
+	remove(TEST_EXIT_PATH);
+}
+
+
+TEST_TEAR_DOWN(unistd_exit)
+{
+}
+
+
+TEST(unistd_exit, status_vals)
```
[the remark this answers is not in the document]
damianloew: It potentially may change.

exit/test-exit.c
damianloew: Ok, it would be hard.

exit/test-exit.c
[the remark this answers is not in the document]
adamdebek: It will not change, these function are designed to be called in this manner.

libc/exit/exit.c
```diff
@@ -0,0 +1,1236 @@
...
+
+
+int main(int argc, char *argv[])
+{
+	const char *tiniPath = "/tini";
+
+	/* Add test with tini (init system) if test is run on host in container */
+	if (access(tiniPath, F_OK) == 0 && getenv("use_tini") == NULL) {
+		setenv("use_tini", "yes", 1);
+		execl("/tini", "/tini", "-s", "/github/workspace/_fs/host-generic-pc/root/bin/test-libc-exit", NULL);
```
damianloew, line 1227: It would be the best to use `argv[0]` instead of the absolute path, which may change.

libc/exit/exit.c
```diff
@@ -0,0 +1,1236 @@
...
+	RUN_TEST_GROUP(stdlib_exit);
+}
+
+
+int main(int argc, char *argv[])
+{
+	const char *tiniPath = "/tini";
+
+	/* Add test with tini (init system) if test is run on host in container */
+	if (access(tiniPath, F_OK) == 0 && getenv("use_tini") == NULL) {
```
damianloew, line 1225: IMHO more readable, `use_tini` not set sound's like we don't want to use it.

```suggestion
	if (access(tiniPath, F_OK) == 0 && getenv("tini_used") == NULL) {
```

libc/exit/exit.c
```diff
@@ -0,0 +1,1236 @@
...
+{
+	RUN_TEST_GROUP(unistd_exit);
+	RUN_TEST_GROUP(unistd_Exit);
+	RUN_TEST_GROUP(stdlib_exit);
+}
+
+
+int main(int argc, char *argv[])
+{
+	const char *tiniPath = "/tini";
```
damianloew, line 1222: I'd like to move this code with tini to the PR, where you will enable the ignored cases (when docker devel is ready). For now let's merge the version without it.

damianloew (changes requested):

libc/exit/exit.c
```diff
@@ -0,0 +1,1224 @@
...
+
+	/* Add test with tini (init system) if test is run on host in container */
+	if (access(tiniPath, F_OK) == 0 && getenv("my_env") == NULL) {
+		setenv("use_tini", "yes", 1);
+		execl("/tini", "/tini", "-s", "/project/_fs/host-generic-pc/root/bin/test-libc-exit", NULL);
```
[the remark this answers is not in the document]
adamdebek: it is invoked as `/tini`

libc/exit/exit.c
adamdebek: I think `use_tini` sound better.

damianloew (approved):
Thanks for all the corrections, LGTM

niewim19 (approved):

PR #214, opened 2023-06-14
move printf test

mateusz-bloch:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/printf/test.yaml
+++ b/libc/printf/test.yaml
@@ -6,3 +6,10 @@ test:
       targets:
         exclude: [armv7m4-stm32l4x6-nucleo]
         include: [host-generic-pc]
+
+    # Executing test-printf on 'armv7m4-stm32l4x6-nucleo' takes ~9 min, that's why we want to run it only during nightly campaigns 
+    - name: unit
+      execute: test-libc-printf
+      nightly: true
+      targets:
+        value: [armv7m4-stm32l4x6-nucleo]
--- a/libc/printf/test.yaml
+++ b/libc/printf/test.yaml
@@ -1,0 +1,8 @@
+test:
+  type: unity
+  tests:
+    - name: unit
+      execute: test-libc-printf
+      targets:
+        exclude: [armv7m4-stm32l4x6-nucleo]
+        include: [host-generic-pc]
--- a/Makefile.host
+++ b/Makefile.host
@@ -1,3 +1,4 @@
 DEFAULT_COMPONENTS = $(filter test_meterfs_%,$(ALL_COMPONENTS))
 DEFAULT_COMPONENTS += $(SAMPLE_TESTS)
 DEFAULT_COMPONENTS += test-libc
+DEFAULT_COMPONENTS += test-libc-printf
--- a/libc/main.c
+++ b/libc/main.c
@@ -64,14 +64,6 @@ void runner(void)
 	RUN_TEST_GROUP(string_strlcat);
 	RUN_TEST_GROUP(string_errsign);
 	RUN_TEST_GROUP(signal_psignal);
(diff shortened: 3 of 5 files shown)


printf/stdio_printf.c
```diff
@@ -0,0 +1,2645 @@
...
+					close(fd); \
+					close(stdout_fd); \
+					break; \
+				} \
+				case SNTEST: \
+					snprintf(buff, strlen(expect) + 1, format, value); \
+					break; \
+				case STEST: \
+					sprintf(buff, format, value); \
+					break; \
```
damianloew, line 88: We use braces even for one line blocks.

```suggestion
				case SNTEST: \
					snprintf(buff, strlen(expect) + 1, format, value); \
					break; \
				} \
				case STEST: { \
					sprintf(buff, format, value); \
					break; \
				} \
```

printf/stdio_printf.c
```diff
@@ -0,0 +1,2645 @@
...
+#define NUM_OUTPUT_TYPES 5
+
+#define PATH "stdio_printf_test"
+
+/* comma delimiter macro ',' for multiple arguments parsing to *printf() function via macros */
+#define COMMA ,
+
+/* macro assertions*/
+
+#define printf_test(format, value, expect) \
```
damianloew, line 60: Let's have a common convention in our test and prefix all global functions/variables by `test` and use camelCase as it's recommended in our coding convention.

```suggestion
static FILE *test_outFile;
static char test_buff[700];

enum test_outType { DTEST,
	FTEST,
	TEST,
	SNTEST,
	STEST };

#define NUM_OUTPUT_TYPES 5

#define PATH "stdio_printf_test"

/* comma delimiter macro ',' for multiple arguments parsing to *printf() function via macros */
#define COMMA ,

/* macro assertions*/

#define test_assertPrintfs(format, value, expect) \
```

printf/stdio_printf.c
```diff
@@ -0,0 +1,2645 @@
...
+			} \
+			fseek(output_file, 0, SEEK_SET); \
+			fflush(output_file); \
+			TEST_ASSERT_EQUAL_STRING(expect, fgets(buff, strlen(expect) + 1, output_file)); \
+			TEST_ASSERT_EQUAL_STRING(expect, buff); \
+		} \
+	}
+
+/* wrappers for vprintf tests*/
+static void vprintf_test(char *expect, const char *format, ...)
```
damianloew, line 98:
```suggestion
static void test_assertVprintfs(char *expect, const char *format, ...)
```

damianloew (changes requested):
Thanks for decreasing the size of binary. We can order names of variables/functions by an occasion. Above that please prefix commit messages by `libc_`, give some commit description about why decreasing size/moving to the separate binary is needed. Please place this directory inside the libc one as you suggested.

printf/stdio_printf.c
nalajcie: please note that adding subscopes (`{ }`) in switch `case's` is not needed (unless You're declaring variables there) and IMHO decreases readability.

damianloew: Yeah right, it isn't needed there, we will remove them.

libc/printf/test.yaml
```diff
@@ -0,0 +1,14 @@
+test:
+  type: unity
+  tests:
+    - name: unit
```
damianloew, line 4: We always want to leave a comment, why do we use an unusual approach, like this one:
```suggestion
    # Executing test-printf on 'armv7m4-stm32l4x6-nucleo' takes ~9 min, that's why we want to run it only during nightly campaigns 
    - name: unit
```

damianloew (approved):
Thanks for the changes, it looks much simpler

PR #216, opened 2023-06-21
libc: resolvepath test improvments

mateusz-bloch:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/misc/resolve_path.c
+++ b/libc/misc/resolve_path.c
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
 }
 
 
+/* check if path temporarily longer than PATH_MAX while resolving won't crash */
+TEST(resolve_path, symlink_long_resolution)
+{
+	char symNameLong[NAME_MAX / 2];
+	char resolved[PATH_MAX + 1];
+	char basePath[PATH_MAX - 5] = "symShort/";
+	int basePathLen = strlen(basePath);
+	const char symNameShort[] = "symShort";
+	const char pathSegment[] = "dev/../";
+	const int pathSegmentLen = sizeof(pathSegment) - 1;
+
+	while (basePathLen + pathSegmentLen < sizeof(basePath) - 5) {
+		strcat(basePath, pathSegment);
+		basePathLen += pathSegmentLen;
+	}
+	strcat(basePath, "dev");
+
+	memset(symNameLong, 'a', sizeof(symNameLong));
+	symNameLong[sizeof(symNameLong) - 1] = 0;
+
+	unlink(symNameShort);
+	unlink(symNameLong);
+
+	TEST_ASSERT_EQUAL_INT(0, symlink("/", symNameLong));
+	TEST_ASSERT_EQUAL_INT(0, symlink(symNameLong, symNameShort));
+
+	errno = 0;
+
+	/* As it's described in the `realpath()` doc the function MAY fail in such case */
+	if (realpath(basePath, resolved) != NULL) {
+		TEST_ASSERT_EQUAL_STRING("/dev", resolved);
+	}
+	else {
+		TEST_ASSERT_EQUAL_INT(ENAMETOOLONG, errno);
+	}
+
+	/* cleanup */
+	unlink(symNameShort);
+	unlink(symNameLong);
+}
+
+
 TEST_GROUP_RUNNER(resolve_path)
 {
 	RUN_TEST_CASE(resolve_path, canonicalize_abs_simple);
@@ -497,4 +539,5 @@ TEST_GROUP_RUNNER(resolve_path)
 	RUN_TEST_CASE(resolve_path, symlink_dir);
 	RUN_TEST_CASE(resolve_path, symlink_loop);
 	RUN_TEST_CASE(resolve_path, symlink_rename);
+	RUN_TEST_CASE(resolve_path, symlink_long_resolution);
 }


libc/resolve_path.c
```diff
@@ -475,6 +475,47 @@ TEST(resolve_path, symlink_rename)
...
+TEST(resolve_path, symlink_enametoolong)
+{
+	char *dev = "/dev";
+	char *pathSegment = "dev/../";
+	char symNameShort[] = "symShort";
+	char symNameLong[NAME_MAX / 2];
+	char basePath[PATH_MAX - 5] = "symShort/";
+	char resolved[PATH_MAX + 1];
+	int basePathLen = strlen(basePath);
+	int pathSegmentLen = strlen(pathSegment);
```
damianloew, line 487:
```suggestion
	const char *dev = "/dev";
	const char *pathSegment = "dev/../";
	const char symNameShort[] = "symShort";
	char symNameLong[NAME_MAX / 2];
	char basePath[PATH_MAX - 5] = "symShort/";
	char resolved[PATH_MAX + 1];
	int basePathLen = strlen(basePath);
	const int pathSegmentLen = strlen(pathSegment);
```

libc/resolve_path.c
```diff
@@ -475,6 +475,47 @@ TEST(resolve_path, symlink_rename)
 }
 
 
+TEST(resolve_path, symlink_enametoolong)
```
damianloew, line 478: That's not only about ENAMETOOLONG, rather a special case, which may raise ENAMETOOLONG errno. Above that we have test case named errno in this test, which also checks ENAMETOOLONG (it could be misleading that it's not in enametoolong).

```suggestion
TEST(resolve_path, symlink_long_resolution)
```

damianloew (changes requested):

libc/resolve_path.c
```diff
@@ -475,6 +475,47 @@ TEST(resolve_path, symlink_rename)
...
+
+	unlink(symNameShort);
+	unlink(symNameLong);
+
+	TEST_ASSERT_EQUAL_INT(0, symlink("/", symNameLong));
+	TEST_ASSERT_EQUAL_INT(0, symlink(symNameLong, symNameShort));
+
+	errno = 0;
+
+	if (realpath(basePath, resolved) != NULL) {
```
damianloew, line 506:
```suggestion
        /* As it's described in the `realpath()` doc the function may fail in such case */
	if (realpath(basePath, resolved) != NULL) {
```

damianloew (changes requested):

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
 }
 
 
+TEST(resolve_path, symlink_long_resolution)
+{
+	const char *dev = "/dev";
+	const char *pathSegment = "dev/../";
+	const char symNameShort[] = "symShort";
```
niewim19, line 482: Not doing a review, just curious - why is one a local array and two other pointers to strings?

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
 }
 
 
+TEST(resolve_path, symlink_long_resolution)
+{
+	const char *dev = "/dev";
```
nalajcie, line 480: used once in line 508, write it directly there

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
...
+	int basePathLen = strlen(basePath);
+	const int pathSegmentLen = strlen(pathSegment);
+
+	while (basePathLen + pathSegmentLen < PATH_MAX - 10) {
+		strcat(basePath, pathSegment);
+		basePathLen += pathSegmentLen;
+	}
+	strcat(basePath, "dev");
+
+	memset(symNameLong, 'a', NAME_MAX / 2);
```
nalajcie, line 495: use `sizeof(symNameLong)`  and similar instead of direct values (that way eventual future len change would need to be done only in one place)

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
 }
 
 
+TEST(resolve_path, symlink_long_resolution)
```
nalajcie, line 478: please provide one-line description as with other tests

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
...
+{
+	const char *dev = "/dev";
+	const char *pathSegment = "dev/../";
+	const char symNameShort[] = "symShort";
+	char symNameLong[NAME_MAX / 2];
+	char basePath[PATH_MAX - 5] = "symShort/";
+	char resolved[PATH_MAX + 1];
+	int basePathLen = strlen(basePath);
+	const int pathSegmentLen = strlen(pathSegment);
```
nalajcie, line 487: place directly after `pathSegment` declaration of better readability

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
...
+
+	while (basePathLen + pathSegmentLen < PATH_MAX - 10) {
```
nalajcie, line 489: use `sizeof(basePath)`

nalajcie (commented):
some nitpicks :)

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
...
+/* check the behavior of the realpath function when it encounters pathname resolution of a symbolic link that result with a huge length */
+TEST(resolve_path, symlink_long_resolution)
+{
+	const char symNameShort[] = "symShort";
+	char symNameLong[NAME_MAX / 2];
+	char resolved[PATH_MAX + 1];
+	char basePath[PATH_MAX - 5] = "symShort/";
+	int basePathLen = strlen(basePath);
+	const char *pathSegment = "dev/../";
+	const int pathSegmentLen = sizeof(pathSegment);
```
damianloew, line 487: As you can see below using len name when assigning sizeof value may be misleading.

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
...
+
+	while (basePathLen + pathSegmentLen < sizeof(basePath) - 5) {
+		strcat(basePath, pathSegment);
+		basePathLen += pathSegmentLen;
```
damianloew, line 491: **basePathLen is much bigger than an actual size of basePath. Note that you add sizeof each time, so 8 when in fact only 7 bytes are added.**

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
...
 
+/* check the behavior of the realpath function when it encounters pathname resolution of a symbolic link that result with a huge length */
+TEST(resolve_path, symlink_long_resolution)
+{
+	const char symNameShort[] = "symShort";
+	char symNameLong[NAME_MAX / 2];
+	char resolved[PATH_MAX + 1];
+	char basePath[PATH_MAX - 5] = "symShort/";
+	int basePathLen = strlen(basePath);
+	const char *pathSegment = "dev/../";
```
damianloew, line 486: Not so important, but imho when we don't need to allocate more space, the pointer is ok. To increase readability we could place it next to the second declared string.

```suggestion
	char symNameLong[NAME_MAX / 2];
	char resolved[PATH_MAX + 1];
	char basePath[PATH_MAX - 5] = "symShort/";
	int basePathLen = strlen(basePath);
        const char *symNameShort = "symShort";
	const char *pathSegment = "dev/../";
```

libc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
 }
 
 
+/* check the behavior of the realpath function when it encounters pathname resolution of a symbolic link that result with a huge length */
```
damianloew, line 478:
```suggestion
/* check the behavior of the realpath function when it encounters pathname resolution of a symbolic link that results with a length longer than PATH_MAX */
```

damianloew (changes requested):

libc/resolve_path.c
damianloew: Not so crucial, so we can leave it, but I meant to place it a few lines lower.

damianloew (approved):
All requested changes seems to be applied, so I'd rather merge it at once :)

libc/resolve_path.c
nalajcie: yeah, for `sizeof` use name `pathSegmentSize`, in this case you could write `pathSegmentLen = sizeof(pathSegment) -1`

libc/misc/resolve_path.c
```diff
@@ -475,6 +475,48 @@ TEST(resolve_path, symlink_rename)
 }
 
 
+/* check the behavior of the realpath function when it encounters pathname resolution of a symbolic link that results with a length longer than PATH_MAX */
```
nalajcie, line 478: be specific and short with comments, eg.:

_check if path temporarily longer than PATH_MAX while resolving won't crash_

that way the if/else in the test makes sense (the implementation might support resolving the path or might return an error - crash is being considered a failure :))

damianloew (approved):

libc/misc/resolve_path.c
damianloew: I've corrected it as Mateusz is not available (It could be good to merge it today to provide running this test on PR with new changes)

PR #228, opened 2023-07-20
trunner: phoenixd usb lookup

maska989:
## Description

trunner: check the solution using vid, pid instead of device name for plo

## Motivation and Context

JIRA: CI-304

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/tools/phoenix.py
+++ b/trunner/tools/phoenix.py
@@ -3,23 +3,36 @@ import os
 import signal
 import threading
 import time
-from contextlib import contextmanager
-
 import pexpect
 
 from trunner.harness import ProcessError
+from serial.tools import list_ports
+from contextlib import contextmanager
 from .common import add_output_to_exception
 
 
-def wait_for_dev(port, timeout=0):
-    asleep = 0
+def wait_for_vid_pid(vid: int, pid: int, timeout=0):
+    """wait for connected usb serial device with required vendor & product id"""
 
-    # naive wait for dev
-    while not os.path.exists(port):
+    asleep = 0
+    found_ports = []
+
+    while not found_ports:
         time.sleep(0.01)
         asleep += 0.01
+
+        found_ports = [port for port in list_ports.comports() if port.pid == pid and port.vid == vid]
+
+        if len(found_ports) > 1:
+            raise Exception(
+                "More than one plo port was found! Maybe more than one device is connected? Hint used to find port:"
+                f"{vid:04x}:{pid:04x}"
+            )
+
         if timeout and asleep >= timeout:
-            raise TimeoutError
+            raise TimeoutError(f"Couldn't find plo USB device with vid/pid: '{vid:04x}:{pid:04x}'")
+
+    return found_ports[0].device
 
 
 class PsuError(ProcessError):
@@ -70,8 +83,9 @@ class PhoenixdError(ProcessError):
 class Phoenixd:
     """Handler for phoenixd process"""
 
-    def __init__(self, port="/dev/serial/by-id/usb-Phoenix_Systems_plo_CDC_ACM-if00", cwd=".", directory="."):
-        self.port = port
+    def __init__(self, vid=0x16F9, pid=0x0003, cwd=".", directory="."):
+        self.vid = vid
+        self.pid = pid
         self.proc = None
         self.reader_thread = None
         self.cwd = cwd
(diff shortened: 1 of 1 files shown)


trunner/tools/phoenix.py
```diff
@@ -22,6 +22,23 @@ def wait_for_dev(port, timeout=0):
             raise TimeoutError
 
 
+def wait_for_vid_pid(port_hint: str, timeout=0):
```
mateusz-bloch, line 25: I would renaming it to be more descriptive of its function.
Maybe something like: `find_device_by_vid_pid`

trunner/tools/phoenix.py
```diff
@@ -22,6 +22,23 @@ def wait_for_dev(port, timeout=0):
...
+    asleep = 0
+    port = None
+
+    while port is None:
```
mateusz-bloch, line 29: I thinking if it's worth changing it to `while not port,` for better readability, but then the loop would check if it's not just None, but if it's not every state that python considers falsy

trunner/tools/phoenix.py
```diff
@@ -22,6 +22,23 @@ def wait_for_dev(port, timeout=0):
...
+        time.sleep(0.01)
+        asleep += 0.01
+
+        for p in list_ports.grep(port_hint):
```
nalajcie, line 33: grep uses regexp as a param, either make a regexp or search for vid/pid by looping (they are properties of [ListPortInfo](https://pyserial.readthedocs.io/en/latest/tools.html#serial.tools.list_ports.ListPortInfo)).

Also - don't iterate over results (if there is more than one it should be considered an error?)

trunner/tools/phoenix.py
```diff
@@ -22,6 +22,23 @@ def wait_for_dev(port, timeout=0):
...
+            port = p
+
+        if timeout and asleep >= timeout:
+            raise PhoenixdError(f"Couldn't find USB device with vid/pid: '{port_hint}'. Timeout.") from TimeoutError
```
nalajcie, line 37: this is a generic function (which might be reused not only in `Phoenixd` class so it should throw generic exception `TimeoutError`, not phoenixd-specific one

trunner/tools/phoenix.py
maska989: I'm not 100% sure about the solution I used, so I'll wait for your feedback

trunner/tools/phoenix.py
```diff
@@ -22,6 +22,30 @@ def wait_for_dev(port, timeout=0):
...
+def wait_for_vid_pid(port_hint: str, timeout=0):
+    """Searcher for connected usb device via vendor & product id"""
+
+    asleep = 0
+    port = None
+
+    while port is None:
+        time.sleep(0.01)
+        asleep += 0.01
+
```
nalajcie, line 34: i would go for specifying pid/vid as a number and than:

```python
found_ports = []

while not found_ports:
  time.sleep(0.01)
  asleep += 0.01

  found_ports = [ port for port in list_ports.comports() if port.pid == pid and port.vid == vid ]
  if len(found_ports) > 1:
     raise Exception(...)

  # timeout handling here

return found_ports[0].device
```
(the code was not tested, I'm just presenting a more robust approach)

trunner/tools/phoenix.py
```diff
@@ -100,11 +123,7 @@ def _reader(self):
         self.proc.expect(pexpect.EOF, timeout=None)
 
     def _run(self):
-        try:
-            wait_for_dev(self.port, timeout=10)
-        except TimeoutError as exc:
-            raise PhoenixdError(f"couldn't find {self.port}") from exc
-
+        self.port = wait_for_vid_pid(0x16F9, 0x0003, timeout=10)
```
nalajcie, line 126: - probably you should catch exception(s) and translate them to `PhoenixdError` to remain compatible with the current API
- `wait_for_dev` is now dead code, should be removed

trunner/tools/phoenix.py
```diff
@@ -22,6 +22,30 @@ def wait_for_dev(port, timeout=0):
             raise TimeoutError
 
 
+def wait_for_vid_pid(vid: int, pid: int, timeout=0):
+    """Searcher for connected usb device via vendor & product id"""
```
nalajcie, line 26:
```suggestion
    """wait for connected usb serial device with required vendor & product id"""
```

trunner/tools/phoenix.py
```diff
@@ -101,9 +115,9 @@ def _reader(self):
 
     def _run(self):
         try:
-            wait_for_dev(self.port, timeout=10)
-        except TimeoutError as exc:
-            raise PhoenixdError(f"couldn't find {self.port}") from exc
+            self.port = wait_for_vid_pid(self.vid, self.pid, timeout=1)
```
nalajcie, line 118: You've changed timeout from 10s to 1s, is this intended? Maybe make it a separate commit so git log would clearly show that not only the lookup mechanism has changed but also the timeout itself.

maska989: Sorry, that was a mistake, I used this timeout to speed up debugging. We probably want to wait this 10 sec. because there is an option that the USB port will raise slower after some changes.

nalajcie (approved):
note for the future: dot (`.`) at the end of the commit message title is not needed

damianloew (approved):

PR #230, opened 2023-07-24
trunner: add reboot property in yaml

mateusz-bloch:
JIRA CI-303

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


trunner/config.py
```diff
@@ -43,6 +43,7 @@ class ConfigParser:
     class MainConfig:
         targets: Optional[Set] = None
         type: Optional[str] = None
+        should_reboot: bool = False
```
damianloew, line 46: Let's be consistent with other short names. I think thanks to specifying type the purpose is clear.

```suggestion
        reboot: bool = False
```

psh/test.yaml
```diff
@@ -38,6 +39,7 @@ test:
 
         - name: kill
           harness: test-kill.py
+          reboot: True
```
damianloew, line 42: I think reboot before test-gibber would be sufficient.

psh/test.yaml
```diff
@@ -6,6 +6,7 @@ test:
     tests:
         - name: gibber
           harness: test-gibber.py
+          reboot: True
```
damianloew, line 9: Please add a comment why we want to reboot there. Sth like `# psh tests need more resources, especially on smaller syspage targets, we don't want to have a test app from the previous test in the campaign loaded`

damianloew (changes requested):
Thanks for this change, please set this property for test-gibber in the separate commit (you can use the same jira id).

trunner/test_runner.py
```diff
@@ -156,6 +156,8 @@ def run_tests(self, tests: Sequence[TestOptions]):
         """
 
         fail, skip = 0, 0
+        # Reboot is required for the initial test.
```
nalajcie, line 159: and previously wasn't? Seems like it's being taken care of by sub-clause (3) below

trunner/config.py
```diff
@@ -43,6 +43,7 @@ class ConfigParser:
     class MainConfig:
         targets: Optional[Set] = None
         type: Optional[str] = None
+        reboot: bool = False
```
nalajcie, line 46: it's not a global (immutable) config property, remove!

trunner/types.py
```diff
@@ -208,7 +208,7 @@ class TestOptions:
     target: Optional[str] = None
     bootloader: Optional[BootloaderOptions] = None
     shell: Optional[ShellOptions] = None
-    should_reboot: bool = True
+    reboot: bool = False
```
nalajcie, line 211: I'm not sure if the `reboot` property means reboot BEFORE the test or AFTER the test (AFAIK right now we don't have tests which require rebooting after the test but maybe in the future we might?)

Please either name the property differently or at least add a comment

trunner/test_runner.py
```diff
@@ -191,17 +193,16 @@ def run_tests(self, tests: Sequence[TestOptions]):
...
                 # 2. The test has failed.
                 # 3. We have to enter the bootloader in order to load applications.
                 if idx == len(tests) - 1:
                     return
 
-                tests[idx + 1].should_reboot = False
-
                 if result.is_skip():
-                    tests[idx + 1].should_reboot = tests[idx].should_reboot
+                    tests[idx + 1].reboot = tests[idx].reboot
```
nalajcie, line 205: this is not entirely correct right now - IMHO the explicit "reboot before the test" property should not be forwarded to the next test on SKIP (I assume previously it might have been needed if we skipped the test which loaded necessary binary? - I'm not sure - maybe it's invalid altogether)

psh/test.yaml
```diff
@@ -6,6 +6,9 @@ test:
     tests:
         - name: gibber
           harness: test-gibber.py
+          # additional resources may be necessary for psh tests, especially on smaller syspage targets
+          # it is essential to avoid loading a test app from the previous campaign during the psh tests
```
nalajcie, line 10:
```suggestion
          # additional resources may be necessary for psh tests, especially on smaller syspage targets
          # ensure no other binaries are loaded from the previous tests during the psh tests
```

nalajcie (changes requested):

trunner/test_runner.py
damianloew: Thanks for suggestions. It was done previously, because `should_reboot` was set True by default. Reboot means also spawning an emulator process for emu targets. So we always want to reboot before starting a test campaign - for emu targets the mentioned reason, for hw targets - reboot needed after flashing.

As we can see in CI, there was always reboot before a first test and we want to leave this approach.

trunner/types.py
damianloew: I'd leave a comment to have consistent names for yaml properties.

trunner/test_runner.py
damianloew: I think the initial purpose of this line was passing `should_reboot` to next test when there was a fail and next skip. IMHO we should check whether there was a fail before and only for this case do it.

trunner/types.py
nalajcie: yeah, it was a more generic statement (wheter yaml property shouldn't be named `reboot_before` (and maybe `reboot_after` in the future if there is a need)

trunner/config.py
```diff
@@ -169,6 +169,12 @@ def _parse_shell_command(self, config: dict):
             cmd=parsed_cmd,
         )
 
+    def _parse_reboot(self, config: dict) -> None:
+        reboot = config.get("reboot", self.test.should_reboot)
+
+        if not isinstance(reboot, bool):
+            raise ParserError(f"reboot must be a boolean value (true/false) not {reboot}")
+
```
nalajcie, line 177: missing `self.test.reboot = reboot` now

trunner/test_runner.py
```diff
@@ -191,29 +194,35 @@ def run_tests(self, tests: Sequence[TestOptions]):
             def set_reboot_flag(tests, idx, result):
                 # If the test is successful and the next test doesn't require loading via
                 # the plo we don't want to reboot the entire device (to speed up the test execution).
-                # There are three exceptions to this rule:
+                # There are five exceptions to this rule:
+                # 0. The yaml config has been configured with the reboot parameter set to True.
                 # 1. Runner runs in the "nightly" mode when we are not concerned about slow execution.
                 # 2. The test has failed.
                 # 3. We have to enter the bootloader in order to load applications.
+                # 4. The test is skipped, but there should be a reboot.
```
damianloew, line 202: As we discussed:
```suggestion
                # 2. The last executed test has failed
                # 3. We have to enter the bootloader in order to load applications.
```

trunner/test_runner.py
```diff
@@ -156,6 +156,9 @@ def run_tests(self, tests: Sequence[TestOptions]):
         """
 
         fail, skip = 0, 0
+        # Reboot is required for the initial test.
+        tests[0].should_reboot = True
+        next_reboot = False
```
damianloew, line 161: I think this name could be better:

```suggestion
        reboot_next = False
```

damianloew (commented):
IMHO this solution with `reboot_next` looks good and thanks to it `should_reboot` is passed to the next test only, when there was a fail. @nalajcie could you take a look, whether it's also acceptable for you?

trunner/test_runner.py
```diff
@@ -191,29 +194,35 @@ def run_tests(self, tests: Sequence[TestOptions]):
...
                     or (
                         not self.ctx.target.rootfs
                         and tests[idx + 1].bootloader is not None
                         and tests[idx + 1].bootloader.apps
                     )
+                    or result.is_skip() and next_reboot
                 ):
                     tests[idx + 1].should_reboot = True
 
+                if not result.is_skip():
```
damianloew, line 219: I know that it's not the most optimal solution to set it to False each time, but I hope it's not a big problem. Alternatively:

```suggestion
                if not result.is_skip() and next_reboot:
```

trunner/test_runner.py
```diff
@@ -191,29 +194,34 @@ def run_tests(self, tests: Sequence[TestOptions]):
...
                 # If the test is successful and the next test doesn't require loading via
                 # the plo we don't want to reboot the entire device (to speed up the test execution).
-                # There are three exceptions to this rule:
+                # There are five exceptions to this rule:
+                # 0. The yaml config has been configured with the reboot parameter set to True.
                 # 1. Runner runs in the "nightly" mode when we are not concerned about slow execution.
-                # 2. The test has failed.
+                # 2. The last executed test has failed
                 # 3. We have to enter the bootloader in order to load applications.
+                nonlocal reboot_next
```
nalajcie, line 202: I don't like how stateless function was made stateful one with nonlocal variable - this makes code flow much less readable.

 also:
 - don't write how many exceptions (maybe "a few"), right now the number is invalid. I think if the code would be readable enough, the comment might be simplified (no need to explicitly write down each case reflected in the code)
 - (connected with the above) split the single multiline `if` with internal `or`'s into multiple ifs (one if per rule exception) to make it more readable
 - IMHO it would be more forward to call set_reboot_flag with `idx + 1` and internally use only idx (the first if can be moved to the caller) 
 - change name to `update` as it doesn't always set the `should_reboot` field

generic idea for the outer code:
```python
if result.is_fail():
    last_test_failed = True
    fail += 1
elif resut.is_success():
    last_test_failed = False
else: # result.is_skip()
    skip += 1

# [...]

if  idx + 1 != len(tests):
   update_reboot_flag(tests, idx + 1, last_test_failed)  # no need for result(?)
   # please note that this call can also be moved to the beginning of the loop (probably with if idx > 0) 
   #  - not sure if it would make it more or less readable (but more intuitive - you would be changing current test options, not the next one)
   # if the function would be short enough, consider removing it and placing code directly in the loop
```

nalajcie (approved):

damianloew (approved):
Ok, the current approach is much simpler and more readable, thanks @nalajcie. We didn't want to change a lot of things in the original code, but in this case it was worth it.

PR #237, opened 2023-08-08
libc: add math tests

adamdebek:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+}
+
+
+double test_getRandomLogSubnrm()
+{
+	double a = log(5.0e-324);
+	double b = log(2.0e-308);
+	double r = (double)rand() / (double)RAND_MAX;
+
+	return exp(a + r * (b - a));
```
agkaminski, line 42:
```suggestion
	return test_getRandomLog(5.0e-324, 2.0e-308);
```

libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+	test_common.dblSignif = 0;
+
+	/* Initialize seed for test */
+	srand(time(0));
+
+	/* Initialize maxPowTwoPi value */
+	do {
+		test_common.maxPowTwoPi++;
+		num *= 2.0;
+	} while (num - M_PI != num);
```
agkaminski, line 63: `num` starts as `M_PI`, and we loop until `num - M_PI != num`, so until `num != 2 * M_PI`.
So I see two problems here:
- there's a equality comparison between doubles, and it's somewhat lucky this loop end at all,
- `test_common.maxPowTwoPi = 1`, no need for calculating this in the runtime.

Or did I got something wrong?

libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+		test_common.maxPowTwo++;
+		num *= 2.0;
+	} while (num + 0.5 != num);
+
+	num = 1.0;
+	/* Initialize digSignif value */
+	do {
+		test_common.dblSignif++;
+		num *= 2.0;
+	} while ((((num + one) - num) - one) == zero);
```
agkaminski, line 77: This on the other hand looks ok, I think

libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+}
+
+double test_checkResult(double f, double g)
+{
+	double digSignif, digLost, diff;
+
+	if (g != 0.0) {
+		diff = fabs((f - g) / g);
+		if (diff == 1.0) {
+			diff = fabs((f - g) / f);
```
agkaminski, line 87: As this can only mean that `f == 2 *g`
```suggestion
			diff = 0.5;
```

libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+	} while ((((num + one) - num) - one) == zero);
+}
+
+double test_checkResult(double f, double g)
+{
+	double digSignif, digLost, diff;
+
+	if (g != 0.0) {
+		diff = fabs((f - g) / g);
+		if (diff == 1.0) {
```
agkaminski, line 86: Comparing floats is kinda safe only in case of 0.0, NAN and INF

libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+			diff = fabs((f - g) / f);
+		}
+	}
+	else {
+		diff = f;
+	}
+
+	if (diff != 0.0) {
```
agkaminski, line 94:
```suggestion
	if (diff > 0.0) {
```

libc/math/common.h
```diff
@@ -0,0 +1,57 @@
...
+		char errStr[150]; \
+		if (digLost > acceptLoss) { \
+			sprintf(errStr, _fun "(%g) gave lost of %.f significant digits of base 2 where " \
+								 "maximal acceptable loss is %.f signicant digits of base 2", \
+				_x, _digLost, _acceptLoss); \
+			TEST_FAIL_MESSAGE(errStr); \
+		} \
+	} while (0)
+
+#define test_check_digLost2(_fun, _x, _y, _digLost, _acceptLoss) \
```
agkaminski, line 32: Why macros instead of functions?

libc/math/common.h
```diff
@@ -0,0 +1,57 @@
...
+				_x, _y, _digLost, _acceptLoss); \
+			TEST_FAIL_MESSAGE(errStr); \
+		} \
+	} while (0)
+
+extern struct {
+	int maxPowTwoPi;
+	int maxPowTwo;
+	int dblSignif;
+} test_common;
```
agkaminski, line 47: So the struct definition is in one place only
```suggestion
struct test_common_s {
	int maxPowTwoPi;
	int maxPowTwo;
	int dblSignif;
};

extern struct test_common_s test_common;
```

libc/math/main.c
```diff
@@ -0,0 +1,41 @@
...
+ * %LICENSE%
+ */
+
+#include "unity_fixture.h"
+
+struct {
+	int maxPowTwoPi;
+	int maxPowTwo;
+	int dblSignif;
+} test_common;
```
agkaminski, line 22:
```suggestion
struct test_common_s test_common;
```

libc/math/exp.c
```diff
@@ -0,0 +1,255 @@
...
+{
+	int i, iterrs = 100 * ITERR_FACTOR;
+	double acceptLoss = 50.0;
+	double xmax = log(DBL_MAX);
+	double xmin = 1.0;
+	double x, y, ymin, ymax, f, g, digLost;
+
+	for (i = 0; i < iterrs; i++) {
+		x = test_getRandomLog(xmin, xmax);
+		if (x < 0.0) {
```
agkaminski, line 53: Not possible, `exp()` is used to return the result

libc/math/exp.c
```diff
@@ -0,0 +1,255 @@
...
+			sprintf(errStr, "frexp(%g, int *exp) returned %g - value out of range <0.5, 1)", x, y);
+			TEST_FAIL_MESSAGE(errStr);
+		}
+
+		if (e > 0.0) {
+			f = y * pow(2.0, e);
+		}
+		else {
+			f = y / pow(2.0, -e);
+		}
```
agkaminski, line 105:
```suggestion
		f = y / pow(2,0, e);
```

libc/math/frac.c
```diff
@@ -0,0 +1,303 @@
...
+#include "common.h"
+#include <unity_fixture.h>
+
+static double test_ceilAndFloor(double x, int type)
+{
+	double powTwo, frac = x;
+
+	if (frac < 0.0) {
+		frac = -frac;
+	}
```
agkaminski, line 37:
```suggestion
	double powTwo, frac = fabs(x);
```

libc/math/frac.c
```diff
@@ -0,0 +1,303 @@
...
+
+	for (powTwo = pow(2.0, test_common.maxPowTwo); powTwo >= 2.0; powTwo /= 2.0) {
+		while (frac > powTwo) {
+			frac -= powTwo;
+		}
+	}
+
+	while (frac >= 1.0) {
+		frac -= 1.0;
+	}
```
agkaminski, line 51: Yeah, it uses division, which is slow, but your code too (in for loop).

I guess you want to avoid using `frexp()`?
```suggestion
	while (frac >= 1.0) {
		frac /= 2.0;
	}
```

libc/math/frac.c
```diff
@@ -0,0 +1,303 @@
...
+#include <limits.h>
+#include <float.h>
+#include <time.h>
+#include <stdlib.h>
+#include <errno.h>
+
+#include "common.h"
+#include <unity_fixture.h>
+
+static double test_ceilAndFloor(double x, int type)
```
agkaminski, line 31: Need to use `and` in function name is a good hint something should be changed.

My suggestion - provide separate `ceil` and `floor` and share common code

libc/math/pow.c
```diff
@@ -0,0 +1,145 @@
...
+TEST_GROUP_RUNNER(math_pow)
+{
+	test_setup();
+
+	RUN_TEST_CASE(math_pow, sqrt_basic);
+	RUN_TEST_CASE(math_pow, sqrt_special_cond);
+
+	RUN_TEST_CASE(math_pow, pow_basic);
+	RUN_TEST_CASE(math_pow, pow_special_cond);
+}
```
agkaminski, line 145: No new line at the end of the file

libc/math/trig.c
```diff
@@ -0,0 +1,590 @@
...
+	double min = -max;
+	double step = max / iterrs - min / iterrs;
+	double xstart = min;
+	double u, x, y, f, g, digLost;
+
+	for (i = 0; i < iterrs; i++) {
+		u = (double)rand() / (double)RAND_MAX;
+		x = step * u + xstart;
+		y = sin(x / 3.0);
+		f = (3.0 * y) - (4.0 * y * y * y);
```
agkaminski, line 61: That's clever. Perhaps some comments about used identities would be nice?

libc/math/common.c
adamdebek: This loop runs until `num` will be such big number that subtracting `M_PI` will result in same number, as I remember it is `test_common.maxPowTwoPi = 53` , this value is needed to tests with periodic funcs eg. `sins(x)` where big `x` is normalized to `<0, 2 * M_PI>` , potential wrong implementation could cause infinity loop continuously subtracting `M_PI` from `x`.

agkaminski: Oh, ok, I misunderstood the code indeed.

libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+		test_common.maxPowTwoPi++;
+		num *= 2.0;
+	} while (num - M_PI != num);
+
+	num = 1.0;
+	/* Initialize maxPowTwo value */
+	do {
+		test_common.maxPowTwo++;
+		num *= 2.0;
+	} while (num + 0.5 != num);
```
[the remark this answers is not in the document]
adamdebek: This loop runs until `num` is big enough that adding fractional numbers do not change `num` value , used in `ceil()` and `floor()`.

libc/math/common.c
adamdebek: Not only, this is done to scenarios when `g` is much bigger than `f` eg. `g = 10^100` and `f = 10^10` then
`diff = fabs((f - g) / g)` gives `1.0` and `diff = fabs((f - g) / f)` gives `10^90` which can indicate when this numbers start to differ.

libc/math/common.h
adamdebek: My misconception.

libc/math/exp.c
adamdebek: I think you meant `f = y * pow(2.0, e)` but i would stick to as it is, cause `pow(2.0, e)` for negative values is more flat and could return less precise values.

libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+
+	return exp(a + r * (b - a));
+}
+
+
+void test_setup(void)
+{
+	double num = M_PI;
+	double one = 1.0;
+	double zero = 0.0;
```
anglov, line 50: Why not literals? (according to c99 6.4.4.2. p 4. unsuffixed floating constant should be double?

In case that it have to be variable, you should constify it.

libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+		num *= 2.0;
+	} while ((((num + one) - num) - one) == zero);
+}
+
+double test_checkResult(double f, double g)
+{
+	double digSignif, digLost, diff;
+
+	if (g != 0.0) {
+		diff = fabs((f - g) / g);
```
anglov, line 85: Hm, in fabs test you use test_checkResult function, in test_checkResult you use fabs. I don't looking that carefully but please be sure that this doesn't cause that test will be malfunctioning. If It's asserted somehow, please place comment about it how it's achieved.

anglov (commented):
For now I looked mainly on common part and on abs test a little.

libc/math/common.c
```diff
@@ -0,0 +1,109 @@
...
+	double num = M_PI;
+	double one = 1.0;
+	double zero = 0.0;
+
+	test_common.maxPowTwoPi = 0;
+	test_common.maxPowTwo = 0;
+	test_common.dblSignif = 0;
+
+	/* Initialize seed for test */
+	srand(time(0));
```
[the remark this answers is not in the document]
agkaminski: I opt for static rand seed too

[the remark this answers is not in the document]
adamdebek: In case of fail, name of function with arguments it was invoked are printed in error message, so reproduction is pretty easy.
Thanks to randomness wider range of arguments is tested.

anglov (commented):
Sorry for late comment - I tried to review it before but each time I come to this I have sense that your solution is nice and clever but don't accually test enough function.

Eg. exp tests: To prove that your test is not enough I create implementation like this:

```c
double exp(double x)
{
	if (x == NAN) {
		return NAN;
	}
	else if (x == INFINITY) {
		return INFINITY;
	}
	else if (x == -INFINITY) {
		return 0.0;
	}
	else {
		return 1.0;
	}
}
```

And it pass all exp test but fails on trig and pow.

All of tested function somehow depends on each other (or even on themself) - log example).

I think that your approach is intersting for providing some random test so increase probability of finding issues. But I think test is not enough for testing those functions.

I also think why you want to count imprecision of double in runtime . Isn't that IEC 60559  defined?

TBH I would think about some static vectors to be checked as basic function test. And then you may want to add random identities tests but with knowledge, that those tests uses also others (eg. asin_basic uses asin and sin etc.)

adamdebek:
I could add check for values that cause identity to be always equal so it would find tricky implementation like that.

I see that some function depend on each other, I could try to arrange testcases in the way that functions which are independent first and if they fail testcases depending on them are skipped.

Imprecision is not standardized, I check it because there could always be differences in FPU architecture and precision of algorithms.

Static vectors could be added, to be considered.

anglov:
> I could add check for values that cause identity to be always equal so it would find tricky implementation like that.
> 

Well, I think that indentities test don't have to check that. It's not the purpose if this tests. 
But I think you should have other test to be sure that function  implementation looks good in most cases.

> I see that some function depend on each other, I could try to arrange testcases in the way that functions which are independent first and if they fail testcases depending on them are skipped.
> 

TBH I don't know best test practises. I think that it's not a bad approach to run always everything. But I think that you indeed need independend test which makes sure that function is implemented correctly. It shouldn't use 

> Imprecision is not standardized, I check it because there could always be differences in FPU architecture and precision of algorithms.
> 

OK thanks for clarification. I tought that float is well described as it's provided by the compiler mostly. 

> Static vectors could be added, to be considered.

If it's possible to do, I believe it can be good idea. I don't think if we can assure that function indeed looks like it work correctly in the other way without using function itsels/using other functions. I'm really concerned about situation when eg. some invalid implementation will start return static variables or weird variables which accidentally is validated by identities.

libc/math/static-vector-data.h
```diff
@@ -0,0 +1,1540 @@
...
+
+#include <float.h>
+
+/*
+ * DO NOT INCLUDE THIS FILE ANYWHERE ELSE EXCEPT STATIC-VECTOR.C !!!!!!!
+ * DO NOT INCLUDE THIS FILE ANYWHERE ELSE EXCEPT STATIC-VECTOR.C !!!!!!!
+ * DO NOT INCLUDE THIS FILE ANYWHERE ELSE EXCEPT STATIC-VECTOR.C !!!!!!!
+ * DO NOT INCLUDE THIS FILE ANYWHERE ELSE EXCEPT STATIC-VECTOR.C !!!!!!!
+ * DO NOT INCLUDE THIS FILE ANYWHERE ELSE EXCEPT STATIC-VECTOR.C !!!!!!!
+ */
```
agkaminski, line 24: No point in screaming if noone is listening. How about:
```
#ifndef TEST_USE_STATIC_VERTOR_H_
#error Do not include static-vector-data.h anywhere except static-vector.c
#endif
```
+ `define TEST_USE_STATIC_VERTOR_H_` in static-vector.c

libc/math/trig.c
agkaminski: I guess this comments is not addressed yet

libc/math/static-vector-data.h
anglov: Or potential size getters for vectors and declare it as extern

anglov (commented):
I rather look over the code than do full review, but in general looks ok. 

As log function which are used everywhere else are tested separatly then I'm kinda ok with current approach. 

I think about static vectors for each functions (maybe smaller? Or maybe smaller for small targets?) and identities tests as additional bigger testcase just for being sure that functions looks like it work good. 

As we talked, let's try to merge current approach and think about enchancement later.

libc/math/trig.c
```diff
@@ -0,0 +1,697 @@
...
+	double max = M_PI_2;
+	double min = -max;
+	double step = max / iters - min / iters;
+	double xstart = min;
+	double u, x, y, f, g, digLost;
+
+	for (i = 0; i < iters; i++) {
+		u = (double)rand() / (double)RAND_MAX;
+		x = step * u + xstart;
+		/* Using sinus trigonometric identity */
```
agkaminski, line 60: What identity exactly? Would be nice to add e.g. equations used in all such instances for a future reference. It's not 100% obvious from the code alone (i.e. requires some analysis from the reader to find out identity used).

adamdebek: These identities do not have particular names and I think they are pretty self-explanatory, in most cases `g` is given function and `f` represents identity, which has to be equal.

adamdebek: There are many places where code looks similar but there are often few differences, what would be hard to abstract it under single function with many arguments and it would cause readability decrease IMHO.

damianloew (commented):
Can we merge it in the current form? @agkaminski, @anglov is it acceptable for you?

agkaminski (approved):

PR #254, opened 2023-09-18
Maska989/ping_test

maska989:
## Description
JIRA: CI-326

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


psh/test-ping.py
```diff
@@ -0,0 +1,206 @@
...
+    # invalid timeout
+    psh.assert_cmd(p, f"ping -W -1 {ping_target}", expected="ping: Wrong timeout value!", result="fail", is_regex=False)
+
+    # invlid_payload_len
+    psh.assert_cmd(p, f"ping -s 2041 {ping_target}", expected="ping: Wrong payload len", result="fail", is_regex=False)
+
+
+def host_ping():
+    global target_ip
+    subprocess.call("echo", shell=True)
```
mateusz-bloch, line 180: Why this is needed?
```suggestion
```

maska989:
TWM for writing about generic remarks, all of this needs rethink about test structure and runner options.
About: 
>     * please don't reformat python files unnecessarily

Rn I'm using this PR to research/configure Ethernet support on `QEMU ia32` because CI if even run network.sh to set up still ping don't pass. 
That's why still this test isn't converted to use just `ifconfig`.  
My fault was not converted this PR to draft, probably after full research I will delete this PR and create new one (For focus strictly on ping test).

Again, thanks 4 all your help.

nalajcie:
> Rn I'm using this PR to research/configure Ethernet support on `QEMU ia32` because CI if even run network.sh to set up still ping don't pass. [https://github.com/phoenix-rtos/phoenix-rtos-project/pull/852](url) That's why still this test isn't converted to use just `ifconfig`. My fault was not converted this PR to draft, probably after full research I will delete this PR and create new one (or clear a little bit for focus strictly on ping test).

IMHO using `ifconfig` directly would make experimenting clearer. From what I see:
 - you're using DHCP setup right now  - so that won't work in qemu `bridged` mode straight away (no DHCP server in GH network would want to give you an address due to security reasons)
 - the `dhcp` could easily be used in `user` qemu set-up but this won't allow us to reach WAN (still, `ping 10.0.2.2`  works (default for qemu server, can be configured by params as described in help))
 - we probably should set-up `bridge` networking with NAT/private DHCP (should be possible in docker, not sure about tap interface tho) if we need access to WAN
 - all the above network-setup tests can be done locally (in docker - you can disconnect it from network)
 - if network setup would still fail in CI after the local testing succeeds - I would advice adding ssh tunnel to CI and experiment interactively in runner (eg: https://github.com/mxschmitt/action-tmate)
 - for PING tests we don't need access to WAN, we could simply use only LAN addresses (eg. the address of the host running test runner)
 - nevertheless, setup with WAN access might be beneficial in the future in other tests (eg. ntp client, wget, etc. -> we won't need to set up local instances of various servers on the test runner host ;))

maska989:
Right now this PR waiting for ifconfig and route implementation.

psh/test.yaml
```diff
@@ -68,7 +68,7 @@ test:
           harness: test-runfile.py
           targets:
               # runfile applet is not intended for non-rootfs targets
-              exclude: [armv7m7-imxrt106x-evk, armv7m7-imxrt117x-evk, armv7m4-stm32l4x6-nucleo, riscv64-generic-qemu]
+              exclude: [armv7m7-imxrt106x-evk, armv7m7-imxrt117x-evk, armv7m4-stm32l4x6-nucleo]
```
mateusz-bloch, line 71: Why did you remove risc?

psh/test-hw-ping-wan.py
```diff
@@ -0,0 +1,192 @@
...
+    psh_output_regex += r": ttl=[0-9]+ icmp_seq=[0-9]+ time=([0-9.]+) ms)"
+    psh_output_regex += f"{psh.EOL}){iter_str}"
+
+    return psh_output_regex
+
+
+def ping_error_regex(error_msg):
+    ping_error_regex = r"PING [()0-9a-z.: ]+ data bytes" + psh.EOL + error_msg + psh.EOL
+
+    return ping_error_regex
```
mateusz-bloch, line 37:
```suggestion
def ping_error_regex(error_msg):
    return rf"PING [()0-9a-z.: ]+ data bytes{psh.EOL}{error_msg}{psh.EOL}"
```

psh/test-hw-ping-wan.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def ping_output_regex(target, iteration):
+    iter_str = r"{" + str(iteration) + r"}"
+
+    psh_output_regex = r"PING [()0-9a-z-.: ]+ data bytes" + psh.EOL
+    psh_output_regex += r"(([0-9]+[a-z\s]+ "
+    psh_output_regex += str(target)
+    psh_output_regex += r": ttl=[0-9]+ icmp_seq=[0-9]+ time=([0-9.]+) ms)"
+    psh_output_regex += f"{psh.EOL}){iter_str}"
```
mateusz-bloch, line 29:
```suggestion
    psh_output_regex = ( 
    		rf"PING [()0-9a-z-.: ]+ data bytes{psh.EOL}"
    		rf"(([0-9]+[a-z\s]+ {target}:  ttl=[0-9]+ icmp_seq=[0-9]+ time=([0-9.]+) ms)"
		rf": ttl=[0-9]+ icmp_seq=[0-9]+ time=([0-9.]+) ms)"
    		rf"{psh.EOL}){iter_str}"
    )
```

psh/tools/netset.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def ip_table():
+    possible_ip = (
+        "192.168.72.20",
+        "192.168.72.21",
+        "192.168.72.120",
+        "192.168.72.121",
+    )
+    return possible_ip
```
mateusz-bloch, line 29:
```suggestion
from typing import Tuple

def ip_table() -> Tuple[str, ...]:
    return (
        "192.168.72.20",
        "192.168.72.21",
        "192.168.72.120",
        "192.168.72.121",
    )
```

psh/tools/netset.py
```diff
@@ -0,0 +1,98 @@
...
+    p,
+    eth="en1",
+    up: Optional[bool] = True,
+    wan: Optional[bool] = False,
+    ip: Optional[str] = None,
+    dynamic: Optional[bool] = False,
+):
+    host_setup = host_interface()
+    host_ip = host_setup[0]
+    host_netmask = host_setup[1]
```
mateusz-bloch, line 58:
```suggestion
    host_ip, host_netmask = host_interface()
```

mateusz-bloch (commented):
Several  morelike cosmetic improvements, you don't have to apply them if you don't want to

psh/tools/netset.py
```diff
@@ -0,0 +1,166 @@
...
+    time.sleep(1)
+
+
+def pppd_setup(
+    p,
+    baudrate: Optional[int] = 460800,
+    port: Optional[str] = "/dev/ttyUSB0",
+    host_ip: Optional[str] = "10.0.0.1",
+    target_ip: Optional[str] = "10.0.0.2",
+    interface_idle_time: Optional[int] = "10",
```
mateusz-bloch, line 112: Parameter is declared as an Optional[int] but has a default value of type str
```suggestion
    interface_idle_time: Optional[int] = 10,
```

psh/tools/netset.py
```diff
@@ -0,0 +1,166 @@
...
+        stderr=subprocess.STDOUT,
+        text=True,
+    )
+
+    asleep = 0
+    ppd_devices = []
+
+    while not ppd_devices:
+        time.sleep(0.01)
+        asleep += 0.01
```
mateusz-bloch, line 149: inefficient timeout ? Is it worth invoking ifconfig every 0.01 second?

psh/tools/netset.py
```diff
@@ -0,0 +1,166 @@
...
+        stderr=subprocess.STDOUT,
+        check=True,
+    )
+
+    eth_device = re.findall("eth[0-9]:.+\n.+[0-9]+", out.stdout)
+    inet = re.findall("inet [0-9.]+", str(eth_device))
+    netmask = re.findall("netmask [0-9.]+", str(eth_device))
+
+    host_setup = inet + netmask
+    return host_setup
```
mateusz-bloch, line 46: maybe, but do whatever you think is more readable
```suggestion
    return  inet + netmask
```

psh/test-hw-ping-wan.py
```diff
@@ -0,0 +1,205 @@
...
+import psh.tools.psh as psh
+import psh.tools.netset as netset
+from typing import Optional
+
+CLOUD_FLARE_DNS = "1.1.1.1"
+GOOGLE_DNS = "8.8.8.8"
+IP_LOOPBACK = "127.0.0.1"
+
+
+def ping_output_regex(target, iteration, possible_fail: Optional[bool] = False):
```
mateusz-bloch, line 24: Is `Optional` needed ?

PR #256, opened 2023-09-25
psh: tests improvements

mateusz-bloch:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/psh/test-cat.py
+++ b/psh/test-cat.py
@@ -4,8 +4,8 @@
 #
 # psh cat command test
 #
-# Copyright 2021, 2022 Phoenix Systems
-# Author: Damian Loewnau
+# Copyright 2021-2023 Phoenix Systems
+# Author: Damian Loewnau, Mateusz Bloch
 #
 # This file is part of Phoenix-RTOS.
 #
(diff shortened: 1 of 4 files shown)


psh/test-cat.py
```diff
@@ -27,8 +25,61 @@ def assert_cat_h(p):
...
 
 
+def assert_cat_err(p):
+    fname = 'nonexistentFile'
+
+    statement = f'cat: {fname}: ENOENT'
+    psh.assert_cmd(p, f'cat {fname}', expected=statement, result='fail')
+
+    statement = rf'cat: {ROOT_TEST_DIR}: EISDIR'
+    psh.assert_cmd(p, f'cat {ROOT_TEST_DIR}', expected=statement, result='fail')
```
damianloew, line 35: We want to leave it flexible (for example on linux the behavior is a little but different: `cat: nonexistent: No such file or directory` - maybe we will change the approach in the future). However checking cat on directory is a good idea and we can leave it.

psh/test-cat.py
```diff
@@ -27,8 +25,61 @@ def assert_cat_h(p):
...
+
+
+def assert_cat_basic(p):
+    fname = ROOT_TEST_DIR + "/textFile.txt"
+    secFname = ROOT_TEST_DIR + "/textFile2.txt"
+    fspecial = ROOT_TEST_DIR + "/text!#$?@File.txt"
+
+    text = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. "
+    secText = "Curabitur eu velit a sem porttitor blandit."
+    ascii = "0123456789?@ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
```
damianloew, line 45: Let's check all printable characters using `string.printable`

psh/test-cat.py
```diff
@@ -27,8 +25,61 @@ def assert_cat_h(p):
...
+
+    # Printing content of file
+    psh.assert_cmd(p, f'echo "{text}" > {fname}')
+    psh.assert_cmd(p, f'cat {fname}', expected=text)
+
+    psh.assert_cmd(p, f'echo "{secText}" > {secFname}')
+    psh.assert_cmd(p, f'cat {secFname}', expected=secText)
+
+    # Concrate two text files
+    psh.assert_cmd(p, f'cat {fname} {secFname}', expected=[text, secText])
```
damianloew, line 62: Let's use tuple if possible.

```suggestion
    psh.assert_cmd(p, f'cat {fname} {secFname}', expected=(text, secText))
```

damianloew: Applies to the other places.

psh/test-cat.py
```diff
@@ -27,8 +25,61 @@ def assert_cat_h(p):
...
+
+    # Print basic characters
+    psh.assert_cmd(p, f'echo "{ascii}" > {secFname}')
+    psh.assert_cmd(p, f'cat {secFname}', expected=ascii)
+
+    # Concrate existing file with non existing
+    statement = 'cat: nonexistentFile: ENOENT'
```
damianloew, line 69: Please use a regex to be more flexible.

damianloew (changes requested):
Thanks or those improvements! Please be more specific in commit messages if possible.

psh/test-cat.py
mateusz-bloch: So, yes, it is possible to use a larger number of characters. I have now added more. 

In theory, by using the echo command, we can write them all thanks to the `-e` argument in echo. However, it is not available on Phoenix.

damianloew: how about whitespaces? We can also test writing them.

psh/test.yaml
```diff
@@ -52,14 +52,10 @@ test:
           harness: test-mkdir.py
 
         - name: touch
-          # TODO touch of netsocket and pf fails...
           harness: test-touch.py
-          ignore: True
 
         - name: touch-rootfs
           harness: test-touch-rootfs.py
-          # temporary disabled because of the following issue: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/621
```
damianloew, line 61: Please remove it as it's done in the another PR.

damianloew (changes requested):
I'd suggest splitting those changes to the following commits:

 - `psh: cat test improvements` -> `psh: test-cat: add cases with writing a content`

- `psh: add cleanup in tests`

- `psh: enable back test-touch`

- `psh: small improvements and cleanup in tests`


IMHO it will be much more readable in commit messages what have been changed.

damianloew:
> I'd suggest splitting those changes to the following commits:
> 
> * `psh: cat test improvements` -> `psh: test-cat: add cases with writing a content`
> * `psh: add cleanup in tests`
> * `psh: enable back test-touch`
> * `psh: small improvements and cleanup in tests`
> 
> IMHO it will be much more readable in commit messages what have been changed.

Of course you can change the messages a little bit if you want

damianloew (approved):
LGTM

PR #260, opened 2023-09-27
psh: fixed tests to allow year printing in `ls -l`

jmaksymowicz:
## Description
`ls -l` command currently always prints the modification time in `HH:MM` format. These changes allow for modification year in `YYYY` format to be printed instead of the time if the modification year is different from the current year.

This change is compatible with current version of `psh` and doesn't break existing tests.

## Motivation and Context
Fixing https://github.com/phoenix-rtos/phoenix-rtos-project/issues/552 would result in current tests not passing because `ls -l` would be printing the modification year sometimes.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic-qemu

## Checklist:

## Special treatment

### DIFF
--- a/psh/tools/common.py
+++ b/psh/tools/common.py
@@ -49,22 +49,21 @@ def create_testdir(p, dirname):
 def assert_mtime(p, datetimes: Dict[str, datetime], dir=''):
     ''' Asserts that files (keys in datetimes dictionary) have modification time
     equal to corresponding datetime values with 1 min margin.
-    Year and seconds are not checked. All files shall be present in the `dir` directory. '''
+    Seconds are not checked. If file's modification year is different from year
+    currently set on the device, the hours and minutes will not be checked either. '''
 
     dir_files = {file.name: file for file in psh.ls(p, dir)}
 
     for filename, target_datetime in datetimes.items():
         assert filename in dir_files, f'File {filename} is not present in the {dir} directory!'
-        date = dir_files[filename].datetime
+        date: datetime = dir_files[filename].datetime
+        date_resolution: timedelta = dir_files[filename].datetime_resolution
 
-        # we do not want to compare years and seconds (not printed by ls -l)
-        date = date.replace(year=target_datetime.year, second=target_datetime.second)
+        delta = target_datetime - date
 
-        # to prevent failed assertion when typing date in 00:59 and touch in 01:00
-        if date - target_datetime == timedelta(minutes=1):
-            date = target_datetime
-
-        assert date == target_datetime, "".join((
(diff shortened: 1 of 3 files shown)


nalajcie:
please use imperative mood in commit message title

psh/tools/psh.py
```diff
@@ -195,15 +195,24 @@ def ls(pexpect_proc, dir=''):
...
         # Name is printed with ascii escape characters - remove them
         name = re.sub(CONTROL_CODE, '', name)
 
-        # the year is set to default 1970, 1 min resolution
-        file_datetime = datetime.strptime(f'1970 {month} {mday} {time}:00', '%Y %b %d %X')
+        # time_year can contain either time (HH:MM) or year (YYYY)
+        # If year was not printed, replace it with default
+        if ':' in time_year:
+            time = time_year
+            year = default_year
```
damianloew, line 209: Then we could use `date`.

psh/tools/psh.py
```diff
@@ -195,15 +195,24 @@ def ls(pexpect_proc, dir=''):
...
+        else:
+            time = '00:00'
```
damianloew, line 211: Hmm, that's not 00:00 in fact, but we could leave it with a proper comment showing that the exact value does not matter when we are comparing two dates without a specific HH:MM

damianloew (changes requested):
Thanks for the improvements in ls. Some thoughts left.

psh/tools/common.py
```diff
@@ -46,19 +46,30 @@ def create_testdir(p, dirname):
...
+    one entry `datetimes`. All files shall be present in the `dir` directory. '''
+    if not datetimes:
+        return
 
-    dir_files = {file.name: file for file in psh.ls(p, dir)}
+    if current_year is None:
+        # get the year of any entry
+        current_year = next(iter(datetimes.values())).year
+
+    dir_files = {file.name: file for file in psh.ls(p, dir, current_year)}
```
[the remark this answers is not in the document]
jmaksymowicz: In new version I used the `date` applet to get the current date.

psh/tools/psh.py
jmaksymowicz: I solved this by passing the timestamp resolution along with the timestamp, so it can be compared correctly by `assert_mtime`.

jmaksymowicz:
I also fixed a bug in the tests - `touch`-ing one file will cause all other files linked to it to change date too, but the tests didn't account for that and expected files to stay the same date when one of the links was touched. I solved this by parsing the number of links from `ls` output and changing the test to only check files without additional links.

PR #267, opened 2023-10-05
psh: avoid SH syntax in test strings

damianloew:
JIRA: CI-357

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/psh/tools/common.py
+++ b/psh/tools/common.py
@@ -27,7 +27,8 @@ SEPARATOR_PATTERN_BASE = r'(?:' + CONTROL_CODE + r'|\s)'
 SEPARATOR_PATTERN = SEPARATOR_PATTERN_BASE + r'+?'
 # acceptable patterns: white spaces (wss) + CC, CC + wss, wss, nothing
 OPTIONAL_SEPARATOR_PATTERN = SEPARATOR_PATTERN_BASE + r'*?'
-CHARS = list(set(string.printable) - set(string.whitespace) - set('/'))
+# all printable chars excluding whitespaces and SH syntax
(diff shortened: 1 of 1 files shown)


psh/tools/common.py
```diff
@@ -27,7 +27,8 @@
 SEPARATOR_PATTERN = SEPARATOR_PATTERN_BASE + r'+?'
 # acceptable patterns: white spaces (wss) + CC, CC + wss, wss, nothing
 OPTIONAL_SEPARATOR_PATTERN = SEPARATOR_PATTERN_BASE + r'*?'
-CHARS = list(set(string.printable) - set(string.whitespace) - set('/'))
+# all printable chars excluding whitespaces and SH syntax
+CHARS = list(set(string.printable) - set(string.whitespace) - set('/<>;|&'))
```
nalajcie, line 31: I would add also:
- `()` (subshells?)
- `{}` (control blocks, variable substitiution)
- `[]` (test syntax) 
- `$` (variables)
- `!` (previous commands)
- `*` (to avoid globbing with existing files)
- `\` (beginning of escape sequence so next chars might be treated differently)

bash man page mentions the following special chars:
image

damianloew: Oh that's right, thanks!

mateusz-bloch (approved):
LGTM

psh/tools/common.py
nalajcie: now that I think - we should also omit `'`, `"` and backtick (`).

idea: maybe instead of excluding special chars we should just explicitly include some chars instead of thinking what we should exclude?

nalajcie: Missed my last comment before the merge?

damianloew: Yes, sorry for that I am now considering change in the next PR

PR #270, opened 2023-10-10
mbedtls: fixed correct recording of test durations to xml file

mateusz-bloch:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


mbedtls/mbedtls.py
```diff
@@ -41,6 +31,15 @@ def harness(dut: Dut, ctx: TestContext, result: TestResult):
             # if there are dots after test name - remove them
             subname = re.sub(r" \.+ ", "", parsed["name"])
             subresult = TestSubResult(subname=subname, status=status)
+
+            if msg and subresult.status == Status.FAIL:
```
nalajcie, line 35: I'm not sure but this would result in appending extra message to invalid testcase (the next one) - assuming the format is:
```
[test name].... FAIL
   [extra messages why it failed]
[test name 2].... SUCCESS
```


IMHO this could be simplified by not resetting `subresult` variable and appending to it directly, eg.

```python
    while True:
        idx = dut.expect([RESULT, FINAL, MSG_LINE], timeout=90)

        if idx == 2:
            # append extra message to the last test (if available)
            if subresult:
                subresult.msg += parsed["line"] + " "

        if idx == 0:
            # the subresult finished running - add it to result now to ensure correct run time
            if parsed["status"] == "----":
                status = Status.SKIP
            else:
                status = Status.from_str(parsed["status"])

            if status == Status.FAIL:
                test_status = Status.FAIL

            # if there are dots after test name - remove them
            subname = re.sub(r" \.+ ", "", parsed["name"])
            subresult = TestSubResult(subname=subname, status=status)
            result.add_subresult_obj(subresult)

        if idx == 1:
            break

    return TestResult(status=test_status)
```

busybox/busybox.py
```diff
@@ -29,8 +29,9 @@ def harness(dut: Dut, ctx: TestContext, result: TestResult):
             msg.append("\t\t" + parsed["line"])
             continue
 
-        if subresult:
-            # We ended processing test result and message
+        if idx == 0:
```
nalajcie, line 32: same as with the mbedtls - msg would be appended to the next subresult instead of the correct one. Similar approach to the one proposed for mbedtls needs to be applied here

(and some testing with tests-failing-on-purpose to ensure messagess are appended to the correct testcase :))

nalajcie (changes requested):

mbedtls/mbedtls.py
```diff
@@ -7,32 +7,17 @@
 
 def harness(dut: Dut, ctx: TestContext, result: TestResult):
     subresult = None
-    msg = []
     test_status = Status.OK
 
-    RESULT = r"(?P<name>.{67}\s)(?P<status>(PASS|----|FAILED))\r+\n"
-    MSG_LINE = r"  (?P<line>[^\r\n]+?)\r+\n"
+    RESULT = r"(?P<name>.{67}\s)(?P<status>(PASS|----))\r+\n"
+    RESULT_FAIL = r"(?P<name>.{67}\s)(?P<status>(FAILED))\r+\n (?P<line>[^\r\n]+?)\r+\n"
```
nalajcie, line 13: IMHO multiline regexps are not a good idea, they might break in unexpected ways depending on how they are applied or might result in exponential execution time (eg. greedy approach to matching regexp might parse next test as a message line). If we can use single-line regexps + state machine (like previously) it's much better (and easier to understand at a glance).

Does my previously proposed solution not work in some cases?

mateusz-bloch: I agree, a single-line regex is better. I’ve just sent another approach.  The general issue is that by the time we collect the test completion time, we must already have a message, especially if the test failed. That's why add_subresult_obj was executed after the completion of the next test. This resulted in the test time being collected from the subsequent test.

nalajcie: ok, I see what's the issue with my approach, the subresult in `trunner` is being added by-value instead of by-reference, probably the best solution would be to change that in test runner as it would simplify the harness implementation greatly.

Should be easy to fix, I can think for a clean solution for that soon.

nalajcie: Please check if this solution would work  in regards to the messages printed from the test cases. - I've changed the API and applied a quickfix only for mbedtls: 
https://github.com/phoenix-rtos/phoenix-rtos-tests/tree/nalajcie/trunner

mateusz-bloch: So, I've tested it. Now, in the case of mbedtls, both the messages and the timing are correctly appended. The only issue is that if there's a fail and any fail messages follow it, the time between the end of the test case and the collection of the message will be added to the next test. But from what I see, it's only a maximum of 0.-0.15 seconds, so I would consider that an acceptable range

nalajcie: Tanks for testing. I consider this a negligible time difference and IMHO the simplicity of the solution is worth it. 

I'll make my branch a PR, please add your commit to this branch (`nalajcie/trunner`) with necessary changes in busybox (yep, we would be PR co-authors ;)).

PR #272, opened 2023-10-16
trunner: provide traceback for internal exceptions when waiting for plo

damianloew:
## Description

Before:
Screenshot from 2023-10-16 10-42-39

After:
Screenshot from 2023-10-25 11-56-48

The mentioned issue has been described on Confluence in Github Known Issues - point 2.5.
## Motivation and Context

Provide more information about internal serial exceptions.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/tools/phoenix.py
+++ b/trunner/tools/phoenix.py
@@ -3,9 +3,11 @@ import os
 import signal
 import threading
 import time
+import traceback
 import pexpect
 
 from trunner.harness import ProcessError
+from typing import Optional
 from serial.tools import list_ports
 from contextlib import contextmanager
 from .common import add_output_to_exception
@@ -24,7 +26,7 @@ def wait_for_vid_pid(vid: int, pid: int, timeout=0):
         found_ports = [port for port in list_ports.comports() if port.pid == pid and port.vid == vid]
 
         if len(found_ports) > 1:
-            raise Exception(
+            raise ValueError(
                 "More than one plo port was found! Maybe more than one device is connected? Hint used to find port:"
                 f"{vid:04x}:{pid:04x}"
             )
@@ -79,6 +81,10 @@ class Psu:
 class PhoenixdError(ProcessError):
     name = "PHOENIXD"
 
+    def __init__(self, msg: str = "", output: Optional[str] = None, traceback: str = None):
+        super().__init__(msg, output)
+        self.traceback = traceback
+
 
 class Phoenixd:
     """Handler for phoenixd process"""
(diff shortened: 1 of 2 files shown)


nalajcie:
IMHO this should be solved in other way (maybe change how phoenixd exception is printed depending on verbosity in raw logs / test summary. Wrapping all phoenixd errors as `PhoenixException` is a nice way of abstracting things and makes it easier to use.

Please note that the original exception traceback is not lost, just not explicitly printed.

trunner/tools/phoenix.py
```diff
@@ -119,8 +120,10 @@ def _run(self):
             # such wait avoids a following issue with loading an app:
             # https://github.com/phoenix-rtos/phoenix-rtos-project/issues/545
             time.sleep(0.5)
-        except (TimeoutError, Exception) as exc:
-            raise PhoenixdError(str(exc)) from exc
+        except (TimeoutError, AssertionError) as exc:
+            raise PhoenixdError(msg=str(exc)) from exc
+        except Exception as exc:
+            raise PhoenixdError(msg=str(exc), output=traceback.format_exc()) from exc
```
nalajcie, line 126: hmm, it looks strange that the traceback is under `OUTPUT`, imho would be better to format the `PhoenixdError` exception directly inside the the exception implementation, sample code below (not tested):
```python
class ProcessError(HarnessError):
    name: str = "PROCESS"
    # [...]

    def __str__(self) -> str:
        # TODO format github actions output
        err = [bold(f"{self.name} ERROR: ") + self.msg]

        if self.output is not None:
            err.extend([bold("OUTPUT:"), self.output])

        err.extend(self._format_additional_info())

        # ADDED CODE:
        # add traceback if it's rather an implementation error instead of a logic one
        if not isinstance(self.__cause__, TimeoutError):
            err.append(bold("TRACEBACK:"))
            err.extend(traceback.format_exception(self))

        err.append("")
        return "\n".join(err)
```

damianloew: I've looked through all ProcessErrors used in test runner and in most of the cases traceback seems to be unnecessary (the line in test runner is then shown), I'd like to print it only for internal exceptions, where we don't know the exact line. That's why I've proposed adding it only when `Exception` is raised and labeled as `TRACEBACK`:

Screenshot from 2023-11-02 13-42-59

What do you think?

trunner/tools/phoenix.py
```diff
@@ -119,8 +125,11 @@ def _run(self):
             # such wait avoids a following issue with loading an app:
             # https://github.com/phoenix-rtos/phoenix-rtos-project/issues/545
             time.sleep(0.5)
-        except (TimeoutError, Exception) as exc:
+        except (TimeoutError, ValueError) as exc:
             raise PhoenixdError(str(exc)) from exc
+        except Exception as exc:
+            # add traceback if it's an internal exception
+            raise PhoenixdError(str(exc), traceback=traceback.format_exc()) from exc
```
nalajcie, line 132: replacing exception traceback is kindof misleading (no PhoenixdErorr exception is raised at the line provided in traceback). My proposed solution provided real traceback (2 tracebacks because of `from` usage)

PR #273, opened 2023-10-17
trunner: armv7a9: change way that plo is loaded to board

maska989:
## Description
Changing way that plo is loaded to board using onboard jtag chip.

## Motivation and Context
JIRA: CI-364

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## How Has This Been Tested?

## Checklist:

## Special treatment

- [X] This PR needs additional PRs to work 
- https://github.com/phoenix-rtos/phoenix-rtos-build/pull/166

### DIFF
--- a/trunner/target/armv7a9.py
+++ b/trunner/target/armv7a9.py
@@ -1,3 +1,4 @@
+from abc import abstractmethod
 import subprocess
 import time
 from typing import Callable, Optional
@@ -18,7 +19,7 @@ from trunner.harness import (
 )
 from trunner.harness import TerminalHarness
 from trunner.host import Host
-from trunner.tools import JLinkGdbServer, Phoenixd
+from trunner.tools import Phoenixd, OpenocdGdbServer
 from trunner.types import TestResult, TestOptions
 from .base import TargetBase, find_port
 
@@ -41,13 +42,13 @@ class ARMv7A9TargetRebooter(Rebooter):
         self.host.set_flash_mode(not flash)
 
 
-class ZynqGdbPloLoader(TerminalHarness, PloInterface):
+class ZynqZedboardGdbPloLoader(TerminalHarness, PloInterface):
     def __init__(self, dut: Dut, script: str, cwd: Optional[str] = None):
         TerminalHarness.__init__(self)
         PloInterface.__init__(self, dut)
         self.script = script
         self.cwd = cwd
-        self.gdbserver = JLinkGdbServer("Zynq 7020")
+        self.gdbserver = OpenocdGdbServer(board="digilent_zedboard")
 
     def __call__(self):
         """Loads plo image to RAM using gdb."""
@@ -86,22 +87,10 @@ class ARMv7A9Target(TargetBase):
     def from_context(cls, ctx: TestContext):
         return cls(ctx.host, ctx.port, ctx.baudrate)
 
+    @abstractmethod
     def flash_dut(self):
-        plo_loader = ZynqGdbPloLoader(
-            dut=self.dut,
-            script=f"{self._project_dir()}/phoenix-rtos-build/scripts/upload-zynq7000-jlink.gdb",
-            cwd=self.boot_dir(),
-        )
-
-        loader = PloImageLoader(
-            dut=self.dut,
-            rebooter=self.rebooter,
-            image=self.image,
-            plo_loader=plo_loader,
-            phoenixd=Phoenixd(directory=self.boot_dir()),
-        )
-
-        loader()
+        # for now there is no common flashing approach for all future armv7a9 targets
+        pass
 
     def build_test(self, test: TestOptions) -> Callable[[TestResult], TestResult]:
         builder = HarnessBuilder()
@@ -141,3 +130,20 @@ class Zynq7000ZedboardTarget(ARMv7A9Target):
             port = find_port("04b4:0008")
 
         super().__init__(host, port, baudrate)
+
+    def flash_dut(self):
+        plo_loader = ZynqZedboardGdbPloLoader(
+            dut=self.dut,
+            script=f"{self._project_dir()}/phoenix-rtos-build/scripts/upload-zynq7000-smt2.gdb",
+            cwd=self.boot_dir(),
+        )
+
+        loader = PloImageLoader(
+            dut=self.dut,
+            rebooter=self.rebooter,
+            image=self.image,
+            plo_loader=plo_loader,
+            phoenixd=Phoenixd(directory=self.boot_dir()),
+        )
+
+        loader()
--- a/trunner/target/armv7m4.py
+++ b/trunner/target/armv7m4.py
@@ -45,7 +45,10 @@ class STM32L4x6OpenocdGdbServerHarness(IntermediateHarness):
         self.harness = harness
 
     def __call__(self, result: TestResult) -> TestResult:
-        with OpenocdGdbServer(interface="stlink", target="stm32l4x").run():
+        # Set of config parameters used in Openocd to flash up stm32l4a6
+        openocd_args = ["-c", "reset_config srst_only srst_nogate connect_assert_srst", "-c", "init;reset"]
+
+        with OpenocdGdbServer(interface="stlink", target="stm32l4x", extra_args=openocd_args).run():
             self.harness(result)
 
         return self.next_harness(result)
--- a/trunner/tools/gdb.py
+++ b/trunner/tools/gdb.py
@@ -2,7 +2,7 @@ import io
 import signal
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Optional, Union
+from typing import Optional, Union, List
 
 import pexpect
 
@@ -94,35 +94,42 @@ class OpenocdError(ProcessError):
 class OpenocdGdbServer:
     """Handler for OpenocdGdbServer process"""
 
-    def __init__(self, interface: str, target: str, extra_args: Optional[list[str]] = None):
+    def __init__(
+        self,
+        interface: Optional[str] = None,
+        target: Optional[str] = None,
+        board: Optional[str] = None,
+        extra_args: Optional[List[str]] = None,
+    ):
         self.proc = None
         self.target = target
         self.interface = interface
+        self.board = board
         self.output = ""
         self.logfile = io.StringIO()
         self.extra_args = extra_args
+
+        if board:
+            if target or interface:
+                raise OpenocdError("Target or Interface arguments provided when board is specified")
+        else:
+            if not target or not interface:
+                raise OpenocdError("Target or Interface arguments missing")
 
     @contextmanager
     @add_output_to_exception(OpenocdError)
     def run(self):
         try:
             # Use pexpect.spawn to run a process as PTY, so it will flush on a new line
-            args = [
-                "-f",
-                f"interface/{self.interface}.cfg",
-                "-f",
-                f"target/{self.target}.cfg",
-            ]
+            if self.board:
+                args = ["-f", f"board/{self.board}.cfg"]
+            else:
+                args = ["-f", f"interface/{self.interface}.cfg", "-f", f"target/{self.target}.cfg"]
 
             if self.extra_args:
                 args.extend(self.extra_args)
 
-            self.proc = pexpect.spawn(
-                "openocd",
-                args,
-                encoding="ascii",
-                logfile=self.logfile,
-            )
+            self.proc = pexpect.spawn("openocd", args, encoding="ascii", logfile=self.logfile)
 
             try:
                 self.proc.expect_exact("Info : Listening on port 3333 for gdb connections")
--- a/trunner/tools/gdb.py
+++ b/trunner/tools/gdb.py
@@ -2,7 +2,7 @@ import io
 import signal
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Union
+from typing import Optional, Union
 
 import pexpect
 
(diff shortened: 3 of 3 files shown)


trunner/target/armv7a9.py
```diff
@@ -41,13 +41,13 @@ def _set_flash_mode(self, flash):
         self.host.set_flash_mode(not flash)
 
 
-class ZynqGdbPloLoader(TerminalHarness, PloInterface):
+class ZynqGdbServer(TerminalHarness, PloInterface):
```
nalajcie, line 44: IMHO don't change the class name, it still can only load plo.

trunner/target/armv7a9.py
```diff
@@ -56,7 +56,7 @@ def __call__(self):
             try:
                 subprocess.run(
                     ["gdb-multiarch", "plo-gdb.elf", "-x", self.script, "-batch"],
-                    encoding="ascii",
+                    encoding="charmap",
```
nalajcie, line 59: why change of encoding? Seems like trying to hide some other errors...  (as charmap-based encoding will accept any byte, will just fail at a later time - when trying to decode into UTF-8).

Also - I'm not sure it's valid without explicitly stating some charmap (like `cp1252`) - do we have non-standard locale somewhere resulting in national chars being printed?

trunner/tools/gdb.py
```diff
@@ -106,21 +106,34 @@ def __init__(self, interface: str, target: str):
...
-                    f"target/{self.target}.cfg",
-                    "-c",
-                    "reset_config srst_only srst_nogate connect_assert_srst",
-                    "-c",
-                    "init;reset",
-                ],
-                encoding="ascii",
-                logfile=self.logfile,
-            )
+            if self.target == "zynq_7000":
```
nalajcie, line 109: IMHO we should not make special cases based on target here - use generic options instead, eg. extra_params or per-feature flag, eg.

possible approach with extra_args - to be defined on per-target basis in particular target implementation

```python
    def __init__(self, interface: str, target: str, extra_args: Optional[List[str]] = None):
        self.target = target
        self.interface = interface
        self.extra_args = extra_args
        # (...)

    def run(self):
        args = [
            "-f",
            f"interface/{self.interface}.cfg",
            "-f",
            f"target/{self.target}.cfg",
        ]

        if self.extra_args:
            args.extend(self.extra_args)
       
        # (...)

```

trunner/tools/gdb.py
```diff
@@ -106,21 +106,34 @@ def __init__(self, interface: str, target: str):
...
+                self.proc = pexpect.spawn(
+                    "openocd",
+                    [
+                        "-f",
+                        f"interface/{self.interface}.cfg",
+                        "-f",
+                        f"target/{self.target}.cfg",
+                    ],
+                    encoding="charmap",
```
nalajcie, line 118: fishy encoding

nalajcie (changes requested):

badochov:
in order to test it do i need to just connect all 3 usb ports on zedboard to pc and run campaign or some additional setup is needed?

maska989:
> in order to test it do i need to just connect all 3 usb ports on zedboard to pc and run campaign or some additional setup is needed?

To test this change you will need to connect UART(taged as J14) and Jtag(taged as PROG) USB's and setup signal on mio (from 2 to 6) jumpers to gnd.
In armv7a9.py you will need to comment out lines 40 and 41.
after that simply run test runner. You will see something similar to this:
image

trunner/target/armv7a9.py
damianloew: > why change of encoding? Seems like trying to hide some other errors... (as charmap-based encoding will accept any byte, will just fail at a later time - when trying to decode into UTF-8).
> 
> Also - I'm not sure it's valid without explicitly stating some charmap (like `cp1252`) - do we have non-standard locale somewhere resulting in national chars being printed?

Yes, that's not perfect, but I'd opt for leaving `charmap`, or rather `cp1252` to provide the full output - that unfortunately contains non-utf8 char. From what I see it's only used in case of an errror and then it's appended to an exception (this specific char is transformed, but that's not problematic in my opinion):
Screenshot from 2023-10-18 11-40-00

What is the exact case, where it may cause some problems?

Removing `.decode("ascii")` from `raise FlashError(msg=str(e), output=e.stdout.decode("ascii") if e.stdout else None) from e` would also be needed.

trunner/target/armv7a9.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, dut: Dut, script: str, cwd: Optional[str] = None):
         PloInterface.__init__(self, dut)
         self.script = script
         self.cwd = cwd
-        self.gdbserver = JLinkGdbServer("Zynq 7020")
+        self.gdbserver = OpenocdGdbServer("ftdi/digilent_jtag_smt2", "zynq_7000")
```
damianloew, line 50: just more readable

```suggestion
        self.gdbserver = OpenocdGdbServer(interface="ftdi/digilent_jtag_smt2", target= "zynq_7000")
```

trunner/tools/gdb.py
```diff
@@ -94,30 +94,32 @@ class OpenocdError(ProcessError):
...
     """Handler for OpenocdGdbServer process"""
 
-    def __init__(self, interface: str, target: str):
+    def __init__(self, interface: str, target: str, extra_args: Optional[list[str]] = None):
         self.proc = None
         self.target = target
         self.interface = interface
         self.output = ""
         self.logfile = io.StringIO()
+        self.extra_args = extra_args
```
damianloew, line 103: I am not sure `if self.extra_args` it's working, but maybe we can extend arguments even with empty list (then if won't be needed)

```suggestion
        self.extra_args = []
```

trunner/tools/gdb.py
```diff
@@ -94,30 +94,32 @@ class OpenocdError(ProcessError):
...
             # Use pexpect.spawn to run a process as PTY, so it will flush on a new line
+            args = [
+                "-f",
+                f"interface/{self.interface}.cfg",
+                "-f",
+                f"target/{self.target}.cfg",
+            ]
+
+            if self.extra_args:
+                args.extend(self.extra_args)
```
damianloew, line 118:
```suggestion
            args.extend(self.extra_args)
```

trunner/tools/gdb.py
```diff
@@ -94,30 +94,32 @@ class OpenocdError(ProcessError):
 class OpenocdGdbServer:
     """Handler for OpenocdGdbServer process"""
 
-    def __init__(self, interface: str, target: str):
+    def __init__(self, interface: str, target: str, extra_args: Optional[list[str]] = None):
```
damianloew, line 97:
```suggestion
    def __init__(self, interface: str, target: str):
```

trunner/tools/gdb.py
```diff
@@ -2,7 +2,7 @@
 import signal
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Union
+from typing import Optional, Union
```
damianloew, line 5:
```suggestion
from typing import Union
```

trunner/target/armv7m4.py
```diff
@@ -48,7 +48,10 @@ def __init__(self, harness: Callable[[TestResult], TestResult]):
         self.harness = harness
```
damianloew, line 49:
```suggestion
        # Set of additional config parameters used in Openocd to flash up stm32l4a6
        self.extra_args = ["-c", "reset_config srst_only srst_nogate connect_assert_srst", "-c", "init;reset"]
```

trunner/target/armv7m4.py
```diff
@@ -48,7 +48,10 @@ def __init__(self, harness: Callable[[TestResult], TestResult]):
...
 
     def __call__(self, result: TestResult) -> TestResult:
-        with OpenocdGdbServer(interface="stlink", target="stm32l4x").run():
+        # Set of config parameters used in Openocd to flash up stm32l4a6
+        openocd_args = ["-c", "reset_config srst_only srst_nogate connect_assert_srst", "-c", "init;reset"]
+
+        with OpenocdGdbServer(interface="stlink", target="stm32l4x", extra_args=openocd_args).run():
```
damianloew, line 54:
```suggestion
        with OpenocdGdbServer(interface="stlink", target="stm32l4x").run():
```

damianloew (changes requested):
Thanks for the research and suggesting new approach. Some potential simplifications left.

trunner/target/armv7m4.py
nalajcie: whaaat?  This is different `self`, please provide the extra args correctly (via a constructor to different class as it's now)

trunner/tools/gdb.py
nalajcie: It's working, `None` and `[]` (empty list) evaluates to False in python

trunner/target/armv7a9.py
nalajcie: Soooo, this hides some underlying, not diagnosed problem without any comment.

IMHO we should try to pin-point (and hopefully fix) underlying issue instead of hiding it like that.

- does it happen also in non-batch (interactive) mode?
- which command results in erroneous char being printed?
- is there a global .gdbinit which may change it?
- is it related to reset not being configured?
- does gdb version change the output (eg. using `arm-none-eabi-gdb` from GCC toolchain?)

damianloew: Yeah, for sure the comment was missing. Thanks for suggestions, you are right, it's worth to check. @maska989 could you follow them?

trunner/tools/gdb.py
damianloew: There was a mistake in my comment, now it's corrected.

trunner/target/armv7m4.py
damianloew: Ooh right, I confused `STM32L4x6OpenocdGdbServerHarness` with `STM32L4x6OpenocdGdbServer`, I will remove wrong suggestions.

trunner/tools/gdb.py
damianloew, line 97: Then `if self.extra_args:` shouldn't be needed.

```suggestion
    def __init__(self, interface: str, target: str, extra_args: Optional[List[str]] = []):
```

trunner/tools/gdb.py
damianloew, line 5:
```suggestion
from typing import Optional, Union, List
```

trunner/tools/gdb.py
nalajcie: no.
https://www.codiga.io/blog/python-avoid-empty-array-parameter/

damianloew: That's right - I forgot about this issue, so we can leave this as it was.

trunner/target/armv7a9.py
maska989: Problem solved via setting up reset_config and tweeks into upload-zynq7000-smt2.gdb.

nalajcie: great to hear that :)

trunner/tools/gdb.py
```diff
@@ -94,33 +94,35 @@ class OpenocdError(ProcessError):
...
+        self,
+        interface: Optional[str] = None,
+        target: Optional[str] = None,
+        board: Optional[str] = None,
+        extra_args: Optional[List[str]] = None,
+    ):
         self.proc = None
         self.target = target
         self.interface = interface
+        self.board = board
```
nalajcie, line 107: right now board  and target/inferface are mutually exclusive (and changed to optional in typing), we might want to add checks for that, eg:

```python
if board:
    if target or interface:
        raise OpenocdError("target or interface arguments provided when board is specified")
else:
   if not target or not interface:
        raise OpenocdError("target or interface arguments missing")
```

probably adding docstring regarding this new semantics would also be helpful

nalajcie (commented):
the code looks mych better now, one more thing to suggest is to enforce / describe new semantics for instantiation `OpenocdGdbServer` to avoid problems in the future. Otherwise LGTM (if it works ;))

maska989:
> the code looks mych better now, one more thing to suggest is to enforce / describe new semantics for instantiation `OpenocdGdbServer` to avoid problems in the future. Otherwise LGTM (if it works ;))

garuda

Thanks to @badochov i tried new version of upload-zynq7000-smt2.gdb, and it works like a charm on his environment.

nalajcie (approved):

trunner/target/armv7a9.py
```diff
@@ -89,7 +89,7 @@ def from_context(cls, ctx: TestContext):
     def flash_dut(self):
         plo_loader = ZynqGdbPloLoader(
             dut=self.dut,
-            script=f"{self._project_dir()}/phoenix-rtos-build/scripts/upload-zynq7000.gdb",
+            script=f"{self._project_dir()}/phoenix-rtos-build/scripts/upload-zynq7000-smt2.gdb",
```
badochov, line 92: I feel that with this change `class ARMv7A9Target` loses it's previous meaning, as now it only gives support for ARMv7A9 targets that have smt2 controller on board, eg I doubt this class could be reused to test zturn target. Said that I think that class structure needs rethinking

trunner/target/armv7a9.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, dut: Dut, script: str, cwd: Optional[str] = None):
         PloInterface.__init__(self, dut)
         self.script = script
         self.cwd = cwd
-        self.gdbserver = JLinkGdbServer("Zynq 7020")
+        self.gdbserver = OpenocdGdbServer(board="digilent_zedboard")
```
badochov, line 50: This change further limited targets supported by this class to zedboard only thus I feel it should be renamed to `ZynqZedboardGdbPloLoader`

damianloew: I think we could rename it to avoid misunderstandings.

trunner/target/armv7a9.py
damianloew: You are right, it can may be misleading. That's why I'd suggest leaving there:

```
    @abstract_method
    def flash_dut(self):
        # for now there is no common flashing approach for all future armv7a9 targets

```

and move the valid flash_dut() to Zynq7000ZedboardTarget.

@badochov what do you think?
@maska989 could you check this solution?

badochov: Seems like a good solution to me

badochov: I'll approve PR once this change is made

maska989: @badochov done

damianloew (commented):
LGTM, could you please rebase changes and re-run CI on this target ~10 times before merge to make sure that there are no extra issues?

badochov (approved):
LGTM

damianloew (approved):

PR #276, opened 2023-10-19
fix touch tests

damianloew:
## Description

 - psh: assert devs mtimes right after touch in touch tests
 - psh: icrease timeout for ls on bin dir in touch-rootfs test

## Motivation and Context

1.
    The mtime of some devices (for example /dev/console) may change.
    The previous approach asserted modification times of the the whole directory after touching all the files, which took some time.
    Within this period of time files could have been accessed by another process.
    
2.
    On zynq7000-qemu target the execution time of ls on this directory right after system startup is increased.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/psh/test-touch-rootfs.py
+++ b/psh/test-touch-rootfs.py
@@ -12,7 +12,7 @@
 # %LICENSE%
 
 import psh.tools.psh as psh
-from psh.tools.common import assert_mtime
+from psh.tools.common import assert_dir_mtimes
 
 
 def assert_bin(p):
@@ -29,7 +29,7 @@ def assert_bin(p):
         msg = f"Prompt hasn't been seen after touching the following file: {file_path}"
         psh.assert_cmd(p, f'touch {file_path}', result='success', msg=msg)
 
-    assert_mtime(p, dates, '/bin')
+    assert_dir_mtimes(p, dates, '/bin')
 
 
 @psh.run
--- a/psh/test-touch-rootfs.py
+++ b/psh/test-touch-rootfs.py
@@ -18,7 +18,9 @@ from psh.tools.common import assert_dir_mtimes
 def assert_bin(p):
     ''' Asserts that all files in /bin are able to be touched'''
     dates = {}
-    for file_info in psh.ls(p, 'bin'):
+    # Due to the long execution time on zynq-qemu target (especially during CI) after reboot timeout is increased
+    # https://github.com/phoenix-rtos/phoenix-rtos-project/issues/893
+    for file_info in psh.ls(p, 'bin', timeout=40):
         # If file has multiple links, dates of all linked files will change when one is touched
         if file_info.n_links > 1:
             continue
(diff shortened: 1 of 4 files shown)


psh/test-touch.py
```diff
@@ -29,21 +29,27 @@ def assert_executable(p):
...
+    dev_files = psh.ls(p, '/dev')
 
-    assert_mtime(p, dates, dir='/dev')
+    # /dev may include directories - skip them
+    dirs = [file for file in dev_files if file.is_dir]
+    for file in dirs:
+        dev_files.remove(file)
+
+    # check only 5 devices to save time
+    for dev in dev_files[:5]:
```
mateusz-bloch, line 47: It could be written in a single line, but I'm not sure if that wouldn't reduce readability
it wouldn't make a difference in performance either.
```suggestion
    # /dev may include directories - skip them
    # Get only 5 devices to save time
    dev_files = [file for file in psh.ls(p, '/dev') if not file.is_dir][:5]

    for dev in dev_files:
```

psh/tools/common.py
```diff
@@ -57,16 +69,27 @@ def assert_mtime(p, datetimes: Dict[str, datetime], dir=''):
...
-        # Check if delta is smaller than resolution and allow for file to be up to 1 second
-        # newer (this can happen because `date` is called first, then `touch`)
-        assert delta < date_resolution and delta >= timedelta(seconds=-1), "".join((
-            f'The modification time for {filename} is not equal to the target one! ',
-            f'file datetime: {date} target datetime: {target_datetime}'))
+        # # Check if delta is smaller than resolution and allow for file to be up to 2 seconds
+        # # newer (this can happen because `date` is called first, then `touch`)
+        # assert delta < date_resolution and delta >= timedelta(seconds=-2), "".join((
+        #     f'The modification time for {dir}/{filename} is not equal to the target one! ',
+        #     f'file datetime: {date} target datetime: {target_datetime}'))
```
mateusz-bloch, line 83: Comments to remove

psh/test-touch-rootfs.py
```diff
@@ -12,13 +12,14 @@
...
-from psh.tools.common import assert_mtime
+from psh.tools.common import assert_dir_mtimes
 
 
 def assert_bin(p):
     ''' Asserts that all files in /bin are able to be touched'''
     dates = {}
-    for file_info in psh.ls(p, 'bin'):
+    # Due to the long execution time on zynq-qemu target (especially during CI) after reboot timeout is increased
+    for file_info in psh.ls(p, 'bin', timeout=40):
```
mateusz-bloch, line 22: don't like increasing the timeout. Does it exceed quite often, or is it not a significant issue?

mateusz-bloch (commented):
Overall it seems okay

psh/test-touch-rootfs.py
damianloew: quite often, from my test the times were:

- 30s
- 31s
- 27s
- 34s

psh/test-touch.py
damianloew: Thanks for this idea, I've applied it partly

mateusz-bloch (approved):

PR #279, opened 2023-10-19
trunner: armv7a9: increase timeout for erase

maska989:
## Description
Increase timeout on Jffs2 according to documentation

## Motivation and Context
On our test targets, we can find Spansion s25fl256s1 32 MB nor flash(2.0)

And to erase we use 64/4 kb sector erase, according to documentation on that Spansion chip:
image

After quick calculation (((32768÷64)×0,65)÷60) where we divide space on flash by 64 kb and multiply that by time (max time in secs) and divide again to gain minutes we get ~ 5.54 min to erase whole flash memory (without time passing from sector to sector)

We increased timeout something above half this value to be certain that Jffs2 will have enough time, but also if time will exceed, we will know about some changes in this case.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/target/armv7a9.py
+++ b/trunner/target/armv7a9.py
@@ -132,7 +132,7 @@ class Zynq7000ZedboardTarget(ARMv7A9Target):
             block_size=0x10000,
             cleanmarker_size=0x10,
         ),
-        timeout=140,
+        timeout=210,
     )
     name = "armv7a9-zynq7000-zedboard"
 


damianloew (commented):
Can we close this PR as the better solution is proposed in https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/283 ?
if so @maska989 please do it.

PR #282, opened 2023-10-20
cpp: test-hello: add hello world test

maska989:
JIRA: CI-355

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/cpp/test.yaml
+++ b/cpp/test.yaml
@@ -1,0 +1,9 @@
+test:
+
+  tests:
+    - name: hello-cpp
+      harness: test-hello-cpp.py
+      execute: /usr/bin/hellocpp
+        # targets excluded because of insufficient space in RAM
+      targets:
+        exclude: [armv7m4-stm32l4x6-nucleo]
--- a/cpp/test-hello-cpp.py
+++ b/cpp/test-hello-cpp.py
@@ -1,0 +1,2 @@
+def harness(p):
+    p.expect_exact("Hello World++!")


nalajcie:
can't we just use `_user/hellocpp/`?

damianloew:
> can't we just use `_user/hellocpp/`?

Yeah, I'd also prefer this approach.

cpp/test.py
```diff
@@ -0,0 +1,2 @@
+def harness(p):
+    assert "Hello world++!" in p.readline()
```
mateusz-bloch, line 2:
```suggestion
    assert "Hello World++!" in p.readline()
```

alternatively (personally prefer the fail message in this case, if occure):

```suggestion
    p.expect_exact('Hello World++!')
```

Makefile.host
```diff
@@ -1,3 +1,4 @@
 DEFAULT_COMPONENTS = $(filter test_meterfs_%,$(ALL_COMPONENTS))
 DEFAULT_COMPONENTS += $(SAMPLE_TESTS)
 DEFAULT_COMPONENTS += $(LIBC_UNIT_TESTS)
+DEFAULT_COMPONENTS += $(CPP_TESTS)
```
mateusz-bloch, line 4: The binary is not currently building for host-generic-pc
```suggestion
```

cpp/test.yaml
```diff
@@ -0,0 +1,7 @@
+test:
+    tests:
+        - name: main
+          harness: test.py
+          execute: ../usr/bin/hellocpp
+          target:
```
mateusz-bloch, line 6:
```suggestion
          targets:
```

cpp/test.yaml
```diff
@@ -0,0 +1,7 @@
+test:
+    tests:
+        - name: main
+          harness: test.py
+          execute: ../usr/bin/hellocpp
```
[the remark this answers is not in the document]
damianloew: After some thinking, I'd consider changing the way how this test name is parsed - we could for example check whether there is `\` at the beginning of string and then assume it's full path and don't add `/bin` prefix. This way we could place there just `/usr/bin`, which is much more readable.

cpp/test.yaml
```diff
@@ -0,0 +1,8 @@
+test:
+    tests:
+        - name: main
+          harness: test.py
```
mateusz-bloch, line 4: Maybe, I would  consider changing the name of the test from `main` to something else, the same with `test.py`

cpp/test.yaml
```diff
@@ -0,0 +1,9 @@
+test:
+
+  tests:
+    - name: test-echo-cpp
+      harness: test-echo.py
```
damianloew, line 5: Why test-echo? You are not sending anything, I'd leave the same structure as in psh tests:
```suggestion
    - name: hello-cpp
      harness: test-hello-cpp.py
```

damianloew (commented):
Moreover, this `test-hello:` is unnecessary in the commit message, the following message would be fine: `cpp: add hello world test`

cpp/test.yaml
maska989: My mistake, I made changes other than I intended.

damianloew (approved):

PR #283, opened 2023-10-25
trunner: assert jffs2 via output instead of time

maska989:
## Description
Changed approach how are determined problems with erase with jffs2.

## Motivation and Context
JIRA: CI-375

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a9-zynq7000-zedboard.

## Checklist:

## Special treatment

### DIFF
--- a/trunner/harness/plo.py
+++ b/trunner/harness/plo.py
@@ -146,14 +146,24 @@ class PloInterface:
             e.cmd = cmd
             raise e
 
-    def jffs2(self, device: str, erase: bool, cleanmarkers: PloJffs2CleanmarkerSpec, timeout: Optional[int] = None):
+    def jffs2(self, device: str, erase: bool, cleanmarkers: PloJffs2CleanmarkerSpec, block_timeout: int):
         """Performs jffs2 command."""
+
+        block_count = cleanmarkers.number_of_blocks
 
         cmd = f"jffs2 -d {device} -c {cleanmarkers}"
         if erase:
             cmd += " -e"
 
-        self.cmd(cmd, timeout)
+        self.send_cmd(cmd)
+
+        for i in range(0, block_count):
+            try:
+                self.dut.expect_exact(f"jffs2: block {i}/{block_count}", timeout=block_timeout)
+            except pexpect.TIMEOUT as e:
+                raise PloError("Wrong jffs2 command output!", cmd=cmd, output=self.dut.before) from e
+
+        self._assert_prompt()
 
     def app(
         self, device: str, file: str, imap: str, dmap: str, exec: bool = False
@@ -220,7 +230,8 @@ class PloJffsImageProperty(PloImageProperty):
 
     flash_device_id: str
     cleanmarkers_args: PloJffs2CleanmarkerSpec
-    timeout: int
+    # optimal timeout for erasing 1 block using jffs2 command
+    block_timeout: Optional[int] = 1
 
 
 class PloImageLoader(TerminalHarness, PloInterface):
@@ -258,7 +269,7 @@ class PloImageLoader(TerminalHarness, PloInterface):
         self.plo_loader()
 
         if isinstance(self.image, PloJffsImageProperty):
-            self.jffs2(self.image.flash_device_id, True, self.image.cleanmarkers_args, self.image.timeout)
+            self.jffs2(self.image.flash_device_id, True, self.image.cleanmarkers_args, self.image.block_timeout)
 
         with self.phoenixd.run():
             self.copy_file2mem(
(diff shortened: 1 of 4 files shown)


trunner/target/armv7a9.py
```diff
@@ -121,7 +121,6 @@ class Zynq7000ZedboardTarget(ARMv7A9Target):
             block_size=0x10000,
             cleanmarker_size=0x10,
         ),
-        timeout=140,
```
damianloew, line 124: for `IMX6ULLEvkTarget` it seems to be still set to 275s.

trunner/harness/plo.py
```diff
@@ -220,7 +230,7 @@ class PloJffsImageProperty(PloImageProperty):
 
     flash_device_id: str
     cleanmarkers_args: PloJffs2CleanmarkerSpec
-    timeout: int
+    timeout: Optional[int] = 1
```
damianloew, line 233: Maybe:
```suggestion
    # optimal timeout for erasing 1 block using jffs2 command
    timeout: Optional[int] = 1
```

trunner/harness/plo.py
```diff
@@ -146,14 +146,24 @@ def erase(self, device: str, offset: int, size: int, timeout: Optional[int] = No
             e.cmd = cmd
             raise e
 
-    def jffs2(self, device: str, erase: bool, cleanmarkers: PloJffs2CleanmarkerSpec, timeout: Optional[int] = None):
+    def jffs2(self, device: str, erase: bool, cleanmarkers: PloJffs2CleanmarkerSpec, timeout: int):
```
damianloew, line 149:
```suggestion
    def jffs2(self, device: str, erase: bool, cleanmarkers: PloJffs2CleanmarkerSpec, block_timeout: int):
```

trunner/harness/plo.py
damianloew, line 233:
```suggestion
    block_timeout: Optional[int] = 1
```

damianloew (changes requested):
Minor suggestions left

trunner/harness/plo.py
```diff
@@ -146,14 +146,24 @@ def erase(self, device: str, offset: int, size: int, timeout: Optional[int] = No
...
             cmd += " -e"
 
-        self.cmd(cmd, timeout)
+        self.send_cmd(cmd)
+
+        for i in range(0, block_count):
+            try:
+                self.dut.expect(f"(jffs2: block {i}/{block_count})", timeout=timeout) == 0
+            except pexpect.TIMEOUT as e:
+                raise PloError("\nWrong jffs2 command output!", cmd=cmd, output=self.dut.before) from e
```
nalajcie, line 164: the message shouldn't have `\n` at the beginning, if there is a problem with printing the exception please fix it where the exception is being printed

trunner/harness/plo.py
```diff
@@ -146,14 +146,24 @@ def erase(self, device: str, offset: int, size: int, timeout: Optional[int] = No
...
         cmd = f"jffs2 -d {device} -c {cleanmarkers}"
         if erase:
             cmd += " -e"
 
-        self.cmd(cmd, timeout)
+        self.send_cmd(cmd)
+
+        for i in range(0, block_count):
+            try:
+                self.dut.expect(f"(jffs2: block {i}/{block_count})", timeout=timeout) == 0
```
nalajcie, line 162: `== 0` does nothing here (exception won't be raised if different),

also: I think we don't need grouping (`(...)`) in pattern, actually we're not using regexps here so `expect_exact` would be a better fit (as it's faster)

trunner/harness/plo.py
```diff
@@ -146,14 +146,24 @@ def erase(self, device: str, offset: int, size: int, timeout: Optional[int] = No
...
             cmd += " -e"
 
-        self.cmd(cmd, timeout)
+        self.send_cmd(cmd)
+
+        for i in range(0, block_count):
+            try:
+                self.dut.expect_exact(f"jffs2: block {i}/{block_count}", timeout=block_timeout)
+            except pexpect.TIMEOUT as e:
+                raise PloError("Wrong jffs2 command output!", cmd=cmd, output=self.dut.before) from e
```
damianloew, line 164: I see that you removed this newline from there, but it would be good to add it before "ERROR WHILE FLASHING DEVICE" as we discussed during the meeting, so I'd opt for:

```
# the newline is needed to avoid printing exception in the same line as plo prompt
print(bold("\nERROR WHILE FLASHING THE DEVICE"))
```

damianloew (changes requested):

damianloew (changes requested):
I'd create the separate commit for applying black corrections in `test_runner.py`, so for example:

`trunner: apply black corrections in test_runner.py`

damianloew (approved):
LGTM

PR #284, opened 2023-10-26
mbedtls|busybox: fix measuring test duration (trunner API change approach)

nalajcie:
## Description
- mbedtls: fix measuring test duration
  The subtest result was incorrectly "finished" (added to result) one line after
  the result (so actually computed the time of the next test).

  Use simplified API to add optional test message to the correct subtest

- trunner: make subresult API more robust
  - remove add_subresult_obj() as it's misleading (doesn't add by reference,
  just by value
  - make add_subresult() return the subresult object which can be updated at a
  later time

## Motivation and Context
phoenix-rtos/phoenix-rtos-project#877

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/busybox/busybox.py
+++ b/busybox/busybox.py
@@ -17,6 +17,7 @@ EXAMPLE_INPUT = """
 echo -ne '' >input
 echo -ne '' | ./unknown 2>&1
 SKIPPED: busybox --help
+UNTESTED: cp-parents
 PASS: busybox as unknown name
 ---Testcase 'date-@-works' starting---
 FAIL: date-@-works
@@ -36,7 +37,7 @@ def harness(dut: Dut, ctx: TestContext, result: TestResult):
     test_status = Status.OK
 
     START = r"---Testcase '(?P<name>.+?)' starting---\r+\n"
-    RESULT = r"(?P<status>PASS|SKIPPED|FAIL): (?P<name>.+?)\r+\n"
+    RESULT = r"(?P<status>PASS|SKIPPED|UNTESTED|FAIL): (?P<name>.+?)\r+\n"
     FINAL = r"\*\*\*\*(The Busybox Test Suite completed|A single test of the Busybox Test Suite completed)\*\*\*\*\r+\n"
     MESSAGE = r"(?P<line>.*?)\r+\n"
 
--- a/busybox/busybox.py
+++ b/busybox/busybox.py
@@ -9,7 +9,25 @@
 
 from trunner.ctx import TestContext
 from trunner.dut import Dut
-from trunner.types import TestResult, TestSubResult, Status
+from trunner.types import TestResult, Status
+
+
+EXAMPLE_INPUT = """
+---Testcase 'busybox as unknown name' starting---
+echo -ne '' >input
+echo -ne '' | ./unknown 2>&1
+SKIPPED: busybox --help
+PASS: busybox as unknown name
+---Testcase 'date-@-works' starting---
+FAIL: date-@-works
++ TZ=EET-2EEST,M3.5.0/3,M10.5.0/4 busybox date -d @1288486799
++ test xSun Oct 31 03:59:59 EEST 2010 = xSun Oct 31 00:59:59  2010
+---Testcase 'date-R-works' starting---
+PASS: date-R-works
+---Testcase 'bunzip2: doesnt exist' starting---
+- t_actual differ: char 13, line 1
+FAIL: bunzip2: doesnt exist
+"""
 
 
 def harness(dut: Dut, ctx: TestContext, result: TestResult):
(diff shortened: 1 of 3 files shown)


nalajcie:
switching to draft until `busybox` harness changes would be implemented

busybox/busybox.py
```diff
@@ -26,21 +26,23 @@ def harness(dut: Dut, ctx: TestContext, result: TestResult):
...
+            status = Status.from_str(parsed["status"])
+
+            if status == Status.FAIL:
                 test_status = Status.FAIL
-                subresult.msg = "\n".join(msg)
-                msg = []
 
-            result.add_subresult_obj(subresult)
-            subresult = None
+            subresult = result.add_subresult(subname=parsed["name"], status=status, msg="".join(msg))
```
nalajcie, line 43: Don't be afraid of newlines (I think we can use them instead of `" "` above also)

```suggestion
            subresult = result.add_subresult(subname=parsed["name"], status=status, msg="\n".join(msg))
```

busybox/busybox.py
```diff
@@ -26,21 +26,23 @@ def harness(dut: Dut, ctx: TestContext, result: TestResult):
         parsed = dut.match.groupdict()
 
         if idx == 2:
-            msg.append("\t\t" + parsed["line"])
-            continue
+            line_msg = parsed["line"] + " "
```
nalajcie, line 29: I don't really know how the full output from the succesful/failed tescases looks like. Maybe we can put sample output as a mulitiline variable / comment after imports (to be used for pytest harness unit test in the near future hopefully).

From your code I suspect the output might be:
```
FAIL: test-name-1
why-failed-line-1
why-failed-line-2
msg-for-test-2
msg-for-test-2
PASS: test-name-2
```

If so - we can't assign message lines to the testcases in any sensible way (your code would append all 4 lines to `test-name-1`, but that's ok if the input is really like that) - but we should mention in comment that the messages might be invalid in that case.

mateusz-bloch: After checking, the output can appear in several ways. It might be produced during the test. In the event of a failure, lines can be printed both during the test when something fails and after the test. However, when printing strictly informational messages, they are preceded by `======================` to distinguish them.

e.g.:
image
image

nalajcie: I've checked the busybox testsuite and it seems seems to be true only for newstyle tests (oldstyle tests don't print `===` nor messages before the test result - see `mkdir` for example). I don't see how to distinguish them in harness, maybe we should patch busybox testsuite framework instead?

busybox/busybox.py
```diff
@@ -26,21 +26,29 @@ def harness(dut: Dut, ctx: TestContext, result: TestResult):
         parsed = dut.match.groupdict()
 
         if idx == 2:
-            msg.append("\t\t" + parsed["line"])
-            continue
+            # checks whether the message belongs to the next test
+            if "---Testcase" in parsed["line"]:
+                subresult = None
+                continue
```
damianloew, line 32: please create a separate regex pattern for the testcase initial log and add it to the `expect` method.

busybox/busybox.py
```diff
@@ -9,7 +9,7 @@
 
 from trunner.ctx import TestContext
 from trunner.dut import Dut
-from trunner.types import TestResult, TestSubResult, Status
+from trunner.types import TestResult, Status
 
 
 def harness(dut: Dut, ctx: TestContext, result: TestResult):
```
damianloew, line 15: Please provide a few characteristic cases from both mbedtls and busybox output in those harness files to make it more understandable and prepare some data for tests (testing the harnesses) as @nalajcie suggested before.

```suggestion
# exemplary output from the busybox testsuite:
# ...

def harness(dut: Dut, ctx: TestContext, result: TestResult):
```

damianloew (changes requested):

mbedtls/mbedtls.py
```diff
@@ -2,12 +2,19 @@
 
 from trunner.ctx import TestContext
 from trunner.dut import Dut
-from trunner.types import TestResult, Status, TestSubResult
+from trunner.types import TestResult, Status
+
+# Example of parsed outputs
+#
+#   test-name-1 ................................... FAIL
```
nalajcie, line 9: `FAILED ` (see regex below)

also change to multiline var

busybox/busybox.py
```diff
@@ -9,39 +9,63 @@
...
+#   SKIPPED: test-name-3
+#
 
 
 def harness(dut: Dut, ctx: TestContext, result: TestResult):
     subresult = None
     msg = []
     test_status = Status.OK
 
+    START = r"---Testcase (?P<name>.*?)starting---\r+\n"
```
nalajcie, line 36: should probably be 
```suggestion
    START = r"---Testcase '(?P<name>.+?)' starting---\r+\n"
```

use regexp101.com to test the regexes against sample output provided above

busybox/busybox.py
```diff
@@ -9,39 +9,63 @@
 
 from trunner.ctx import TestContext
 from trunner.dut import Dut
-from trunner.types import TestResult, TestSubResult, Status
+from trunner.types import TestResult, Status
+
+# Example of parsed outputs:
```
nalajcie, line 14: use multiline string , right now I don't know if the input has spaces in front or not, eg:
```python
EXAMPLE_INPUT = """
test-msg-line-1
test-msg-line-2
FAIL: test-name-1
fail-msg-line-1
fail-msg-line-2
---Testcase test-name-2 starting---
test-msg-line-1
test-msg-line-2
PASS: test-name-2
---Testcase test-name-3 starting---
SKIPPED: test-name-3
"""
```
this would also make it easier to write unit tests (use  variable as an input).

Also - test names should be in `'`?

busybox/busybox.py
```diff
@@ -9,39 +9,64 @@
...
+from trunner.types import TestResult, Status
+
+
+EXAMPLE_INPUT = """
+---Testcase 'busybox as unknown name' starting---
+echo -ne '' >input
+echo -ne '' | ./unknown 2>&1
+PASS: busybox as unknown name
+---Testcase 'date-@-works' starting---
+FAIL: date-@-works
```
damianloew, line 21: I think that it would be better to place it in a comment.

busybox/busybox.py
```diff
@@ -9,39 +9,64 @@
...
++ TZ=EET-2EEST,M3.5.0/3,M10.5.0/4 busybox date -d @1288486799
++ test xSun Oct 31 03:59:59 EEST 2010 = xSun Oct 31 00:59:59  2010
+---Testcase 'date-R-works' starting---
+PASS: date-R-works
```
damianloew, line 25: Please add also a SKIPPED case and it would be fine for me.

damianloew (changes requested):
Thanks for the latest improvements. Just 2 things more

busybox/busybox.py
mateusz-bloch: I based on the previous comment:
image

nalajcie: Please keep it as a multiline variable (reasoning provided previously).

damianloew: Oh, right. Ok, I was worried about some potential issues (flake and so on) when we don't use those variables in fact.

damianloew (commented):
@mateusz-bloch please add JIRA ID to the first two commits and I'd like to merge it :)

damianloew (dismissed):
LGTM

damianloew (approved):
One case from busybox test suite was not handled, thanks for adding it @mateusz-bloch. Looks good to me

PR #290, opened 2023-11-07
trunner: yaml add absolute path support in execute parameter

mateusz-bloch:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -157,16 +157,25 @@ class ConfigParser:
             raise ParserError("execute/run attribute cannot be empty")
 
         if execute_binary:
-            prefix = self.ctx.target.exec_dir() + "/" if self.ctx.target.rootfs else "sysexec "
-            parsed_cmd = shlex.split(prefix + cmd[0]) + cmd[1:]
-            binary = cmd[0]
+            # it may be both an absolute path to a binary or just binary name
+            binary_path = Path(cmd[0])
+
+            binary = binary_path.name
+            path = binary_path.parent if binary_path.is_absolute() else self.ctx.target.exec_dir()
+            prefix = str(path) + "/" if self.ctx.target.rootfs else "sysexec "
+            parsed_cmd = shlex.split(prefix + binary) + cmd[1:]
         else:
             binary = None
             parsed_cmd = cmd
+            path = Path()
 
         self.test.shell = ShellOptions(
             binary=binary,
             cmd=parsed_cmd,
+            # when launching host tools like phoenixd absolute path
+            # from phoenix has to be combined with absolute path for host.
+            # Two absolute paths can't be combined, so phoenix path is transformed to a relative one
+            path=path.relative_to("/"),
         )
 
     def _parse_reboot(self, config: dict) -> None:
--- a/trunner/target/armv7m4.py
+++ b/trunner/target/armv7m4.py
@@ -162,7 +162,7 @@ class STM32L4x6Target(TargetBase):
                 app_loader = STM32L4x6PloAppLoader(
                     dut=self.dut,
                     apps=test.bootloader.apps,
-                    gdb=GdbInteractive(port=3333, cwd=self.bin_dir()),
+                    gdb=GdbInteractive(port=3333, cwd=self.root_dir() / test.shell.path),
                 )
 
             builder.add(PloHarness(self.dut, app_loader=app_loader))
--- a/trunner/target/armv7m7.py
+++ b/trunner/target/armv7m7.py
@@ -69,7 +69,9 @@ class ARMv7M7Target(TargetBase):
 
             if test.bootloader.apps:
                 app_loader = PloPhoenixdAppLoader(
-                    dut=self.dut, apps=test.bootloader.apps, phoenixd=Phoenixd(directory=self.bin_dir())
+                    dut=self.dut,
+                    apps=test.bootloader.apps,
+                    phoenixd=Phoenixd(directory=self.root_dir() / test.shell.path),
                 )
 
             builder.add(PloHarness(self.dut, app_loader=app_loader))
--- a/trunner/target/base.py
+++ b/trunner/target/base.py
@@ -78,13 +78,13 @@ class TargetBase(ABC):
 
         raise FileNotFoundError("phoenix-rtos-tests directory is missing")
 
-    def exec_dir(self) -> str:
+    def exec_dir(self) -> Path:
         """Returns a directory where test binaries are located in rootfs."""
-        return "/bin"
+        return Path("/bin")
 
-    def bin_dir(self) -> str:
-        """Returns a directory with binaries that will be loaded by bootloader."""
-        return f"{self.project_dir}/_fs/{self.name}/root/bin"
+    def root_dir(self) -> Path:
+        """Returns a host path to root directory for the specified target"""
+        return Path(f"{self.project_dir}/_fs/{self.name}/root")
 
     def boot_dir(self) -> str:
         """Returns a directory with system and bootloader images."""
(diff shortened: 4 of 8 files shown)


trunner/types.py
```diff
@@ -372,6 +372,7 @@ class BootloaderOptions:
 class ShellOptions:
     binary: Optional[str] = None
     cmd: Optional[List[str]] = None
+    path: Optional[str] = "/bin/"
```
damianloew, line 375: If it's a path we can use `Path` type and then use `/` instead of `+`.

trunner/target/armv7m7.py
```diff
@@ -69,7 +69,9 @@ def build_test(self, test: TestOptions) -> Callable[[TestResult], TestResult]:
 
             if test.bootloader.apps:
                 app_loader = PloPhoenixdAppLoader(
-                    dut=self.dut, apps=test.bootloader.apps, phoenixd=Phoenixd(directory=self.bin_dir())
+                    dut=self.dut,
+                    apps=test.bootloader.apps,
+                    phoenixd=Phoenixd(directory=self.root_dir() + test.shell.path),
```
damianloew, line 74: Assuming that `Path` type will be used.
```suggestion
                    phoenixd=Phoenixd(directory=self.root_dir() / test.shell.path),
```

trunner/target/armv7m4.py
```diff
@@ -162,7 +162,7 @@ def build_test(self, test: TestOptions):
                 app_loader = STM32L4x6PloAppLoader(
                     dut=self.dut,
                     apps=test.bootloader.apps,
-                    gdb=GdbInteractive(port=3333, cwd=self.bin_dir()),
+                    gdb=GdbInteractive(port=3333, cwd=self.root_dir() + test.shell.path),
```
damianloew, line 165:
```suggestion
                    gdb=GdbInteractive(port=3333, cwd=self.root_dir() / test.shell.path),
```

trunner/target/base.py
```diff
@@ -82,6 +82,10 @@ def exec_dir(self) -> str:
         """Returns a directory where test binaries are located in rootfs."""
         return "/bin"
 
+    def root_dir(self) -> str:
+        """Returns a root directory."""
+        return f"{self.project_dir}/_fs/{self.name}/root"
```
damianloew, line 87: Then also other dirs in TargetBase should be changed.

```suggestion
    def root_dir(self) -> Path:
        """Returns a root directory."""
        return Path(f"{self.project_dir}/_fs/{self.name}/root")
```

damianloew (commented):
Let's take a moment to check the approach with `Path` type.

mateusz-bloch:
> Let's take a moment to check the approach with `Path` type.

Based on what I've tested, I'm not sure if it makes sense to switch to paths, as we'll still need the string at the final stage.

For example, in the case of imx106x and imx117x, where we use phoenixd to upload binaries to the hardware, we do ithis using pexpect:
image
where self.dir is directory from:
image

trunner/config.py
```diff
@@ -157,16 +157,25 @@ def _parse_shell_command(self, config: dict):
...
+            path = path_cmd.parent if path_cmd.is_absolute() else self.ctx.target.exec_dir()
         else:
             binary = None
             parsed_cmd = cmd
+            path = Path()
 
         self.test.shell = ShellOptions(
             binary=binary,
             cmd=parsed_cmd,
+            path=path.relative_to("/"),
```
damianloew, line 178: As I suggested before it wuld be good to place a comment why it's done

```suggestion
            # when launching host tools like phoenixd absolute path from phoenix has to be combined with absolute path for host.
            # Two absolute paths can't be combined, so phoenix path is transformed to a relative one
            path=path.relative_to("/"),
```

trunner/config.py
```diff
@@ -157,16 +157,25 @@ def _parse_shell_command(self, config: dict):
             raise ParserError("execute/run attribute cannot be empty")
```
damianloew, line 158: I think it may be simplified a little bit, not fully tested, but there is my proposal:

```suggestion
        if execute_binary:
            # binary_path may be both an absolute path to a binary or just binary name (then the path will be default)
            binary_path = Path(cmd[0])

            binary = binary_path.name
            path = binary_path.parent if binary_path.is_absolute() else self.ctx.target.exec_dir()
            prefix = str(path) + "/" if self.ctx.target.rootfs else "sysexec "
            parsed_cmd = shlex.split(prefix + binary) + cmd[1:]
        else:
            binary = None
            parsed_cmd = cmd
            path = Path()

        self.test.shell = ShellOptions(
            binary=binary,
            cmd=parsed_cmd,
            path=path.relative_to("/"),
        )
```

trunner/target/host.py
```diff
@@ -46,7 +47,7 @@ def from_context(cls, _: TestContext):
         return cls()
 
     def exec_dir(self) -> str:
-        return self.bin_dir()
+        return self.bin_dir() / Path("bin")
```
damianloew, line 50: Won't it be fine?

```suggestion
        return self.bin_dir() / "bin"
```

damianloew (changes requested):

damianloew (commented):
Moreover, it would be good to support this utility also on host:
Screenshot from 2023-11-15 13-20-58

From my quick research we would have to do sth like that:
Screenshot from 2023-11-15 13-48-29

(Maybe there is more convenient way)

trunner/target/base.py
```diff
@@ -78,13 +78,13 @@ def _project_dir(self) -> str:
...
+    def exec_dir(self) -> Path:
         """Returns a directory where test binaries are located in rootfs."""
-        return "/bin"
+        return Path("/bin")
 
-    def bin_dir(self) -> str:
-        """Returns a directory with binaries that will be loaded by bootloader."""
-        return f"{self.project_dir}/_fs/{self.name}/root/bin"
+    def root_dir(self) -> Path:
+        """Returns a root directory"""
```
damianloew, line 86:
```suggestion
        """Returns a host path to root directory for the specified target"""
```

trunner/target/host.py
```diff
@@ -63,6 +60,7 @@ def fail(result: TestResult):
             builder.add(fail)
             return builder.get_harness()
 
+        test.shell.cmd[0] = f"{self.root_dir()}{test.shell.cmd[0]}"
```
damianloew, line 63: Some comment may be helpful there
```suggestion
        # on host we execute binaries directly from the root directory located in phoenix-rtos-project/_fs
        test.shell.cmd[0] = f"{self.root_dir()}{test.shell.cmd[0]}"
```

trunner/tools/gdb.py
```diff
@@ -78,7 +78,8 @@ def run(self):
                 "gdb-multiarch",
                 encoding="ascii",
                 timeout=8,
-                cwd=self.cwd,
+                # pexpect required cwd to be provided as a string
```
damianloew, line 81: I know that it was my suggestion, but maybe it's not as much needed - too many comments decreases readability.
```suggestion
```

trunner/tools/phoenix.py
```diff
@@ -122,7 +122,8 @@ def _run(self):
         # Use pexpect.spawn to run a process as PTY, so it will flush on a new line
         self.proc = pexpect.spawn(
             "phoenixd",
-            ["-p", self.port, "-s", self.dir],
+            # pexpect required self.dir to be provided as a string
```
damianloew, line 125:
```suggestion
```

damianloew (changes requested):
Only a few small remarks

damianloew (approved):
LGTM

PR #295, opened 2023-11-14
K arkadiusz/stdio indicator test

KArkadiusz:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/stdio/stdio_indicator.c
+++ b/libc/stdio/stdio_indicator.c
@@ -1,0 +1,628 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libc-tests
+ *
+ * Testing POSIX file operations.
+ *
+ * Copyright 2023-2024 Phoenix Systems
+ * Author: Arkadiusz Kozlowski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+#include <limits.h>
+
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <sys/stat.h>
+#include <sys/socket.h>
+#include <fcntl.h>
+#include "../testdata.h"
+
+#include <unity_fixture.h>
+
+
+#define STR_EQUAL(a, b) !strcmp(a, b)
+#define ALL_ASCII       256
+
+#define CHECK_MATCH(_res, _first, ...) \
+	do { \
+		const char *_strs[] = { __VA_ARGS__ }; \
+		int _match = 0; \
+		for (size_t _idx = 0; _idx < sizeof(_strs) / sizeof(_strs[0]); ++_idx) { \
+			if (strcmp(_first, _strs[_idx]) == 0) { \
+				_match = 1; \
+				break; \
+			} \
+		} \
+		_res = _match; \
+	} while (0);
+
+const static char *modes[] = { "r", "r+", "w", "w+", "a", "a+", "rb", "rb+", "wb", "wb+", "ab", "ab+" };
+const int num_modes = sizeof(modes) / sizeof(modes[0]);
+
+const char *tellMode(const char *mode)
+{
+	static char msg[80];
+	sprintf(msg, "Tested file mode: %s Error: %s", mode, strerror(errno));
+	return msg;
+}
+
+
+TEST_GROUP(stdio_feof);
+
+TEST_SETUP(stdio_feof)
+{
+}
+
+
+TEST_TEAR_DOWN(stdio_feof)
+{
+}
+
+
+TEST(stdio_feof, not_empty_all_modes)
+{
+	const char *filename = "test_stdio_feof_filled";
+	char *data = testdata_createCharStr(ALL_ASCII);
+	int tmp;
+	for (int i = 0; i < num_modes; ++i) {
+		FILE *f;
+		f = fopen(filename, "w+");
+
+		TEST_ASSERT_NOT_NULL_MESSAGE(f, tellMode(modes[i]));
+		TEST_ASSERT_NOT_NULL_MESSAGE(data, tellMode(modes[i]));
+
+		TEST_ASSERT_NOT_EQUAL_INT_MESSAGE(EOF, fputs(data, f), tellMode(modes[i]));
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, fclose(f), tellMode(modes[i]));
+
+		f = fopen(filename, modes[i]);
+
+		TEST_ASSERT_NOT_NULL_MESSAGE(f, tellMode(modes[i]));
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, feof(f), tellMode(modes[i]));
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, fseek(f, 0, SEEK_END), tellMode(modes[i]));
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, feof(f), tellMode(modes[i]));
+		TEST_ASSERT_EQUAL_INT(-1, fgetc(f));
+		CHECK_MATCH(tmp, modes[i], "w", "a", "wb", "ab");
+		TEST_ASSERT_EQUAL_MESSAGE(tmp ? 0 : 1, feof(f), tellMode(modes[i]));
+
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, fseek(f, 0, SEEK_SET), tellMode(modes[i]));
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, feof(f), tellMode(modes[i]));
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, fclose(f), tellMode(modes[i]));
+	}
+	free(data);
+	TEST_ASSERT_EQUAL_INT(0, remove(filename));
+}
+
+TEST(stdio_feof, empty_all_modes)
+{
+	FILE *f;
+	const char *filename = "test_stdio_feof_empty";
+	f = fopen(filename, "w+");
+	int tmp;
+
+	TEST_ASSERT_NOT_NULL(f);
+	TEST_ASSERT_EQUAL_INT(0, fclose(f));
+
+	for (int i = 0; i < num_modes; ++i) {
+		f = fopen(filename, modes[i]);
+		TEST_ASSERT_NOT_NULL_MESSAGE(f, tellMode(modes[i]));
+
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, feof(f), tellMode(modes[i]));
+
+		TEST_ASSERT_EQUAL(-1, fgetc(f));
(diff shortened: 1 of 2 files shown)


libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,599 @@
...
+	close(pipefd[1]);
+
+
+	const char *fifo_path = "myfifo";
+	mkfifo(fifo_path, 0666);
+
+	pid_t pid = fork();
+	TEST_ASSERT_NOT_EQUAL(-1, pid);
+
+	if (pid) {
```
damianloew, line 323: Please resolve warnings:
Screenshot from 2023-11-20 17-53-14

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,599 @@
...
+	char **buffer = NULL;
+	int n;
+	int *counter = &n;
+	*counter = 2;
+
+	FILE *f = fopen("decently_long_test", "w+");
+	fputs("RelativelyLongTextThatWillBeBiggerThanNInGetdelim", f);
+	rewind(f);
+
+	TEST_ASSERT_EQUAL(-1, getdelim(buffer, counter, '\n', f));
```
damianloew, line 452: Please resolve the following warnings, you can always make sure that no warnings were generated in `files changed` tab.
Screenshot from 2023-11-20 18-13-09

damianloew (changes requested):
Haven't checked ftell in detail yet.

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,599 @@
...
+		fclose(f);
+
+		f = fopen(filename, modes[i]);
+
+		TEST_ASSERT_EQUAL(0, feof(f));
+
+
+		fseek(f, 0, SEEK_SET);
+		TEST_ASSERT_EQUAL(0, feof(f));
+		fseek(f, 1, SEEK_END);
```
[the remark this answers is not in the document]
KArkadiusz: Yes, i wanted to make sure that ftell() returns correct result even after exceeding the file.

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,599 @@
...
+TEST(stdio_feof, open_and_closed_not_empty_file_every_mode)
+{
+	char *modes[] = { "r", "r+", "w", "w+", "a", "a+", "rb", "rb+", "wb", "wb+", "ab", "ab+" };
+
+	for (int i = 0; i < 12; ++i) {
+		char filename[32];
+		sprintf(filename, "file_FEOF_filled_%s", modes[i]);
+
+		FILE *f;
+		f = fopen(filename, "w+");
```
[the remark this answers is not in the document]
KArkadiusz: Probably yes, but making files with distinct names might  help localize the potential flaw or error a bit faster, you'd know instantly when there would be a few not deleted files with similar names, and you'd be able to know at which loop run did it happen, with single file you'd lose that information.

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,599 @@
...
+		fseek(f, 1, SEEK_END);
+		fgetc(f);
+
+		if (EQ("w", modes[i]) ||
+			EQ("wb", modes[i]) ||
+			EQ("a", modes[i]) ||
+			EQ("ab", modes[i]))
+			TEST_ASSERT_EQUAL(0, feof(f));
+		else
+			TEST_ASSERT_NOT_EQUAL_MESSAGE(0, feof(f), modes[i]);
```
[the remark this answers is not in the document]
KArkadiusz: I temporarily changed the assertion requirement, but it looks something like this. I should've added message in upper assertion too.
image

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,599 @@
...
+		fputs(data, f);
+		free(data);
+		fclose(f);
+
+		f = fopen(filename, modes[i]);
+
+		TEST_ASSERT_EQUAL(0, feof(f));
+
+
+		fseek(f, 0, SEEK_SET);
```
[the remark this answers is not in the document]
KArkadiusz: Kind of yes, I wanted to make sure that feof() will return right value regardless of way we reached the end of the file, no matter if it's with getc(), fseek() and so on.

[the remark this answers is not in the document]
damianloew: Yeah, but when you are opening a new file the file offset should be always 0, so you don't need to set it once again.

libc/stdio/stdio_indicator.c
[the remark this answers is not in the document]
damianloew: So let's change this message to sth like "Tested file mode: r", you can use sprintf, applies also to the other places, so you can use a common string.

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,630 @@
...
+	TEST_ASSERT_EQUAL_INT(0, fclose(f));
+	TEST_ASSERT_EQUAL(-1, ftell(f));
+	TEST_ASSERT_EQUAL_INT(EBADF, errno);
+	TEST_ASSERT_EQUAL_INT(0, remove(filename));
+}
+
+
+TEST(stdio_ftell, wrong_stream_type)
+{
+	TEST_IGNORE_MESSAGE("Issue 923");
```
damianloew: Let's use the following format as in the rest of our tests.

```suggestion
	TEST_IGNORE_MESSAGE("#923 issue");
```

damianloew: Moreover the issue is only about sockets, so you can enable checks for pipe and fifo.

damianloew (changes requested):

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,628 @@
...
+	TEST_ASSERT_EQUAL(0, remove(fifo_path));
+}
+
+
+TEST_GROUP_RUNNER(stdio_ftell)
+{
+	RUN_TEST_CASE(stdio_ftell, correct_position_not_empty);
+	RUN_TEST_CASE(stdio_ftell, correct_position_empty);
+	RUN_TEST_CASE(stdio_ftell, bad_file_descriptor);
+	RUN_TEST_CASE(stdio_ftell, wrong_stream_type);
```
damianloew, line 389: I think that `EOVERFLOW` errno may be also checked.

damianloew: We can prepare a few edge cases, so when file content equals long max, long max - 1 and long max + 1, same for off_t.

damianloew: Maybe some separate testcase for checking this reuirement? (or checking it everywhere else where ftell call is successful)
`The ftell() function shall not change the setting of errno if successful.`

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,630 @@
...
+
+	TEST_ASSERT_EQUAL_INT(0, fclose(pipe_stream));
+	TEST_ASSERT_NOT_EQUAL_INT(EOF, close(pipefd[0]));
+	TEST_ASSERT_NOT_EQUAL_INT(EOF, close(pipefd[1]));
+
+
+	const char *fifo_path = "myfifo";
+	TEST_ASSERT_NOT_EQUAL_INT(-1, mkfifo(fifo_path, 0666));
+
+	pid_t pid = fork();
```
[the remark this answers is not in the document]
KArkadiusz: At first I had some problems and thought that it was necessary to have two processes using the fifo in order for it to not freeze the test, but that was not the cause of the problem. It's solved now.

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,630 @@
...
+
+TEST(stdio_ftell, bad_file_descriptor)
+{
+	errno = 0;
+	char filename[] = "test_bfd.txt";
+	FILE *f = fopen(filename, "w+");
+	TEST_ASSERT_EQUAL_INT(0, fclose(f));
+	TEST_ASSERT_EQUAL(-1, ftell(f));
+	TEST_ASSERT_EQUAL_INT(EBADF, errno);
+	TEST_ASSERT_EQUAL_INT(0, remove(filename));
```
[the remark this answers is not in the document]
KArkadiusz: I was thinking about such case, but it raises exception 14, both on host and qemu, so I decided it's not worth implementing.
image

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,628 @@
...
+	size_t len = 0;
+
+	/* read using getdelim from write-only file */
+	FILE *f = fopen("test_wronly", "a");
+	TEST_ASSERT_NOT_NULL(f);
+
+	{
+		rewind(f);
+		TEST_ASSERT_EQUAL_INT(-1, getdelim(&line, &len, 'x', f));
+		TEST_ASSERT_EQUAL_INT(EBADF, errno);
```
mateusz-bloch, line 609: errno not reset earlier

mateusz-bloch (changes requested):

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,628 @@
...
+		}
+
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, fseek(f, 0, SEEK_SET), tellMode(modes[i]));
+
+		{
+			int i = 1;
+			int errnoBefore;
+			char c;
+			while ((c = fgetc(f)) != -1) {
+				TEST_ASSERT_EQUAL_CHAR(c, data[i - 1]);
```
mateusz-bloch, line 221: What about this error, I see that it occurs on all targets except ia32, host-generic-pc. 
image

mateusz-bloch (changes requested):

libc/stdio/stdio_indicator.c
KArkadiusz: Seems not related to ftell nor fgetc function, removed the assertion that caused it.

mateusz-bloch: Even if it's not directly related to fread or fgetc, it would be worth to check whether it's something desired or not. If not, an issue should be reported.

As I see there are still problems with the last character of the extended ascii ?
image

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,627 @@
...
+			// free(data);
+			continue;
+		}
+
+		TEST_ASSERT_EQUAL_INT_MESSAGE(0, fseek(f, 0, SEEK_SET), tellMode(modes[i]));
+
+		{
+			int i = 1;
+			int errnoBefore;
+			char c;
```
mateusz-bloch, line 219:
```suggestion
			int c;
```

libc/stdio/stdio_indicator.c
```diff
@@ -0,0 +1,628 @@
...
+			while ((c = fgetc(f)) != -1) {
```
mateusz-bloch, line 221: On our ARM architectures targets char is from 0 to 255,, maybe use signed char or int for `c`

PR #298, opened 2023-11-17
K arkadiusz/perror test

KArkadiusz:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/string/string_errsig.c
+++ b/libc/string/string_errsig.c
@@ -24,6 +24,7 @@
 #include <signal.h>
 #include <limits.h>
 #include <unistd.h>
+#include <fcntl.h>
 #include <unity_fixture.h>
 
 #include "testdata.h"
@@ -53,10 +54,30 @@ const unsigned int signal_codes_len = sizeof(signal_codes) / sizeof(signal_codes
 
 char *perrorToFile(char *msg)
 {
-	FILE *file = fopen("error.log", "w+");
+	remove("error.txt");
+
+	errno = 0;
+
+	creat("error.txt", 0777);
+
+
+	printf("%d, %s\n\n", errno, strerror(errno));
+
+	FILE *file = fopen("error.txt", "w+");
+
+	printf("%d, %s\n\n", errno, strerror(errno));
+
+
+	TEST_MESSAGE(strerror(errno));
+
+	chmod("error.txt", 0777);
+
+	TEST_MESSAGE(strerror(errno));
 
 	TEST_ASSERT_NOT_NULL(file);
 	TEST_ASSERT_NOT_EQUAL(-1, dup2(fileno(file), 2));
+
+	TEST_MESSAGE(strerror(errno));
 
 	perror(msg);
 
@@ -73,7 +94,6 @@ char *perrorToFile(char *msg)
 	buffer[fileSize] = '\0';
 
 	fclose(file);
-	remove("error.log");
 
 	return buffer;
 }
@@ -213,6 +233,7 @@ TEST_GROUP(string_perror);
 
 TEST_SETUP(string_perror)
 {
+	errno = 0;
 }
 
 
@@ -236,9 +257,9 @@ TEST(string_perror, perror_basic)
 
 TEST(string_perror, perror_empty_message)
 {
-	errno = 0;
 	char *res;
 
+	errno = 0;
 	TEST_ASSERT_EQUAL_STRING("Success\n", res = perrorToFile(""));
 	free(res);
 }
--- a/libc/string/string_errsig.c
+++ b/libc/string/string_errsig.c
@@ -54,31 +54,17 @@ const unsigned int signal_codes_len = sizeof(signal_codes) / sizeof(signal_codes
 
 char *perrorToFile(char *msg)
 {
-	remove("error.txt");
-
-	errno = 0;
-
-	creat("error.txt", 0777);
-
-
-	printf("%d, %s\n\n", errno, strerror(errno));
+	int errnoBefore = errno;
 
 	FILE *file = fopen("error.txt", "w+");
 
-	printf("%d, %s\n\n", errno, strerror(errno));
 
-
-	TEST_MESSAGE(strerror(errno));
-
-	chmod("error.txt", 0777);
-
-	TEST_MESSAGE(strerror(errno));
+	// chmod("error.txt", 0777);
 
 	TEST_ASSERT_NOT_NULL(file);
 	TEST_ASSERT_NOT_EQUAL(-1, dup2(fileno(file), 2));
 
-	TEST_MESSAGE(strerror(errno));
-
+	errno = errnoBefore;
 	perror(msg);
 
 	fseek(file, 0, SEEK_END);
@@ -96,6 +82,7 @@ char *perrorToFile(char *msg)
 	fclose(file);
 
 	return buffer;
+	remove("error.txt");
 }
 
 
@@ -246,11 +233,13 @@ TEST(string_perror, perror_basic)
 
 	errno = 0;
 	char msg[] = "Some error message";
-
 	char *res;
+	char exp[100];
+	sprintf(exp, "Some error message: %s\n", strerror(errno));
+
 	res = perrorToFile(msg);
 
-	TEST_ASSERT_EQUAL_STRING("Some error message: Success\n", res);
+	TEST_ASSERT_EQUAL_STRING(exp, res);
 	free(res);
 }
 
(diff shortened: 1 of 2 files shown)


libc/string/string_errsig.c
```diff
@@ -48,6 +51,41 @@ static const unsigned int error_codes_len = sizeof(error_codes) / sizeof(error_c
...
+	errno = errnoBefore;
+	perror(msg);
+
+	fseek(file, 0, SEEK_END);
+	long fileSize = ftell(file);
+	rewind(file);
+
+	rewind(file);
+
+	char *buffer = (char *)malloc(fileSize + 1);
```
mateusz-bloch, line 76: There's no `#include <stdlib.h>`
image

libc/string/string_errsig.c
```diff
@@ -48,6 +51,41 @@ static const unsigned int error_codes_len = sizeof(error_codes) / sizeof(error_c
...
+	TEST_ASSERT_NOT_EQUAL(-1, dup2(fileno(file), 2));
+
+	errno = errnoBefore;
+	perror(msg);
+
+	fseek(file, 0, SEEK_END);
+	long fileSize = ftell(file);
+	rewind(file);
+
+	rewind(file);
```
mateusz-bloch, line 74: Why rewind for the second time?

libc/string/string_errsig.c
```diff
@@ -48,6 +51,41 @@ static const unsigned int error_codes_len = sizeof(error_codes) / sizeof(error_c
...
+	char *buffer = (char *)malloc(fileSize + 1);
+
+
+	fread(buffer, 1, fileSize, file);
+	buffer[fileSize] = '\0';
+
+	fclose(file);
+
+	return buffer;
+	remove("error.txt");
```
mateusz-bloch, line 85:
```suggestion
	fclose(file);
	remove("error.txt");
	return buffer;
```

libc/string/string_errsig.c
```diff
@@ -48,6 +51,41 @@ static const unsigned int error_codes_len = sizeof(error_codes) / sizeof(error_c
...
+{
+	int errnoBefore = errno;
+
+	FILE *file = fopen("error.txt", "w+");
+
+
+	// chmod("error.txt", 0777);
+
+	TEST_ASSERT_NOT_NULL(file);
+	TEST_ASSERT_NOT_EQUAL(-1, dup2(fileno(file), 2));
```
mateusz-bloch, line 65: From what I remember, there should be a variable defining which file descriptor is for stderr. I know this is likely to be constant, but in case of changes, we are sure that nothing will break.

```suggestion
	TEST_ASSERT_NOT_EQUAL(-1, dup2(fileno(file), STDERR_FILENO));
```

libc/string/string_errsig.c
```diff
@@ -175,3 +213,107 @@ TEST_GROUP_RUNNER(string_errsign)
...
+}
+
+TEST(string_perror, perror_basic)
+{
+
+	errno = 0;
+	char msg[] = "Some error message";
+	char *res;
+	char exp[100];
+	sprintf(exp, "Some error message: %s\n", strerror(errno));
```
mateusz-bloch, line 238: "Shouldn't this work the same way?
```suggestion
	char msg[] = "Some error message";
	char *res;
	char exp[100];
	sprintf(expected, "%s: %s\n", msg, strerror(0));
```

libc/string/string_errsig.c
```diff
@@ -175,3 +213,107 @@ TEST_GROUP_RUNNER(string_errsign)
...
+TEST(string_perror, perror_every_ascii)
+{
+	char *msg = testdata_createCharStr(257);
+
+
+	char *res;
+	errno = 8;
+	res = perrorToFile(msg);
+
+	char expected[356];
```
mateusz-bloch, line 269:
```suggestion
	char *msg = testdata_createCharStr(257);
	char expected[356];
	char *res;

	errno = 8;
	res = perrorToFile(msg);
```

libc/string/string_errsig.c
```diff
@@ -175,3 +213,107 @@ TEST_GROUP_RUNNER(string_errsign)
...
+	sprintf(expected, "%s: %s\n", msg, strerror(errno));
+	TEST_ASSERT_EQUAL_STRING(expected, res);
+	free(res);
+	free(msg);
+}
+
+
+TEST(string_perror, perror_huge_argument)
+{
+	char *msg = testdata_hugeStr;
```
mateusz-bloch, line 279: Remember to check for warnings and try to find out why they are occurring
image

mateusz-bloch (changes requested):
This isn't a full review, I just took a quick look

mateusz-bloch (changes requested):
This isn't a full review, I just took a quick look

libc/string/string_errsig.c
```diff
@@ -48,6 +52,35 @@ static const unsigned int error_codes_len = sizeof(error_codes) / sizeof(error_c
...
 
+
+char *perrorToFile(const char *msg)
+{
+	int errnoBefore = errno;
+
+	FILE *file = fopen("error.txt", "w+");
+
+	TEST_ASSERT_NOT_NULL(file);
+	TEST_ASSERT_NOT_EQUAL(-1, dup2(fileno(file), STDERR_FILENO));
```
mateusz-bloch, line 63: I'm thinking if in case dup2 fails, then the file wouldn't be closed, and could this result in the rest of the test cases failing even though only one might, Am i correct ?

----

@damianloew 
After giving it some thought, I'm thinking how to approach this because at the moment, in most tests, if something fails in the middle of the test, we then don't do free(), fclose(), etc. 
Is it worth switching to something like:
```C
if (OPERATION(RESOURCE) == NULL) {
    free(RESOURCE);
    TEST_FAIL_MESSAGE("OPERATION FAILED");
}
```
Or to stay with our current approach

libc/string/string_errsig.c
```diff
@@ -175,3 +208,120 @@ TEST_GROUP_RUNNER(string_errsign)
...
 	RUN_TEST_CASE(string_errsign, strsignal_real_time);
 }
+
+
+TEST_GROUP(string_perror);
+
+
+TEST_SETUP(string_perror)
+{
+	errno = 0;
```
mateusz-bloch, line 218: There's no need, as we set the appropriate errno in each test case anyway.

libc/string/string_errsig.c
```diff
@@ -175,3 +208,120 @@ TEST_GROUP_RUNNER(string_errsign)
...
+	char *new_msg;
+
+	errno = 0;
+
+	for (int i = 0; i < 150; i++) {
+		errno = i;
+		old_msg = perrorToFile("Some msg");
+		errno++;
+		new_msg = perrorToFile("Some msg");
+		errno--;
```
mateusz-bloch, line 296: Not needed ?
```suggestion
```

libc/string/string_errsig.c
```diff
@@ -175,3 +208,120 @@ TEST_GROUP_RUNNER(string_errsign)
...
+
+TEST_GROUP_RUNNER(string_perror)
+{
+	RUN_TEST_CASE(string_perror, perror_basic);
+	RUN_TEST_CASE(string_perror, perror_empty_message);
+	RUN_TEST_CASE(string_perror, perror_every_ascii);
+	RUN_TEST_CASE(string_perror, perror_huge_argument);
+	RUN_TEST_CASE(string_perror, perror_every_errno);
+	RUN_TEST_CASE(string_perror, perror_null_argument);
+}
```
mateusz-bloch, line 327: I'm considering adding more test cases, possibly for:

- extreme values of errno
- what happens when an invalid character is placed in msg
- a huge message (we could use memset)

libc/string/string_errsig.c
```diff
@@ -48,6 +52,35 @@ static const unsigned int error_codes_len = sizeof(error_codes) / sizeof(error_c
...
+	errno = errnoBefore;
+	perror(msg);
+
+	fseek(file, 0, SEEK_END);
+	long fileSize = ftell(file);
+	rewind(file);
+
+	char *buffer = (char *)malloc(fileSize + 1);
+
+	fread(buffer, 1, fileSize, file);
```
mateusz-bloch, line 74: image

mateusz-bloch (changes requested):

libc/string/string_errsig.c
KArkadiusz: Huge message has already been checked in `perror_huge_argument`,
Invalid character was checked in `perror_every_ascii`,
and extreme values of errno were checked in `perror_every_errno`.
I might modify conditions if I misunderstood something.

libc/string/string_errsig.c
KArkadiusz: Too late, already done.

mateusz-bloch (approved):

PR #301, opened 2023-11-29
trunner: plo: force to pass block erase timeout

maska989:
## Types of changes

## How Has This Been Tested?

- [X] Tested by hand on: armv7a9-zynq7000-zedboard / armv7a7-imx6ull-evk.

## Checklist:

## Special treatment

### DIFF
--- a/trunner/harness/plo.py
+++ b/trunner/harness/plo.py
@@ -230,8 +230,8 @@ class PloJffsImageProperty(PloImageProperty):
 
     flash_device_id: str
     cleanmarkers_args: PloJffs2CleanmarkerSpec
-    # optimal timeout for erasing 1 block using jffs2 command
-    block_timeout: Optional[int] = 1
+    # !!! NOTE: For new flash memories, block_timeout should be set with documentation.
+    block_timeout: int
 
 
 class PloImageLoader(TerminalHarness, PloInterface):
--- a/trunner/harness/plo.py
+++ b/trunner/harness/plo.py
@@ -230,7 +230,7 @@ class PloJffsImageProperty(PloImageProperty):
 
     flash_device_id: str
     cleanmarkers_args: PloJffs2CleanmarkerSpec
-    # !!! NOTE: For new flash memories, block_timeout should be set with documentation.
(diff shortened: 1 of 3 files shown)


trunner/target/armv7a7.py
```diff
@@ -84,6 +84,15 @@ class IMX6ULLEvkTarget(ARMv7A7Target):
...
             cleanmarker_size=0x10,
         ),
+        # Setup of neutral block erase timeout
+        # This approach was led by documentation of MT25QL256 IC Flash.
+        # On system is declared 64kb block size but to erase is used 4kb
+        # subsector erase, in order to create average, is needed to divide
+        # block size by sector erase to achieve 16 erases on block and simply
+        # multiply that by subsector erase time. The time achieved by this
+        # calculation is the maximal possible erase time. To get average,
+        # is needed to divide that value by 2.
```
damianloew, line 94: I suggest a bit shorter comment, but showing why such value has been set:

```suggestion
        # Based on MT25QL256 IC docs 4KB subsector erase time equals 0.4s max.
        # Blocksize set for jffs2 command is 64KB, so 16 sectors have to be erased for 1 block
        # That's why we set (0.4 * 16 / 2) as a timeout. The value is divided by 2 to early detect slowing the flash
```

Please do sth similar for the second target.

trunner/harness/plo.py
```diff
@@ -230,8 +230,8 @@ class PloJffsImageProperty(PloImageProperty):
 
     flash_device_id: str
     cleanmarkers_args: PloJffs2CleanmarkerSpec
-    # optimal timeout for erasing 1 block using jffs2 command
-    block_timeout: Optional[int] = 1
+    # !!! NOTE: For new flash memories, block_timeout should be set with documentation.
```
damianloew, line 233: I think the comment without screaming would be fine :smiley: 
```suggestion
    # For new flash memories, block_timeout should be set with documentation.
```

damianloew (changes requested):

damianloew (approved):

PR #304, opened 2023-12-04
trunner: mute dmesg while truner lasts

maska989:
## Description
Disabling dmesg output until test campaign are done

## Motivation and Context
JIRA: CI-393

## Types of changes

## How Has This Been Tested?

- [X] Tested by hand on: armv7a7-imx6ull-evk.

## Checklist:

## Special treatment

### DIFF
--- a/psh/tools/psh.py
+++ b/psh/tools/psh.py
@@ -1,4 +1,4 @@
-'''psh module with tools for psh related tests'''
+"""psh module with tools for psh related tests"""
 # Phoenix-RTOS
 #
 # phoenix-rtos-tests
@@ -23,30 +23,30 @@ import trunner
 
 from datetime import datetime, timedelta
 
-EOL = r'(?:\r+)\n'
-EOT = '\x04'
-PROMPT = r'(\r*)\x1b\[0J' + r'\(psh\)% '
+EOL = r"(?:\r+)\n"
+EOT = "\x04"
+PROMPT = r"(\r*)\x1b\[0J" + r"\(psh\)% "
 # according to control sequence introducer pattern
-CONTROL_CODE = r'(\x1b\[[\x30-\x3F]*[\x20-\x2F]*[\x40-\x7E])'
-OPTIONAL_CONTROL_CODE = CONTROL_CODE + r'?'
+CONTROL_CODE = r"(\x1b\[[\x30-\x3F]*[\x20-\x2F]*[\x40-\x7E])"
+OPTIONAL_CONTROL_CODE = CONTROL_CODE + r"?"
 
 
-File = namedtuple('File', ['name', 'owner', 'is_dir', 'datetime', 'datetime_resolution', 'n_links'])
+File = namedtuple("File", ["name", "owner", "is_dir", "datetime", "datetime_resolution", "n_links"])
 
 
 def _readable(exp_regex):
-    ''' Gets more readable expected regex with new lines instead of raw string EOL,
-    and prompt without raw esc sequences. '''
-    exp_regex = exp_regex.replace(EOL, '\n')
-    return exp_regex.replace(PROMPT, '(psh)% ')
+    """Gets more readable expected regex with new lines instead of raw string EOL,
+    and prompt without raw esc sequences."""
+    exp_regex = exp_regex.replace(EOL, "\n")
+    return exp_regex.replace(PROMPT, "(psh)% ")
 
 
 def _check_result(pexpect_proc, result):
-    if result == 'dont-check':
+    if result == "dont-check":
         return
-    elif result == 'success':
+    elif result == "success":
         assert_cmd_successed(pexpect_proc)
-    elif result == 'fail':
+    elif result == "fail":
         assert_cmd_failed(pexpect_proc)
     else:
         raise Exception(
@@ -55,15 +55,15 @@ def _check_result(pexpect_proc, result):
         )
 
 
-def assert_cmd(pexpect_proc, cmd, *, expected='', result='success', msg='', is_regex=False, timeout=-1):
-    ''' Sends specified command and asserts that it's displayed correctly
-    with optional expected output and next prompt. Exit status is asserted depending on `result`'''
+def assert_cmd(pexpect_proc, cmd, *, expected="", result="success", msg="", is_regex=False, timeout=-1):
+    """Sends specified command and asserts that it's displayed correctly
+    with optional expected output and next prompt. Exit status is asserted depending on `result`"""
     pexpect_proc.sendline(cmd)
     cmd = re.escape(cmd)
-    exp_regex = ''
+    exp_regex = ""
     if is_regex:
         exp_regex = expected
-    elif expected != '':
+    elif expected != "":
         if not (isinstance(expected, tuple) or isinstance(expected, list)):
             expected = tuple((expected,))
         for line in expected:
@@ -72,20 +72,20 @@ def assert_cmd(pexpect_proc, cmd, *, expected='', result='success', msg='', is_r
 
     exp_regex = cmd + EOL + exp_regex + PROMPT
     exp_readable = _readable(exp_regex)
-    msg = f'Expected output regex was: \n---\n{exp_readable}\n---\n' + msg
+    msg = f"Expected output regex was: \n---\n{exp_readable}\n---\n" + msg
 
     assert pexpect_proc.expect([exp_regex, pexpect.TIMEOUT, pexpect.EOF], timeout=timeout) == 0, msg
 
     _check_result(pexpect_proc, result)
 
 
-def assert_prompt_after_cmd(pexpect_proc, cmd, result='success', msg=None):
-    ''' Sends specified command and asserts that the command and next prompt are displayed correctly.
-    Exit status is asserted depending on the result argument '''
+def assert_prompt_after_cmd(pexpect_proc, cmd, result="success", msg=None):
+    """Sends specified command and asserts that the command and next prompt are displayed correctly.
+    Exit status is asserted depending on the result argument"""
     pexpect_proc.sendline(cmd)
     exp_regex = EOL + PROMPT
     if not msg:
-        msg = f'Prompt not seen after sending the following command: {cmd}'
+        msg = f"Prompt not seen after sending the following command: {cmd}"
     assert pexpect_proc.expect([exp_regex, pexpect.TIMEOUT]) == 0, msg
     output = pexpect_proc.before
 
@@ -94,8 +94,8 @@ def assert_prompt_after_cmd(pexpect_proc, cmd, result='success', msg=None):
     return output
 
 
-def assert_prompt(pexpect_proc, msg='', timeout=-1, catch_timeout=True):
-    patterns = ['(psh)% ']
+def assert_prompt(pexpect_proc, msg="", timeout=-1, catch_timeout=True):
+    patterns = ["(psh)% "]
     if catch_timeout:
         patterns.append(pexpect.TIMEOUT)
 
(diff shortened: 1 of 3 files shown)


trunner/harness/psh.py
```diff
@@ -52,16 +52,31 @@ class ShellHarness(IntermediateHarness):
...
     def __call__(self, result: TestResult) -> TestResult:
         try:
             self.dut.expect_exact(self.prompt, timeout=self.prompt_timeout)
+
+            # Mute dmesg to focus output of system strictly on
+            # tests, logs are collected until re-enable dmesg.
+            # Additional flags are unavailable on stm32l4x6
+            if self.name != "armv7m4-stm32l4x6-nucleo":
+                self.dut.send("dmesg -D\n")
+
```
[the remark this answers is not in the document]
mateusz-bloch: So, since we are sending something through the DUT, maybe it would also be worth checking if it has definitely been sent (?) (expect)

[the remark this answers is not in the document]
nalajcie: IMHO wrapper is too constrained, we probably need to wait for prompt before sending `dmesg -E` - and maybe do some other things in the future (like get the data until next prompt and put it in separate GH log group).

Actually, all harnessess (`IntermediateHarness`) are wrappers, so we could introduce `SuppresKlogHarness` but AFAIK we would still need `prompt` / `prompt_timeout` there - so I'm ok with intergrating it into `ShellHarness`
(still, you may experiment with making it a separate `IntermediateHarness` to see if the code would be clearer or not.

damianloew (commented):
I'd opt for sth like this in https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/8c6d062c2fbd327e2664ec371fcf8bddf1135483/trunner/harness/pyharness.py#L73

```python
        try:
            # TODO maybe we should catch the output from the test, for example based on test configuration
            if self.ctx.suppress_dmesg:
                self.dut.pexpect_proc.sendline("dmesg -D")
                self.dut.pexpect_proc.expect_exact(self.ctx.target.shell_prompt)
                test_result = self.pyharness(*args, **kwargs)
                self.dut.pexpect_proc.sendline("dmesg -E")
                self.dut.pexpect_proc.expect_exact(self.ctx.target.shell_prompt)
            else:
                test_result = self.pyharness(*args, **kwargs)
```

What do you all think? For me it's more readable as we can see disabling and enabling logs in the same place. Maybe some wrapper also could be used there.

trunner/test_runner.py
```diff
@@ -280,6 +280,11 @@ def run_tests(self, tests: Sequence[TestOptions]) -> Sequence[TestResult]:
                 except HarnessError as e:
                     result.fail(str(e))
 
+            # Additional flags for dmesg are unavailable on stm32l4x6
+            if self.ctx.target.name != "armv7m4-stm32l4x6-nucleo":
+                # Re-enabling dmesg output to investigate collected data.
+                self.target.dut.send("dmesg -E\r")
```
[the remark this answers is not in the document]
nalajcie: also: this should be a part of `ShellHarness`

trunner/harness/psh.py
```diff
@@ -52,16 +52,25 @@ class ShellHarness(IntermediateHarness):
...
         self.dut = dut
         self.prompt = prompt
         self.cmd = " ".join(map(shlex.quote, cmd)) if cmd is not None else cmd
         self.prompt_timeout = prompt_timeout
+        self.suppress_dmesg = suppress_dmesg
 
     def __call__(self, result: TestResult) -> TestResult:
         try:
             self.dut.expect_exact(self.prompt, timeout=self.prompt_timeout)
+
```
nalajcie, line 73: remove empty line

trunner/harness/psh.py
```diff
@@ -81,4 +90,14 @@ def __call__(self, result: TestResult) -> TestResult:
                 ) from e
 
         result.set_stage(TestStage.RUN)
-        return self.next_harness(result)
+
+        if self.suppress_dmesg:
+            self.dut.pexpect_proc.sendline("dmesg -D")
+            self.dut.expect_exact(self.prompt)
+            test_result = self.next_harness(result)
```
nalajcie, line 97: IMHO we should call next_harness only once (split the dmesg disable /enable into separate IFs in respective places)

trunner/target/armv7m4.py
```diff
@@ -176,7 +176,7 @@ def build_test(self, test: TestOptions):
             builder.add(STM32L4x6OpenocdGdbServerHarness(setup))
 
         if test.shell is not None:
-            builder.add(ShellHarness(self.dut, self.shell_prompt, test.shell.cmd))
+            builder.add(ShellHarness(self.dut, self.shell_prompt, test.shell.cmd, suppress_dmesg=False))
```
nalajcie, line 179: add comment explaining WHY

trunner/harness/psh.py
```diff
@@ -52,16 +52,25 @@ class ShellHarness(IntermediateHarness):
...
     """
 
-    def __init__(self, dut: Dut, prompt: str, cmd: Optional[List[str]] = None, prompt_timeout: Optional[int] = -1):
+    def __init__(
+        self,
+        dut: Dut,
+        prompt: str,
+        cmd: Optional[List[str]] = None,
+        prompt_timeout: Optional[int] = -1,
+        suppress_dmesg: Optional[bool] = True,
```
nalajcie, line 61: Type is not `Optional`, only `bool` (takes True / False values, not True / False / None). 

The same with `prompt_timeout` I assume (not related to this PR, not sure when it was introduced)

nalajcie (commented):
fix also commit message

trunner/harness/psh.py
```diff
@@ -69,6 +77,11 @@ def __call__(self, result: TestResult) -> TestResult:
                 output=self.dut.before,
             ) from e
 
+        # Suppress klog output to console while test is running to avoid problems with parsing
+        if self.suppress_dmesg:
+            self.dut.pexpect_proc.sendline("dmesg -D")
+            self.dut.expect_exact(self.prompt)
```
nalajcie, line 83: missing `prompt_timeout support` and we should translate `pexpect` exceptions to `ShellError`

Make waiting for prompt a separate function of this class and call it in all places needed

trunner/harness/psh.py
```diff
@@ -81,4 +94,11 @@ def __call__(self, result: TestResult) -> TestResult:
                 ) from e
 
         result.set_stage(TestStage.RUN)
-        return self.next_harness(result)
+        test_result = self.next_harness(result)
+
+        # Re-enabling log output to release collected output
+        if self.suppress_dmesg:
+            self.dut.pexpect_proc.sendline("dmesg -E")
+            self.dut.expect_exact(self.prompt)
```
nalajcie, line 102: missing `prompt_timeout support` and we should translate `pexpect` exceptions to `ShellError` (see above)

trunner/harness/psh.py
```diff
@@ -81,4 +94,11 @@ def __call__(self, result: TestResult) -> TestResult:
                 ) from e
 
         result.set_stage(TestStage.RUN)
-        return self.next_harness(result)
+        test_result = self.next_harness(result)
+
+        # Re-enabling log output to release collected output
```
nalajcie, line 99: use the same tense in all comments (`re-enable`)

trunner/harness/psh.py
```diff
@@ -52,12 +52,20 @@ class ShellHarness(IntermediateHarness):
 
     """
 
-    def __init__(self, dut: Dut, prompt: str, cmd: Optional[List[str]] = None, prompt_timeout: Optional[int] = -1):
+    def __init__(
```
nalajcie, line 55: FIX COMMIT MESSAGE !!!! 

(already written twice in this PR, making it a separate comment so it won't be ignored again)

nalajcie (changes requested):

trunner/harness/psh.py
maska989: Sorry for that, I misunderstood.

trunner/harness/psh.py
```diff
@@ -52,14 +52,22 @@ class ShellHarness(IntermediateHarness):
...
+    ):
         super().__init__()
         self.dut = dut
         self.prompt = prompt
         self.cmd = " ".join(map(shlex.quote, cmd)) if cmd is not None else cmd
         self.prompt_timeout = prompt_timeout
+        self.suppress_dmesg = suppress_dmesg
 
-    def __call__(self, result: TestResult) -> TestResult:
+    def assert_self_prompt(self):
```
nalajcie, line 70: IMHO `assert_prompt` name would be enough, not sure what `self` means in this context

nalajcie (commented):
one small comment, other than that LGTM.

nalajcie (approved):

trunner/harness/psh.py
```diff
@@ -81,4 +97,11 @@ def __call__(self, result: TestResult) -> TestResult:
                 ) from e
 
         result.set_stage(TestStage.RUN)
-        return self.next_harness(result)
+        test_result = self.next_harness(result)
+
+        # re-enable log output to release collected output
+        if self.suppress_dmesg:
+            self.dut.pexpect_proc.sendline("dmesg -E")
```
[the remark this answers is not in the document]
maska989: There is strange error on imxrt 117x while this change are applied, I will investigate that ASAP

psh/tools/psh.py
```diff
@@ -285,7 +285,9 @@ def run(harness):
     def wrapper_harness(*args, **kwargs):
         # to avoid CI fails caused by https://github.com/phoenix-rtos/phoenix-rtos-project/issues/866
         if trunner.ctx.target.name == "armv7m7-imxrt117x-evk":
-            return harness(*args, **kwargs)
+            x = harness(*args, **kwargs)
```
damianloew, line 288:
```suggestion
            ret = harness(*args, **kwargs)
```

psh/tools/psh.py
```diff
@@ -285,7 +285,9 @@ def run(harness):
     def wrapper_harness(*args, **kwargs):
         # to avoid CI fails caused by https://github.com/phoenix-rtos/phoenix-rtos-project/issues/866
         if trunner.ctx.target.name == "armv7m7-imxrt117x-evk":
```
damianloew, line 287: And then we can remove such line below.
```suggestion
        dut = args[0]
        if trunner.ctx.target.name == "armv7m7-imxrt117x-evk":
```

psh/tools/psh.py
```diff
@@ -285,7 +285,9 @@ def run(harness):
...
-            return harness(*args, **kwargs)
+            x = harness(*args, **kwargs)
+            args[0].sendline("\n")
```
damianloew, line 289:
```suggestion
            # one prompt after a test should be left to be asserted by test runner
            dut.sendline("\n")
```

psh/tools/psh.py
```diff
@@ -285,7 +285,9 @@ def run(harness):
...
+            return x
```
damianloew, line 290:
```suggestion
            return ret
```

damianloew (changes requested):

psh/tools/psh.py
```diff
@@ -23,45 +23,51 @@
...
-    ''' Sends specified command and asserts that it's displayed correctly
-    with optional expected output and next prompt. Exit status is asserted depending on `result`'''
+        raise Exception(
+            "".join(
+                (
+                    f"The value {result} for `result` argument is not correct. ",
+                    "Please choose between susccess/fail/dont-check",
+                )
+            )
+        )
```
mateusz-bloch, line 59: I'm not sure if using this formatting must be applied in every case, I think the previous form was more clear, although this is just my opinion.

Especially since in this case, `join` was used to shorten the lines.

mateusz-bloch: Maybe this would be the clearest approach, but again, this is just my opinion.

```suggestion
        raise Exception(
            f"The value {result} for `result` argument is not correct. "
            "Please choose between susccess/fail/dont-check'."
        )
```

damianloew (commented):
About the commit message: You haven't corrected all files in psh directory, so:
 - `psh: fix formating` -> `psh: fix formatting in psh.py`

Please apply the @mateusz-bloch suggestion and we can merge it

damianloew (approved):
LGTM

PR #326, opened 2024-02-20
network: add tcp test

adamdebek:
JIRA: CI-343

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


mateusz-bloch (commented):
I understand that this is a draft and WIP, So these are just some inital suggestions, if you're using unity_harness() in its unchanged form, you might consider importing it from trunner, something like `from trunner.harness.unity import unity_harness`. This also applies to en_setup(), host_interface() (`from psh.tools.netset import en_setup, host_interface`). I know these changes haven't been implemented yet, and if merging the ping takes too long, perhaps Maska989 could submit netset as a separate PR. Additionally, I was wondering if, will be more test cases, it might be good idea to create a class that would initialize the network at the beginning and then have a function responsible for launching a specific test along with the harness (?)

network/network.py
```diff
@@ -0,0 +1,177 @@
...
+        peer_socket, peer_address = host_socket.accept()
+
+    host_socket.close()
+
+    print('closed')
+
+    return peer_socket, peer_address
+
+
+def harness(dut: Dut, ctx: TestContext, result: TestResult):
```
mateusz-bloch, line 117: In such a case, do not copy another function; it would suffice to write a wrapper or sth

adamdebek: Yea I know it's draft

network/network.py
```diff
@@ -0,0 +1,135 @@
...
+    ifconfig_setup_cmd = "ifconfig en1 192.168.72.120 netmask 255.255.255.0 up"
+
+    psh.assert_prompt_after_cmd(p, ifconfig_setup_cmd, "success")
+
+
+def con_setup(dut: Dut, host_ip, host_port):
+    try:
+        host_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        host_socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        host_socket.bind(('172.17.0.2', host_port))
```
mateusz-bloch, line 50: I'm thinking if you're using port forwarding through Docker, shouldn't you generally bind the socket in the container to something like `0.0.0.0:1025` or simply `:1025 `so that it listens on all interfaces inside the container, to enabling port forwarding?

adamdebek: No, binding in this manner has nothing to do with port forwarding. I just used -p option in docker run command.

julianuziemblo (changes requested):

network/network_unity.py
```diff
@@ -0,0 +1,199 @@
...
+            ), "".join(("There is a mismatch between the number of parsed tests and overall results!\n",
+                        "Parsed results from the final Unity message (total, failed, ignored): ",
+                        f"{parsed['total']}, {parsed['fail']}, {parsed['ignore']}\n",
+                        "Found test summary lines (total, failed, ignored): ",
+                        f"{sum(stats.values())}, {stats['FAIL']}, {stats['IGNORE']}"))
+
+            break
+
+        elif idx == 4:
+            network_test_responder.do_cmd(parsed["cmd"])
```
julianuziemblo: `match ... case` could be used instead, but this is a purely stylistic decision

adamdebek: `match case` was introduced in `python 3.10` and some of our raspberries use still python 3.9

nalajcie: we should be running inside gh-runner container (ubuntu 24.04 based) - so the python version should be at least `3.12`.

network/common.c
```diff
@@ -0,0 +1,235 @@
...
+
+		if (ifr.ifr_flags & IFF_RUNNING) {
+			close(sock);
+			return true;
+		}
+
+		usleep(200000);  // 200 ms
+	} while (--try);
+
+	return false;
```
mateusz-bloch: Not closing sock after timeout?

network/network_unity.py
```diff
@@ -0,0 +1,300 @@
...
+            if conn is None:
+                print(f"Send received on None object (idx: {idx})")
+                return
+
+            try:
+                conn.send_received(size)
+            except (socket.timeout, BlockingIOError):
+                conn.events |= self.EVENT_SEND_BLOCKED
+            except socket.timeout:
+                print("Timeout during send_received")
```
mateusz-bloch: 2 times except same exception

network/test.yaml
```diff
@@ -0,0 +1,15 @@
...
+      execute: test_tcp en1
+      network_setup: True
+      targets:
+        value: [armv7a7-imx6ull-evk, armv7m7-imxrt106x-evk]
+
+    - name: tcp
+      execute: test_tcp en2
+      network_setup: True
+      targets:
+        value: [armv7m7-imxrt117x-evk]
```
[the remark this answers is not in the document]
adamdebek: I only used the interfaces that had a cable plugged in inside the rack units. I know the other interfaces differ in hardware on those targets, are they supported?

network/network_unity.py
adamdebek: Yes, but I was running these tests locally during development, and debugging is also done locally. Despite that, I prefer using if statements rather than `match case` when there are only a few cases

trunner/harness/psh.py
```diff
@@ -1,8 +1,9 @@
+import time
 import re
+import pexpect
 import shlex
-from typing import Optional, List
 
-import pexpect
```
damianloew: why the import order is changed? You should only remove List in this case.

trunner/harness/psh.py
```diff
@@ -40,72 +41,82 @@ def __str__(self):
...
     def __init__(
         self,
         dut: Dut,
-        prompt: str,
-        cmd: Optional[List[str]] = None,
-        prompt_timeout: int = -1,
-        suppress_dmesg: bool = True,
+        prompt: str
     ):
         super().__init__()
```
damianloew: It doesn't make sense - you don't have a parent class.

network/network_unity.py
lukkrusz: @damianloew given our previous discussion in the [PR about compatibility](https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/457) that resulted in the decision to upgrade the OS on our rasp pi's, should we stop keeping the code compatible with python 3.9?

If so, I too would be in favour of turning the `if/elif/else` into `match ... case` here

PR #330, opened 2024-02-23
psh: make test-kill more flexible

damianloew:
JIRA: CI-415

## Motivation and Context

Nedded for  https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/215

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/psh/test-kill.py
+++ b/psh/test-kill.py
@@ -19,40 +19,50 @@ from time import sleep
 
 def assert_kill_procs(p, pid_list):
     for pid in pid_list:
-        msg = f'Wrong output when killing process with the following pid: {pid}'
-        psh.assert_cmd(p, f'kill {pid}', result='dont-check', msg=msg)
+        msg = f"Wrong output when killing process with the following pid: {pid}"
+        psh.assert_cmd(p, f"kill {pid}", result="dont-check", msg=msg)
 
     sleep(0.5)
 
-    dead = set(pid_list) - set(proc['pid'] for proc in get_process_list(p))
-    assert len(dead) == len(pid_list), f'failed to kill {len(pid_list) - len(dead)} processes out of {len(pid_list)}'
+    dead = set(pid_list) - set(proc["pid"] for proc in get_process_list(p))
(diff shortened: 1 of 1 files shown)


psh/test-kill.py
```diff
@@ -19,29 +19,50 @@
...
+        result="fail",
+        msg=msg,
+        is_regex=True,
+    )
+
+    msg = "Wrong kill help message"
+    psh.assert_cmd(
+        p,
+        "kill",
+        expected=rf"(kill: |)(u|U)sage: kill[^\r]+{psh.EOL}",
```
adamdebek, line 51: Make it search lazily, greedy search could result in not matching the regex, because `psh.EOL` first character is not `\r`
```suggestion
        expected=rf"(kill: |)(u|U)sage: kill[^\r]+?{psh.EOL}",
```
same in other places

adamdebek, line 51: I wonder why here are 2 ORs in this regex, output is not predefined?

adamdebek (commented):
If this test try to kill processes you should send `SIGKILL` signal, you are sending default signal from kill, which is `SIGTERM`, it works but to be 100% sure you should send `SIGKILL`

psh/test-kill.py
damianloew: kill when called on Ubuntu returns `kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]` I want to accept also this `kill:` prefix, Usage may be printed both with uppercase and lowercase

damianloew:
> If this test try to kill processes you should send `SIGKILL` signal, you are sending default signal from kill, which is `SIGTERM`, it works but to be 100% sure you should send `SIGKILL`

I think we can change it later, after merging the mentioned changes - to provide proper CI for the PR in utils

psh/test-kill.py
damianloew: Done

adamdebek (approved):

PR #331, opened 2024-03-04
libc: unix-socket: separate test cases

adamdebek:
JIRA: CI-404

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/socket/unix-socket.c
+++ b/libc/socket/unix-socket.c
@@ -106,7 +106,7 @@ TEST_TEAR_DOWN(test_unix_socket)
 }
 
 
-TEST(test_unix_socket, unix_zero_len_send)
+TEST(test_unix_socket, zero_len_send)
 {
 	int fd[3];
 	struct msghdr msg;
@@ -230,7 +230,7 @@ TEST(test_unix_socket, unix_zero_len_send)
 }
 
 
-TEST(test_unix_socket, unix_zero_len_recv)
+TEST(test_unix_socket, zero_len_recv)
 {
 	int fd[2];
 	struct msghdr msg;
@@ -314,7 +314,7 @@ TEST(test_unix_socket, unix_zero_len_recv)
 }
 
 
-TEST(test_unix_socket, unix_close)
+TEST(test_unix_socket, close)
 {
 	unsigned int i;
 	int fd[2];
@@ -405,7 +405,8 @@ void unix_msg_data_only(int type)
 	close(fd[1]);
 }
 
-TEST(test_unix_socket, unix_msg_data_only)
+
+TEST(test_unix_socket, msg_data_only)
 {
 	unix_msg_data_only(SOCK_STREAM);
 	unix_msg_data_only(SOCK_DGRAM);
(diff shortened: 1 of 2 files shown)


libc/test.yaml
```diff
@@ -67,7 +67,7 @@ test:
       execute: test-libc-unix-socket
       targets:
         # Excluded due to long execution time
-        exclude: [armv7m4-stm32l4x6-nucleo]
+        exclude: [armv7m4-stm32l4x6-nucleo, armv7a9-zynq7000-qemu]
```
mateusz-bloch, line 70: I see, the issue with the tests taking too long relates to the test cases:` unix_msg_data_and_fd` and `unix_msg_fork`. Could we split each of these into two separate test cases, for example one for `SOCK_STREAM` and one for `SOCK_DGRAM`? This way, we could run them on every target, given that the issue here is simply that the Nucleo and Zynq-QEMU targets are slow.

libc/socket/unix-socket.c
```diff
@@ -405,9 +405,15 @@ void unix_msg_data_only(int type)
 	close(fd[1]);
 }
 
-TEST(test_unix_socket, unix_msg_data_only)
+
+TEST(test_unix_socket, unix_stream_sock_send_data)
```
mateusz-bloch, line 409: Not that test case? From what I see, these should be changed:  `unix_msg_data_and_fd` & `unix_msg_fork` :

Zynq-qemu
image

stm32:
image

mateusz-bloch (commented):
Also change the commit message

libc/test.yaml
```diff
@@ -67,7 +67,6 @@ test:
       execute: test-libc-unix-socket
       targets:
         # Excluded due to long execution time
-        exclude: [armv7m4-stm32l4x6-nucleo]
```
mateusz-bloch, line 70: delete `# Excluded due to long execution time`

libc/socket/unix-socket.c
```diff
@@ -475,9 +476,14 @@ void unix_msg_data_and_fd(int type)
 }
 
 
-TEST(test_unix_socket, unix_msg_data_and_fd)
+TEST(test_unix_socket, unix_stream_sock_send_and_fd)
```
damianloew, line 479: I think that we can remove this `unix` prefix as it's mentioned in the name of test group. Then we can name it `msg_data_and_fd_strem`. The current name is misleading - send and file descriptor, send what (?), also it's not only sending, but receiving too. IMHO it's better to keep the previous convention, where we have test cases like `data_only` and `data_and_fd`

damianloew: We can name the same way the rest of changed cases.

damianloew (changes requested):
Thanks for the PR, have you made sure that fails are no longer seen?

mateusz-bloch:
> Thanks for the PR, have you made sure that fails are no longer seen?

Yes,  from what I remember, in the worst-case scenario there's about a +-12-second safety margin.

damianloew (approved):
LGTM, Could you please update submodules after a merge?

PR #336, opened 2024-03-15
trunner: wait for possible plo interruption

maska989:
JIRA: CI-432

## Description
In case of 'waiting for input' we interrupt PSH loading. We need to wait a bit to be sure that in this case will be able to do this on faster host environments where this can occur a race condition between sending command and non-accessible input collector.

## Types of changes

## How Has This Been Tested?

- [X] Tested by hand on: armv7m4-stm32l4x6-nucleo (On host env).

## Checklist:

- [X] I have added tests to cover my changes.

## Special treatment

### DIFF
--- a/trunner/harness/plo.py
+++ b/trunner/harness/plo.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from typing import Callable, Optional, Sequence
+import time
 
 import pexpect
 
@@ -76,6 +77,9 @@ class PloInterface:
             self.dut.expect_exact("Waiting for input", timeout=3)
         except (pexpect.TIMEOUT, pexpect.EOF) as e:
             raise PloError("Failed to go into bootloader mode", output=self.dut.before) from e
+
+        # Temporary fix for [https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1040]
+        time.sleep(0.1)
 
         self.dut.send("\n")
 


nalajcie:
I believe you've meant `plo`, not `psh`. Still, we should fix our code in `plo`, if it doesn't work as expected - not work around it.

trunner/harness/plo.py
```diff
@@ -77,6 +78,7 @@ def enter_bootloader(self):
         except (pexpect.TIMEOUT, pexpect.EOF) as e:
             raise PloError("Failed to go into bootloader mode", output=self.dut.before) from e
 
+        time.sleep(0.1)
```
mateusz-bloch, line 81: Wouldn't it be better to use dut to expect a message that should be followed by a `\n`, instead of using a sleep, so we are likely covered for most cases, regardless of the host's speed?
For example, maybe `waiting for input `if I remember correctly.

nalajcie (dismissed):
I would just simply write `temporary fix for [issue URL]`. Still, LGTM.

trunner/harness/plo.py
```diff
@@ -75,7 +76,12 @@ def enter_bootloader(self):
         try:
             self.dut.expect_exact("Waiting for input", timeout=3)
         except (pexpect.TIMEOUT, pexpect.EOF) as e:
-            raise PloError("Failed to go into bootloader mode", output=self.dut.before) from e
+            raise PloError(
+                "Failed to go into bootloader mode", output=self.dut.before
+            ) from e
+
+        # Temporary fix for [https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1040]
+        time.sleep(0.2)
```
damianloew, line 84: I asked whether the 0.1 sleep is sufficient - I know that you have increased it to be sure, but IMHO if it worked fine we should leave 0.1 - we don't like sleeps, especially longer ones.

trunner/harness/plo.py
```diff
@@ -112,7 +118,9 @@ def send_cmd(self, cmd: str):
             self.dut.send(cmd + "\r")
             self.dut.expect_exact(cmd)
         except pexpect.TIMEOUT as e:
-            raise PloError("failed to read echoed command", cmd=cmd, output=self.dut.before) from e
+            raise PloError(
```
damianloew, line 121: Please place black corrections in the separate commit - it will be misleading when you click on those lines and see `add sleep` commit.

trunner/harness/plo.py
```diff
@@ -75,7 +76,12 @@ def enter_bootloader(self):
         try:
             self.dut.expect_exact("Waiting for input", timeout=3)
         except (pexpect.TIMEOUT, pexpect.EOF) as e:
-            raise PloError("Failed to go into bootloader mode", output=self.dut.before) from e
+            raise PloError(
```
nalajcie, line 79: styling changes are not needed, we have a soft limit of 120 chars per line for python - please configure your editor correctly :exclamation: 

(eg. https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/master/.github/workflows/runner-ci.yaml#L43)

trunner/harness/plo.py
maska989: Autoformat collected this, sadly (Not done on purpose)

trunner/harness/plo.py
maska989: Funny thing that I have this configuration and it formats this 
image

damianloew (approved):

PR #337, opened 2024-03-15
armv7m4: add stlink host reset

maska989:
JIRA: CI-433

## Description
Implemented soft reset for host environment to run test runner without using GPIO's.

## Types of changes

## How Has This Been Tested?

- [X] Tested by hand on:  armv7m4-stm32l4x6-nucleo (On host env).

## Checklist:

- [X] I have added tests to cover my changes.

## Special treatment

- [X] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/336

### DIFF
--- a/trunner/target/armv7m4.py
+++ b/trunner/target/armv7m4.py
@@ -46,9 +46,16 @@ class STM32L4x6OpenocdGdbServerHarness(IntermediateHarness):
 
     def __call__(self, result: TestResult) -> TestResult:
         # Set of config parameters used in Openocd to flash up stm32l4a6
-        openocd_args = ["-c", "reset_config srst_only srst_nogate connect_assert_srst", "-c", "init;reset"]
+        openocd_args = [
+            "-c",
+            "reset_config srst_only srst_nogate connect_assert_srst",
+            "-c",
+            "init;reset",
+        ]
 
-        with OpenocdGdbServer(interface="stlink", target="stm32l4x", extra_args=openocd_args).run():
+        with OpenocdGdbServer(
+            interface="stlink", target="stm32l4x", extra_args=openocd_args
+        ).run():
             self.harness(result)
 
         return self.next_harness(result)
@@ -94,7 +101,6 @@ class STM32L4x6PloAppLoader(TerminalHarness, PloInterface):
         for app in self.apps:
             path = self.gdb.cwd / Path(app.file)
             sz = path.stat().st_size
-
             self.alias(app.file, offset=offset, size=sz)
             self.app("ramdev", app.file, "ram", "ram")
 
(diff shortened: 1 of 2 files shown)


trunner/target/armv7m4.py
```diff
@@ -24,9 +25,26 @@
 
 
 class ARMv7M4Rebooter(Rebooter):
-    # TODO add text mode
-    # NOTE: changing boot modes not needed/supported for this target
-    pass
+    def _reboot_dut_text(self, hard):
+        # self.dut.send(" reboot\r\n")
```
mateusz-bloch, line 29: To delete ?

trunner/target/armv7m4.py
```diff
@@ -24,9 +25,26 @@
...
+                "-c",
+                "reset_config srst_only srst_nogate connect_assert_srst",
+                "-c",
+                "init;reset;exit",
+            ],
+            encoding="ascii",
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            timeout=20,
+            cwd=os.getcwd(),
```
mateusz-bloch, line 46: I wouldn't use the os module, since `TargetBase` should have specific paths to certain locations created (for example `self.project_dir`).

Also, I'm thinking if we should check if the reset was successful

trunner/target/armv7m4.py
```diff
@@ -46,9 +64,16 @@ def __init__(self, harness: Callable[[TestResult], TestResult]):
...
+        openocd_args = [
+            "-c",
+            "reset_config srst_only srst_nogate connect_assert_srst",
+            "-c",
+            "init;reset",
+        ]
+
+        with OpenocdGdbServer(
+            interface="stlink", target="stm32l4x", extra_args=openocd_args
+        ).run():
```
mateusz-bloch, line 76: It not needed here, but for the future, I it might be better to separate code formatting changes from actual code changes into a separate commit, as it could improve the readability of changes/review

trunner/target/armv7m4.py
```diff
@@ -24,9 +25,26 @@
 
 
 class ARMv7M4Rebooter(Rebooter):
-    # TODO add text mode
-    # NOTE: changing boot modes not needed/supported for this target
-    pass
+    def _reboot_dut_text(self, hard):
```
damianloew, line 28: According to our last meeting it will be better to add another method, for example called `_reboot_dut_command`, because as you can see sending a reset command to onboard debugger is definitely not a text mode.
```suggestion
    def _reboot_dut_command(self, hard):
```

Then we should introduce a launch_context, where we could specify what reset do we want to use. For now we can just change it temporary this way if needed:
Screenshot from 2024-04-15 12-24-41

trunner/target/armv7m4.py
maska989: Sadly, this is not possible, OpenOCD throw non-zero status even if MCU voltage reading is below 3.3v (In our projects this reading is not even used and causing non-zero status return)

damianloew:
A am closing it as it's done in https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/382

PR #342, opened 2024-03-29
libtinyaes: adapt tests for execution in a test runner and CI

mateusz-bloch:
JIRA: CI-441

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


libtinyaes/test_unit_aes_ccm_s.c
```diff
@@ -1,13 +1,11 @@
-#ifdef WITH_AES_CCM_S
```
anglov, line 1: It would be great if you could add header with original authors and copyright dates of test.

Consider adding yourself with current year. 

Applied to all missing headers.

anglov (commented):
Thanks for making this tests working.

One  small comment I found in moved sources:

You should change all occurences of PS_DEBUG to something used in current tests to make it vebose. PS_DEBUG is obsolete in new code.

libtinyaes/test_unit_aes_ccm_s.c
gerard5: Add also me  (c) 2021, because I'm the original author of eax and ccm which are used in g3-plc stack.

anglov: Yep it was my intention saying that please add valid authors :)

GCM tests author is @xvuko  (c) 2023

libtinyaes/test_unit_aes_gcm.c
```diff
@@ -1,3 +1,15 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Tests for AES GCM
+ *
+ * Copyright 2023, 2024 by Phoenix Systems
```
anglov, line 6: nit: 2023-2024

anglov (commented):
Small nitpick, otherwise LGTM

NOTE: I'm not convinced with DEBUG variable as we don't have possibilty to enable it and it can be misleaded with DEBUG which is used to omit NDEBUG define. 

However looking at possibilities we don't have common approach for leaving debug code in tests. So I think I'm ok with that at this moment.

anglov (approved):

damianloew:
> Small nitpick, otherwise LGTM
> 
> NOTE: I'm not convinced with DEBUG variable as we don't have possibilty to enable it and it can be misleaded with DEBUG which is used to omit NDEBUG define.
> 
> However looking at possibilities we don't have common approach for leaving debug code in tests. So I think I'm ok with that at this moment.

Yes, we don't have such cases in the existent tests - I would also leave DEBUG for now

damianloew (approved):

damianloew (approved):

PR #343, opened 2024-03-29
unity tests: return non zero exit code when test failed

mateusz-bloch:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/devices/grlib-multi/test-grlib-multi.c
+++ b/devices/grlib-multi/test-grlib-multi.c
@@ -775,6 +775,5 @@ void runner(void)
 
 int main(int argc, char *argv[])
 {
-	UnityMain(argc, (const char **)argv, runner);
-	return 0;
+	return (UnityMain(argc, (const char **)argv, runner) == 0) ? EXIT_SUCCESS : EXIT_FAILURE;
 }
--- a/ioctl/test-ioctl.c
+++ b/ioctl/test-ioctl.c
@@ -395,7 +395,7 @@ int main(int argc, char *argv[])
 		return 1;
 	}
 
-	UnityMain(argc, (const char **)argv, runner);
+	int failures = UnityMain(argc, (const char **)argv, runner);
 	portDestroy(test_port);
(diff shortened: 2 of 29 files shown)


devices/grlib-multi/test-grlib-multi.c
```diff
@@ -775,6 +775,5 @@ void runner(void)
 
 int main(int argc, char *argv[])
 {
-	UnityMain(argc, (const char **)argv, runner);
-	return 0;
+	return UnityMain(argc, (const char **)argv, runner) == 0 ? EXIT_SUCCESS : EXIT_FAILURE;
```
damianloew, line 778: The problem is still not resolved, please check the following code and just run in on host:

```C
#include <stdio.h>
#include <stdlib.h>

int main(void)
{
	return 256 == 0 ? EXIT_SUCCESS : EXIT_FAILURE;
}
```

mateusz-bloch: In the scenario you provided, it will always return 1, because of EXIT_FAILURE

damianloew: oh, that's right - it seems to work fine

devices/grlib-multi/test-grlib-multi.c
```diff
@@ -775,6 +775,5 @@ void runner(void)
```
damianloew, line 775: Please correct it also in waitpid test. We should also consider corrections in other repos - for example phoenix-rtos devices

devices/grlib-multi/test-grlib-multi.c
damianloew, line 778: I know that this version is also correct, but to make it more clear I would add parentheses:
```suggestion
	return (UnityMain(argc, (const char **)argv, runner) == 0) ? EXIT_SUCCESS : EXIT_FAILURE;
```

damianloew (approved):
LGTM

PR #348, opened 2024-04-15
trunner: change style of printing flash output

mateusz-bloch:
### Before: 
for example: 
- ull
image
- imx106x:
image

### After (running with -S option) : 
for example: 
- ull
image

- imx106x:
image

## Types of changes

## How Has This Been Tested?

- [x] New test added: (add PR link here).

## Checklist:

## Special treatment

### DIFF
--- a/trunner/test_runner.py
+++ b/trunner/test_runner.py
@@ -11,7 +11,7 @@ from trunner.config import ConfigParser
 from trunner.ctx import TestContext
 from trunner.dut import Dut
 from trunner.harness import HarnessError, FlashError
-from trunner.text import bold, green, red, yellow, magenta
+from trunner.text import green, red, yellow, magenta
 from trunner.types import Status, TestOptions, TestResult, TestStage, is_github_actions, get_ci_url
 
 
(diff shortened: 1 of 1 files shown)


trunner/test_runner.py
```diff
@@ -143,11 +143,12 @@ def parse_tests(self) -> Sequence[TestOptions]:
     def flash(self) -> TestResult:
         """Flashes the device under test."""
 
-        print("Flashing an image to device...")
-
         # report flashing as a test result to include in export (especially useful when failed)
         result = TestResult("flash")
 
+        self._print_test_header_begin(result)
+        print("Flashing an image to device...")
```
damianloew, line 150:
```suggestion
```

It may be a bit less readable, but for consistency it would be better to have `flash: OK`, now it looks like this:

Screenshot from 2024-04-16 11-52-33

trunner/test_runner.py
```diff
@@ -158,7 +158,10 @@ def flash(self) -> TestResult:
             print(exc)
             result.fail_harness_exception(exc)
 
-        print("Done!")
+        self._print_test_header_end(result)
+
+        print(result.status)
```
damianloew, line 163: I think that we should use `result.to_str` as it's done in tests, because now it's not consistent, we want to also have `flash: FAIL` and then output in such case:

Screenshot from 2024-04-16 12-35-48


Screenshot from 2024-04-16 12-36-00

mateusz-bloch: After the recent changes, it now looks like this:
image

damianloew (approved):
Thanks for improvements, LGTM

PR #350, opened 2024-04-24
libc: increase time limits in unix socket test

adamdebek:
JIRA: CI-452

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


nalajcie:
- maybe we should use ` TEST_ASSERT_LESS_THAN` and other similar for integer comparison - so we would get the actual value on error (would make it easier to tune the values)
- probably the constraints should be either per-target-family or different on emulated targets (where we don't expect even soft realtime?) - but that's a broader topic not to be breached right now

damianloew (approved):
I am not sure whether the changed values will be 100% fine, but I'd opt for thinking about it in detail when you upload your next improvements in the separate commit.

PR #353, opened 2024-06-04
test-ps: Adapt to change in kernel thread name

agkaminski:
JIRA: RTOS-845

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

- [x] Already covered by automatic testing.
- [ ] New test added: (add PR link here).
- [ ] Tested by hand on: (list targets here).

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing linter checks and tests passed.
- [x] My changes generate no new compilation warnings for any of the targets.

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/556).
- [ ] I will merge this PR by myself when appropriate.

PR #354, opened 2024-06-04
psh: support both [idle] and [kthread] in ps test

damianloew:
JIRA: CI-455

## Description

Flexible alternative for https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/353#pullrequestreview-2096712711

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/psh/test-ps.py
+++ b/psh/test-ps.py
@@ -3,7 +3,7 @@ import psh.tools.psh as psh
 """
 (psh)% ps
   PID  PPID PR STATE  %CPU  WAIT    TIME   VMEM THR CMD
-    0     0  7 ready  97.1  2.9s   31:25   2.6M   1 [idle]
+    0     0  7 ready  97.1  2.9s   31:25   2.6M   1 [kthread]
     5     1  1 sleep   1.7 295ms    0:37   124K   3 uart16550
     4     1  4 ready   0.6   7ms    0:12   140K   1 psh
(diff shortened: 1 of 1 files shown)


psh/test-ps.py
```diff
@@ -43,6 +43,9 @@ def harness(p):
                 task = 'psh'
             if task in expected_tasks:
                 expected_tasks.remove(task)
+            # to provide compatibility with older kernel versions, where it's [idle] not [kthread]
+            for task in ('[kthread]', '[idle]'):
```
nalajcie, line 47: uh, overwriting the `task` from outer scope, this will essentially remove `ktread` and `init` even only if `psh` is present

psh/test-ps.py
```diff
@@ -43,6 +43,9 @@ def harness(p):
...
+                expected_tasks.remove(task) if task in expected_tasks else expected_tasks
```
nalajcie, line 48: no assignment here, so the `else` value is not used.

maaybe the simplest solution would be to just allow old synonym
```python
if task == '[idle]'  # allow old name of kernel thread
  task = '[kthread]'
```
(and keep `kthread` only in expected tasks

psh/test-ps.py
damianloew: Right - I missed it, thanks for the suggestion

mateusz-bloch (approved):
So now it looks good.
I would only optimize line handling in the test, but that's not the case of this PR.

PR #358, opened 2024-06-12
trunner: prevent printing escape codes in output from fail

mateusz-bloch:
JIRA: CI-460

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


nalajcie:
Why? IMHO on error we should show data as-is  (even garbage) as it might be a part of (or a reason for) an error?

The regex is slightly invalid, also we already have `remove_ansi_sequences` function available.

damianloew:
> Why? IMHO on error we should show data as-is (even garbage) as it might be a part of (or a reason for) an error?
> 
> The regex is slightly invalid, also we already have `remove_ansi_sequences` function available.



> Why? IMHO on error we should show data as-is (even garbage) as it might be a part of (or a reason for) an error?
> 
> The regex is slightly invalid, also we already have `remove_ansi_sequences` function available.

The name of task suggested that we could skip clear screen esc code - I didn't mean to escape all of them. I was considering such solution, because we are running tests using trunner on images that refresh, clear and print the whole output once again and again. Then in case of fail the output is not readable

nalajcie:
> The name of task suggested that we could skip clear screen esc code - I didn't mean to escape all of them. I was considering such solution, because we are running tests using trunner on images that refresh, clear and print the whole output once again and again. Then in case of fail the output is not readable

I still don't get it, the log should be as verbose as it could (if we're removing data from the output we might remove why regex match has failed) . If we're talking about better presentation on screen - it should be done at a different level (just before printing message to screen)?

Also please note that current implementation is slow - using simple `replace` twice would probably be faster. If we would have multiple things to remove - the least you could do is compile the regex beforehand.

PR #361, opened 2024-07-02
unity: add handling fails without message in harness

adamdebek:
JIRA: CI-464

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/harness/unity.py
+++ b/trunner/harness/unity.py
@@ -6,7 +6,7 @@ from trunner.types import Status, TestResult
 
 
 def unity_harness(dut: Dut, ctx: TestContext, result: TestResult) -> Optional[TestResult]:
-    assert_re = r"ASSERTION (?P<path>[\S]+):(?P<line>\d+):(?P<status>FAIL|INFO|IGNORE): (?P<msg>.*?)\r"
+    assert_re = r"ASSERTION [\S]+:\d+:(?P<status>FAIL|INFO|IGNORE)(: (?P<msg>.*?))?\r"
     result_re = r"TEST\((?P<group>\w+), (?P<name>\w+)\) (?P<status>PASS|IGNORE)"
     # Fail need to have its own regex due to greedy matching
     result_fail_re = r"TEST\((?P<group>\w+), (?P<name>\w+)\) (?P<status>FAIL) at (?P<path>.*?):(?P<line>\d+)\r"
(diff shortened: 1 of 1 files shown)


trunner/harness/unity.py
```diff
@@ -27,14 +27,15 @@ def unity_harness(dut: Dut, ctx: TestContext, result: TestResult) -> Optional[Te
...
-                parsed["msg"] = last_assertion["msg"]
+                msg = last_assertion["msg"]
                 last_assertion = {}
 
             status = Status.from_str(parsed["status"])
             subname = f"{parsed['group']}.{parsed['name']}"
             if "path" in parsed and "line" in parsed:
-                parsed["msg"] = f"[{parsed['path']}:{parsed['line']}] " + parsed["msg"]
+                parsed["msg"] = f"[{parsed['path']}:{parsed['line']}] " + msg
             result.add_subresult(subname, status, parsed.get("msg", ""))
```
mateusz-bloch, line 39: don't need to use `.get` when we have declared msg earlier as empty, also can put everything into a formatted string
```suggestion
                msg = f"[{parsed['path']}:{parsed['line']}] {msg}"
            result.add_subresult(subname, status, msg)
```

trunner/harness/unity.py
```diff
@@ -27,14 +27,15 @@ def unity_harness(dut: Dut, ctx: TestContext, result: TestResult) -> Optional[Te
...
             if last_assertion:
-                parsed["msg"] = last_assertion["msg"]
+                msg = last_assertion["msg"]
                 last_assertion = {}
 
             status = Status.from_str(parsed["status"])
             subname = f"{parsed['group']}.{parsed['name']}"
             if "path" in parsed and "line" in parsed:
-                parsed["msg"] = f"[{parsed['path']}:{parsed['line']}] " + parsed["msg"]
+                parsed["msg"] = f"[{parsed['path']}:{parsed['line']}] " + msg
```
nalajcie, line 38: - this means that test can fail without any assert? 
 - please note that you're changing the semantics (the msg would be added to subresult only if `path` and `line` are present - not sure if relevant. Maybe just make a minor change (without the changes above):
```suggestion
                parsed["msg"] = f"[{parsed['path']}:{parsed['line']}] " + parsed.get("msg", "")
```

mateusz-bloch: Hmm, from what I understand now, the issue is that the test might fail with an assert, but the regex won't catch it due to the lack of message on the end of assert. In that case, I would adjust the regex, that this `: (?P<msg>.*?)` can be optional

adamdebek: Test can fail without any message, for example using `TEST_FAIL()` and in this situation there was an error:
image

nalajcie: I'm not sure writing tests which can fail without an explicit reason should be encouraged.

Still, adding `parsed.get("msg", "")` probably saves the day?

adamdebek: Yea, I agree, but I think such a case should be taken into account and handled. I used your suggestion and also adjusted assert regex.

damianloew (approved):

PR #364, opened 2024-08-01
libc/socket: add unix socket tests

adamgreloch:
Adds tests of accept, connect, recv(MSG_PEEK)

JIRA: RTOS-862, RTOS-877

## Description

Adds unix socket accept/connect tests, both blocking and non-blocking, as well as a small test for proper recv(MSG_PEEK) behavior. 

## Types of changes

## How Has This Been Tested?

- [X] Tested by hand on: (list targets here).
ia32-generic-qemu

## Checklist:

## Special treatment

- [X] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/572
https://github.com/phoenix-rtos/libphoenix/pull/388

### DIFF
--- a/libc/socket/unix-socket.c
+++ b/libc/socket/unix-socket.c
@@ -6,7 +6,7 @@
  * unix socket tests
  *
  * Copyright 2021, 2024 Phoenix Systems
- * Author: Ziemowit Leszczynski, Adam Debek
+ * Author: Ziemowit Leszczynski, Adam Debek, Adam Greloch
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -16,6 +16,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <time.h>
@@ -27,18 +28,40 @@
 #include <arpa/inet.h>
 #include <signal.h>
 #include <poll.h>
+#include <sys/select.h>
 #include <sys/uio.h>
 #include <sys/wait.h>
 
 #include "common.h"
 #include "unity_fixture.h"
 
+#define BAD_FD 33333 /* should be bad descriptor */
+
+
+#define FAIL_OR_EXIT(pid, msg) \
+	do { \
+		if (pid) \
+			FAIL(msg); \
+		else \
+			exit(1); \
+	} while (0)
+
+
+#define CHILD_ASSERT(pred) \
+	do { \
+		if (!(pred)) \
+			exit(1); \
+	} while (0)
+
+
+#define MS_BETWEEN(ts0, ts1) \
+	((ts1).tv_sec - (ts0).tv_sec) * 1000 + ((ts1).tv_nsec - (ts0).tv_nsec) / 1000000;
 
 char data[DATA_SIZE];
 char buf[DATA_SIZE];
 
 
-ssize_t unix_named_socket(int type, const char *name)
+static ssize_t unix_named_socket(int type, const char *name)
 {
 	int fd;
 	struct sockaddr_un addr = { 0 };
@@ -60,7 +83,7 @@ ssize_t unix_named_socket(int type, const char *name)
 }
 
 
-int unix_connect(int fd, const char *name)
+static int connect_to_named(int fd, const char *name)
 {
 	struct sockaddr_un addr = { 0 };
 
@@ -71,7 +94,7 @@ int unix_connect(int fd, const char *name)
 }
 
 
-int unlink_files(size_t cnt)
+static int unlink_files(size_t cnt)
 {
 	size_t i;
 	char buf[64];
@@ -83,6 +106,21 @@ int unlink_files(size_t cnt)
 	}
 
 	return 0;
+}
+
+
+static pid_t safe_fork(void)
+{
+	pid_t pid;
+	if ((pid = fork()) < 0) {
+		if (errno == ENOSYS) {
+			TEST_IGNORE_MESSAGE("fork syscall not supported");
+		}
+		else {
+			FAIL("fork");
+		}
+	}
+	return pid;
 }
 
 
@@ -208,7 +246,7 @@ TEST(test_unix_socket, zero_len_send)
 		n = sendmsg(fd[0], &msg, 0);
 		TEST_ASSERT(n == 0);
 
-		fd[2] = 33333; /* should be bad descriptor */
+		fd[2] = BAD_FD;
 		memset(&msg, 0, sizeof(msg));
 		msg.msg_iov = &iov;
 		msg.msg_iovlen = 0;
@@ -370,7 +408,7 @@ TEST(test_unix_socket, close)
 }
 
 
-void unix_msg_data_only(int type)
+static void unix_msg_data_only(int type)
 {
 	unsigned int i;
 	int fd[2];
@@ -413,7 +451,7 @@ TEST(test_unix_socket, msg_data_only)
 }
 
 
-void unix_msg_data_and_fd(int type)
+static void unix_msg_data_and_fd(int type)
 {
 	int i;
 	int fd[2];
@@ -488,7 +526,7 @@ TEST(test_unix_socket, dgram_sock_data_and_fd)
 }
 
 
-void unix_msg_fork(int type)
+static void unix_msg_fork(int type)
 {
 	int fd[2];
 	pid_t pid;
@@ -499,14 +537,7 @@ void unix_msg_fork(int type)
 	if (socketpair(AF_UNIX, type, 0, fd) < 0)
 		FAIL("socketpair");
 
-	if ((pid = fork()) < 0) {
-		if (errno == ENOSYS) {
-			TEST_IGNORE_MESSAGE("fork syscall not supported");
-		}
-		else {
-			FAIL("fork");
-		}
-	}
+	pid = safe_fork();
 
 	if (pid) {
 		int sfd[MAX_FD_CNT];
@@ -549,13 +580,13 @@ void unix_msg_fork(int type)
 			exit(1);
 
 		if (read_files(rfd, rfdcnt, data, buf) < 0)
-			FAIL("read_files");
+			exit(1);
 
 		if (close_files(rfd, rfdcnt) < 0)
 			exit(2);
 
 		if (stat_files(rfd, rfdcnt, 0) < 0)
-			FAIL("stat_files");
+			exit(1);
 
 		exit(0);
 	}
@@ -582,7 +613,7 @@ TEST(test_unix_socket, dgram_sock_msg_fork)
 }
 
 
-int unix_data_cmp(char *buf, size_t pos, size_t len)
+static int unix_data_cmp(char *buf, size_t pos, size_t len)
 {
 	size_t i;
 
@@ -595,7 +626,7 @@ int unix_data_cmp(char *buf, size_t pos, size_t len)
 }
 
 
-void unix_transfer(int type)
+static void unix_transfer(int type)
 {
 	int fd[2];
 	pid_t pid;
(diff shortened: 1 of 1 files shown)


nalajcie:
a few generic remarks:
 - tests should not print extra info apart from the assertions/test results (we might add verbose mode in the future if needed) - it breaks the `trunner` state machine (especially printing lines without `\n`)
 - declare functions as `static` if you don't want to use them outside of this compilation unit
 - **using unity you can't ASSERT in different thread/process than the main one (failed assertion uses `longjmp`) - the errors from other threads/processes need to be read in a different way**  - see if it's feasible to use unity the use cases you want to test (the other option is to make separate test with custom harness)

adamgreloch:
> a few generic remarks:
> 
>   * tests should not print extra info apart from the assertions/test results (we might add verbose mode in the future if needed) - it breaks the `trunner` state machine (especially printing lines without `\n`)

Done

>   * declare functions as `static` if you don't want to use them outside of this compilation unit

Sure. Should I also make the rest of functions static in this test file that seem to be local to the test suite (not declared in any header)?
> 
>   * **using unity you can't ASSERT in different thread/process than the main one (failed assertion uses `longjmp`) - the errors from other threads/processes need to be read in a different way**  - see if it's feasible to use unity the use cases you want to test (the other option is to make separate test with custom harness)

I wasn't aware of that, thanks. Hopefully this is related only to asserts, but `FAIL()` is fine (as done in previous unix socket tests)?

nalajcie:
> > a few generic remarks:
> > 
> > * tests should not print extra info apart from the assertions/test results (we might add verbose mode in the future if needed) - it breaks the `trunner` state machine (especially printing lines without `\n`)
> 
> Done
> 
> > * declare functions as `static` if you don't want to use them outside of this compilation unit
> 
> Sure. Should I also make the functions for previous tests static in this test file that seem to be local to the test suite (not declared in any header)?

Yes, I think so.

> 
> > * **using unity you can't ASSERT in different thread/process than the main one (failed assertion uses `longjmp`) - the errors from other threads/processes need to be read in a different way**  - see if it's feasible to use unity the use cases you want to test (the other option is to make separate test with custom harness)
> 
> I wasn't aware of that, thanks. Hopefully this is related only to asserts, but `FAIL()` is fine (as done in previous tests)?

I don't think so, checking call stack it goes down to `longjmp`, so the current code is also invalid.

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+		FAIL("set_nonblock");
+
+	rv = connect_to_named(fd, socket_name);
+	TEST_ASSERT(rv < 0);
+	TEST_ASSERT_EQUAL_INT(EINPROGRESS, errno);
+
+	rv = connect_to_named(fd, socket_name);
+	TEST_ASSERT(rv < 0);
+	TEST_ASSERT_EQUAL_INT(EALREADY, errno);
+
```
ziemleszcz, line 1006: Could add EISCONN test case?

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+	rv = connect_to_named(fd, socket_name);
+	TEST_ASSERT(rv < 0);
+	TEST_ASSERT_EQUAL_INT(EINPROGRESS, errno);
+
+	fds[0].fd = named;
+	fds[0].events = POLLIN;
+	fds[1].fd = fd;
+	fds[1].events = POLLOUT;
+
+	TEST_ASSERT_EQUAL_INT(1, poll(fds, 1, 1000));
```
ziemleszcz, line 1058: Please add a comment indicating that this is done to verify the completion of `accept`. Consider moving the fds[1] assignment below, since `poll` is not using it yet.

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+	TEST_ASSERT_EQUAL_INT(POLLIN, fds[0].revents);
+
+	fds[2].fd = accept(fds[0].fd, NULL, NULL);
```
ziemleszcz, line 1061: Change to `accept(named, NULL, NULL)` for clarity.

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+	fds[2].events = POLLIN;
+	TEST_ASSERT(fds[2].fd > 0);
+
+	TEST_ASSERT_EQUAL_INT(1, poll(fds, 2, 1000));
```
ziemleszcz, line 1065: Please add a comment indicating that this is done to verify the completion of `connect`.

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+	TEST_ASSERT_EQUAL_INT(0, fds[0].revents);
+	TEST_ASSERT_EQUAL_INT(POLLOUT, fds[1].revents);
+
+	send_msg(fds[1].fd, 1, 0);
```
ziemleszcz, line 1069: Change to `send_msg(fd, 1,  0)` for clarity.

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+		TEST_ASSERT_EQUAL_INT(0, WEXITSTATUS(status));
+
+		close(conn);
+		close(named);
+	}
+	else {
+		if ((fd = socket(AF_UNIX, type, PF_UNSPEC)) < 0)
+			exit(1);
+
+		while (true) {
```
ziemleszcz, line 1133: Please set a reasonable limit for `connect` to complete here and below.

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+
+		fds[0].fd = named;
+		fds[0].events = POLLIN;
+
+		TEST_ASSERT_EQUAL_INT(1, poll(fds, 1, 500));
+		TEST_ASSERT_EQUAL_INT(POLLIN, fds[0].revents);
+
+		conn = accept(fds[0].fd, NULL, NULL);
+		TEST_ASSERT(conn > 0);
+		fds[1].fd = conn;
```
ziemleszcz, line 1172: It appears that `fds[1]` is not used anywhere.

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+		if ((named = unix_named_socket(type, socket_name)) < 0)
+			exit(1);
+
+		if (listen(named, 0) < 0)
+			exit(1);
+
+		usleep(500); /* sleep so that connect would block */
+
+		if ((conn = accept(named, NULL, NULL)) < 0)
+			exit(1);
```
ziemleszcz, line 1264: Use TEST_ASSERT?

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+		close(fd);
+
+		exit(0);
+	}
+
+	/* nonblocking connect, blocking accept */
+
+	pid = checked_fork();
+
+	if (pid) {
```
ziemleszcz, line 1210: Why are the server and client blocks reversed?

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+		if ((fd = socket(AF_UNIX, type, PF_UNSPEC)) < 0)
+			FAIL("socket");
+
+		if (set_nonblock(fd, 1) < 0)
+			FAIL("set_nonblock");
+
+		while (true) {
+			rv = connect_to_named(fd, socket_name);
+			if (rv < 0 && errno == EINPROGRESS) {
```
ziemleszcz, line 1219: What about EALREADY error?

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+
+	if (pid) {
+		if ((fd = socket(AF_UNIX, type, PF_UNSPEC)) < 0)
+			FAIL("socket");
+
+		if (set_nonblock(fd, 1) < 0)
+			FAIL("set_nonblock");
+
+		while (true) {
+			rv = connect_to_named(fd, socket_name);
```
ziemleszcz, line 1218: There is no break for `rv >= 0`.

libc/socket/unix-socket.c
adamgreloch: There's more stuff to assert in the server block and since asserts are invalid in the child procs these blocks had to be reversed. Otherwise the server would need to call `exit(1)` on failed assertions which isn't informative

libc/socket/unix-socket.c
adamgreloch: I check EALREADY behavior in `accept_connect_errnos` test. Here, EALREADY is not possible - connect will get called only once

libc/socket/unix-socket.c
adamgreloch: This case expects `rv >= 0` to never happen - see the usleep in the child proc just under the listen call. I will make this test to fail if the sleep is too short

libc/socket/unix-socket.c
adamgreloch: Added EISCONN check to async connection test

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,399 @@ TEST(test_unix_socket, poll)
...
+	if (socketpair(AF_UNIX, SOCK_STREAM, 0, fd) < 0)
+		FAIL("socketpair");
+
+	n = write(fd[0], data, msg_len);
+	TEST_ASSERT(n == msg_len);
+
+	n = write(fd[0], data, 1);
+	TEST_ASSERT(n == 1);
+
+	for (int i = 0; i < 4; i++) {
```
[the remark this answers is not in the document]
adamgreloch: Added a comment

libc/socket/unix-socket.c
adamgreloch: Corrected

libc/socket/unix-socket.c
```diff
@@ -904,6 +933,430 @@ TEST(test_unix_socket, poll)
...
+
+		if (set_nonblock(fd, 1) < 0)
+			FAIL("set_nonblock");
+
+		while (true) {
+			rv = connect_to_named(fd, socket_name);
+			if (rv >= 0) {
+				FAIL("should never happen - child proc should sleep for longer");
+			}
+			else if (rv < 0 && errno == EINPROGRESS) {
```
ziemleszcz, line 1239: Redundant check `rv < 0`.

libc/socket/unix-socket.c
```diff
@@ -904,6 +928,587 @@ TEST(test_unix_socket, poll)
...
+	TEST_ASSERT_EQUAL_INT(0, FD_ISSET(fd, &rfds));
+	TEST_ASSERT_LESS_THAN(350, ms);
+	TEST_ASSERT_GREATER_THAN(290, ms);
+
+	tv.tv_sec = 0;
+	tv.tv_usec = 1000;
+	FD_ZERO(&rfds);
+	FD_SET(fd, &rfds);
+	errno = 0;
+	rv = select(-1, &rfds, NULL, NULL, &tv);
```
ziemleszcz, line 1477: These are general tests to check if `select` correctly identifies invalid parameters. We should create a new `poll` directory and relocate the tests there, as they apply to polling various types of file descriptors.

libc/socket/unix-socket.c
```diff
@@ -904,6 +928,587 @@ TEST(test_unix_socket, poll)
...
+		close(fds[0].fd);
+	}
+	else {
+		if ((named = unix_named_socket(type, socket_name)) < 0)
+			exit(1);
+
+		if (listen(named, 0) < 0)
+			exit(1);
+
+		usleep(500); /* sleep so that connect would block */
```
ziemleszcz, line 1370: 500 µs is too short, the test occasionally fails. Increase the sleep time to at least 1 ms.

ziemleszcz (approved):

PR #366, opened 2024-08-19
trunner: add armv8r52-mps3an536-qemu target

adamdebek:
JIRA: CI-468

## Types of changes

- [] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

## Checklist:

## Special treatment


trunner/target/emulated.py
```diff
@@ -94,3 +94,15 @@ def __init__(self):
     @classmethod
     def from_context(cls, _: TestContext):
         return cls()
+
+
+class ARMv8R52_MPS3_AN536QemuTarget(QemuTarget):
```
mateusz-bloch, line 99: From what I can see, we don't use `_` in other target class names. I would stick to one consistent naming convention.

trunner/target/emulated.py
```diff
@@ -94,3 +94,15 @@ def __init__(self):
...
+    name = "armv8r52-mps3an536-qemu"
+    rootfs = False
+
```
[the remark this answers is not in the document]
mateusz-bloch: So, as I mentioned, all tests are currently being run, but the binaries are not present on the system.
image

mateusz-bloch (changes requested):
Also every exit currently ends with an `Abort`. Is issue was reported? If the fix might take some time, you could add a workaround to run tests earlier.

image

Edit:
There are also a few issues where the OS freezes after certain commands
image

lukileczo:
> Also every exit currently ends with an Abort. Is issue was reported? If the fix might take some time, you could add a workaround to run tests earlier.

@mateusz-bloch I'm aware of this issue, it's related to toolchain - I've found the root cause but now have no time to try to fix it. I will get back to it.

trunner/target/emulated.py
damianloew: @adamdebek please apply the suggestion

trunner/target/emulated.py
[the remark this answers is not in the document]
damianloew: I agree on the suggestion, we should set is as experimental for now and enable only psh tests (in yaml)

damianloew (changes requested):

damianloew (commented):
Please enable this target to CI in a separate commit

damianloew (commented):
The issue that was stopping these changes have been fixed, please prepare it to merge - apply the suggestions and run CI on this target on PR in phoenix-rtos-project

trunner/dut.py
```diff
@@ -202,6 +202,43 @@ def open(self):
...
+        termios.tcsetattr(STDOUT_FILENO, termios.TCSANOW, attr)
+
+    def open(self):
+        # qemu when using stdio serial enforces ONLCR termios flag (converting `\n` to `\r\n`)
+        # on proper guest OSes this results in `\r\r\n` which might be incorrectly interpreted by CI log viewers
+        # use custom preexec_fn to setup termios of child PTY
+        self.pexpect_proc = fdspawncustom(
+            *self.args, preexec_fn=self._set_termios_raw, **self.kwargs
+        )
         self._set_logfiles()
```
mateusz-bloch, line 242: Why don't inherit from QemuDut?
```suggestion
class ARMv8R52MPS3AN536_Dut(QemuDut):
    def open(self):
        # qemu when using stdio serial enforces ONLCR termios flag (converting `\n` to `\r\n`)
        # on proper guest OSes this results in `\r\r\n` which might be incorrectly interpreted by CI log viewers
        # use custom preexec_fn to setup termios of child PTY
        self.pexpect_proc = fdspawncustom(
            *self.args, preexec_fn=self._set_termios_raw, **self.kwargs
        )
        self._set_logfiles()
```

psh/test.yaml
```diff
@@ -11,12 +11,13 @@ test:
 
         - name: auth
           harness: test-auth.py
+
```
mateusz-bloch, line 14:
```suggestion
```

mateusz-bloch (approved):
LGTM

trunner/target/emulated.py
```diff
@@ -107,3 +107,18 @@ def __init__(self):
...
+class ARMv8R52MPS3AN536QemuTarget(QemuTarget):
+    name = "armv8r52-mps3an536-qemu"
+    rootfs = False
+    experimental = True
+
+    def __init__(self):
+        super().__init__("armv8r52-mps3an536-qemu.sh")
+        self.dut = ARMv8R52MPS3AN536_Dut(f"{self.project_dir}/scripts/{self.script}", encoding="utf-8")
+        self.rebooter = QemuDutRebooter(self.dut)
```
mateusz-bloch, line 120: Could you rebase the PR? So we could merge it.

```suggestion
```

damianloew: Yeah, let's close this topic

damianloew (approved):

PR #376, opened 2024-10-04
trunner: add multi possible device search for targets

maska989:
JIRA: CI-500

## Description
In replace target to new version, there was problem with new device ID and vendor ID.
That means we are forced to replace it in target variables, but we want also have the possibility to run older versions of the boards. That's why we want to utilize multiple possible vid/pid configuration to the same target.

## Types of changes

## How Has This Been Tested?

- [X] Tested by hand on: armv7m7-imxrt117x-evk.

## Checklist:

## Special treatment

### DIFF
--- a/trunner/target/base.py
+++ b/trunner/target/base.py
@@ -11,25 +11,42 @@ from trunner.dut import PortNotFound
 from trunner.types import TestOptions, TestResult
 
 
-def find_port(port_hint: str) -> str:
-    port = None
+def vid_pid_parser(vid_pid_string):
+    list = []
+    for vid_pid in vid_pid_string:
+        try:
+            vid_str, pid_str = vid_pid.split(":")
+            vid = int(vid_str, 16)
+            pid = int(pid_str, 16)
+            list.append((vid, pid))
+        except ValueError:
+            raise PortNotFound(f"Wrong format: {vid_pid}. Use 'VID:PID'.")
+    return list
 
-    for p in list_ports.grep(port_hint):
-        if port is not None:
-            raise PortNotFound(
-                "More than one port was found! Maybe more than one device is connected? Hint used to find port:"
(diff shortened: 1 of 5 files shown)


trunner/target/base.py
```diff
@@ -11,25 +11,42 @@
...
+def find_port(port_hint: list[str]) -> list:
+
+    all_ports = list_ports.comports()
+    vid_pid_list = vid_pid_parser(port_hint)
+    ports = []
+
+    for port in all_ports:
+        for vid, pid in vid_pid_list:
+            if port.vid == vid and port.pid == pid:
+                ports.append(port)
```
mateusz-bloch, line 36:
```suggestion
    vid_pid_list = vid_pid_parser(port_hint)
    ports = [port for port in list_ports.comports() if (port.vid, port.pid) in vid_pid_list]
```

damianloew: It seems to be simple, so I'd prefer to use this version

trunner/target/base.py
```diff
@@ -11,25 +11,42 @@
 from trunner.types import TestOptions, TestResult
 
 
-def find_port(port_hint: str) -> str:
-    port = None
-
-    for p in list_ports.grep(port_hint):
```
damianloew, line 17: After some thinking, we can simply pass a part of port name, which is correct for both older and newer board. For older the device string from `list_ports()` is:
```
/dev/ttyACM0 - DAPLink CMSIS-DAP - mbed Serial Port
```

for newer one:

```
/dev/ttyACM0 - MCU-LINK on-board (r0E2) CMSIS-DAP V2.250 - VCOM
```

So I propose looking for `CMSIS-DAP` for armv7m7 targets. Then all this additional code is not needed.

damianloew (changes requested):

PR #378, opened 2024-10-07
libc: add custom timeout discrepancy for emulated target in unix-sock…

adamdebek:
…et test

JIRA: CI-501

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


nalajcie:
maybe it would be better to pass custom timeout param as a commandline param if possible (that way we will be able to provide custom values for any particular target requiring it?)

mateusz-bloch:
image
nit: Commit title should be short, details can be described in the commit message body

libc/socket/unix-socket.c
```diff
@@ -930,6 +931,14 @@ void runner(void)
 
 int main(int argc, char *argv[])
 {
+	/* Custom value (ms) added to timeout in assert since blocking interval
+	 * significantly overruns this timeout due to big kernel scheduling delays on some targets.
+	 * Typical for emulated targets.
+	 */
+	if (argc == 2) {
```
nalajcie, line 938: IMHO would be better to add some option param to possibly not break `UnityMain` parsing `argv` parsing? (see how `UnityMain` parses options without the usage of `getopt`)

(eg. calling `test-libc-unix-socket --extra-poll-delay-ms 100` is more descriptive and needs much less comments)

libc/socket/unix-socket.c
```diff
@@ -930,6 +931,14 @@ void runner(void)
...
+		pollTimeoutDiff = atoi(argv[1]);
+	}
+
 	/* Assume /tmp dir is missing */
```
nalajcie, line 942: is this comment valid? the code looks like we assume initially that it's not missing? Also fix check `if (isMissing)` to be MISRA-compliant (or use `bool`)

libc/socket/unix-socket.c
```diff
@@ -36,6 +36,7 @@
 
 char data[DATA_SIZE];
 char buf[DATA_SIZE];
+int pollTimeoutDiff = 30;
```
nalajcie, line 39: `static`
other global values should probably also be `static` here?

nalajcie (commented):
some nitpicks, also :+1: for writing short/simple commit message titles

nalajcie:
image


image

libc/socket/unix-socket.c
```diff
@@ -859,7 +860,7 @@ void unix_poll(int type)
 	TEST_ASSERT(rv == 0);
 	TEST_ASSERT(fds[0].revents == 0);
 	TEST_ASSERT(fds[1].revents == 0);
-	TEST_ASSERT_LESS_THAN(350, ms);
+	TEST_ASSERT_LESS_THAN(300 + pollTimeoutDelay, ms);
```
adamgreloch, line 863: This causes the bound on `ms` to decrease by 20ms in the default case (`pollTimeoutDelay = 30`). Is this intended?

adamdebek: As I remember it was set 310 initialy but due to fails on emulated target it was increased to 350 and it worked out that it is still too low for zynq-qemu, so in this pr zynq-qemu gets its own custom value, and all others target will use default value of 30 ms, I think it will be sufficient.

adamgreloch: Alright, thanks

libc/socket/unix-socket.c
```diff
@@ -930,6 +931,16 @@ void runner(void)
 
 int main(int argc, char *argv[])
 {
+	/* Due to scheduling delays of host system on which emulator runs,
+	 * add some extra value to poll timeout checks
+	 */
+	for (int i = 1; i < argc; i++) {
+		if (strcmp(argv[i], "--extra-poll-delay-ms") == 0) {
+			pollTimeoutDelay = atoi(argv[i + 1]);
```
adamgreloch, line 939: Should `pollTimeoutDelay` ever be allowed to have negative value? If not, maybe it'd be good to have a simple argument value validation here?

adamdebek: It is passed directly in test.yaml, but I can add it.

libc/socket/unix-socket.c
```diff
@@ -1478,6 +1479,20 @@ void runner(void)
...
+			if (pollTimeoutDelay <= 0) {
+				fprintf(stderr, "--extra-poll-delay-ms argument is not positive integer\n");
+				exit(EXIT_FAILURE);
+			}
+			break;
+		}
+	}
+
 	/* Assume /tmp dir is missing */
 	int isMissing = 0;
```
damianloew, line 1497:
```suggestion
	bool wasTmpMissing = false;
```

Please see the comment above.

libc/socket/unix-socket.c
damianloew: The suggestion hasn't been fully taken into account. Still the comment seems to be wrong. I'd remove the comment and rename the variable name to make readable what is happening in here (`createTmpIfMissing` doesn't check the arg, it check's whether dir exists anyway and then sets the value).

damianloew (commented):
Please look at the latest comments and then we can merge it :)

damianloew (approved):

PR #379, opened 2024-10-14
libc: fix setvbuf_fullbuffer_overflow testcase

damianloew:
## Motivation and Context

The test case checks fully buffered mode. Previous implementation required data overflowing the buffer (sending 10 bytes when a buffer is 8 bytes long) to be written to a file at once which is not required by POSIX. The current solution forces flushing the remaining 2 bytes anyway

JIRA: CI-507

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic-qemu with https://github.com/phoenix-rtos/libphoenix/pull/376/commits changes applied, host-pc).

## Checklist:

## Special treatment

### DIFF
--- a/libc/stdio/stdio_file.c
+++ b/libc/stdio/stdio_file.c
@@ -980,7 +980,10 @@ TEST(stdio_bufs, setvbuf_fullbuffer_overflow)
 	TEST_ASSERT_EQUAL_INT(0, setvbuf(filep, buf2, _IOFBF, sizeof(buf2)));
 
 	TEST_ASSERT_GREATER_THAN_INT(0, fputs(data, filep));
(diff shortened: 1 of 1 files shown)


adamdebek (commented):
At this moment, this testcase is tailored to pass in glibc and doesn't check if buffering works correctly on phoenix. I would make this testcase phoenix specific.

ziemleszcz (approved):
LGTM

damianloew:
> pass in glibc

Yeah, but then we would test characteristics that are implementation-defined. In those tests we focus on satisfying POSIX requirements which only requires those 8 bytes to be written without flush. Also, it would be much simpler to merge those 'flexible' changes as we accept an older Phoenix implementation.

PR #381, opened 2024-10-16
test_mprotect: test if mprotected pages are copied

badochov:
pages_in_child_copied test checks if pages with protection changed to read from read-write by mprotect are copied by child.

JIRA: RTOS-953

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/608

### DIFF
--- a/mem/test_mprotect.c
+++ b/mem/test_mprotect.c
@@ -18,6 +18,7 @@
 #include <unistd.h>
 #include <errno.h>
 #include <stdlib.h>
+#include <sys/wait.h>
 
 #include "unity_fixture.h"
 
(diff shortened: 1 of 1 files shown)


badochov:
DEMO: https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1216

badochov:
DEMO: https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1216

badochov:
DRAFTING: as demo showed issue on ia32, weird that only on one platform there's an issue with this change

mem/test_mprotect.c
```diff
@@ -67,9 +68,42 @@ TEST(test_mprotect, test_mprotect_singlecore)
...
+	TEST_ASSERT_EQUAL_INT(0, mprotect(area, page_size, PROT_READ));
+
+	pid_t pid = fork();
+	TEST_ASSERT(pid >= 0);
+	if (pid == 0) {
+		/* Wait for modifications in parent. */
+		sleep(1);
+		if (area[0] != 0x42) {
+			exit(1);
+		}
```
agkaminski, line 88: Perhaps it might be easier to do this test in reverse - modify the memory in child instead. This would be nicer, as it would only require `waitpid` sync point, without `sleep`. But I'm not 100% such approach would test exactly the same thing

badochov: I'll add separate test case for that

PR #382, opened 2024-10-30
trunner:implement new type of reboot

maska989:
Adds possibility to reset target using jtag standard

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/harness/base.py
+++ b/trunner/harness/base.py
@@ -71,13 +71,22 @@ class FlashError(ProcessError):
 
 
 class Rebooter:
-    """Class that provides all necessary methods needed for rebooting target device."""
+    """
+    Class that provides all necessary methods needed for rebooting target device.
+
+    Possible ways to reboot targets:
+    _reboot_soft / _reboot_hard / _reboot_via_debugger
+    """
 
     def __init__(self, host: Host, dut: Dut):
         self.host = host
         self.dut = dut
 
     def _reboot_soft(self):
+        """
+        Resets DUT by pulling down NRST pin for a while. It needs the specific Rpi pin (declared in host.py)
+        to be connected to NRST on the evaluation board via a proper resistor.
+        """
         self.host.set_reset(0)
         time.sleep(0.10)
         self.dut.clear_buffer()
@@ -86,11 +95,22 @@ class Rebooter:
         time.sleep(0.25)
 
     def _reboot_hard(self):
+        """
+        Uses signal to cut down power from device and perform clean start of all components.
+        It needs the specific Rpi pin (declared in host.py) to be connected to relay on board power line.
+        """
         self.host.set_power(False)
         time.sleep(0.5)
         self.dut.clear_buffer()
         self.host.set_power(True)
         time.sleep(0.5)
+
+    def _reboot_by_debugger(self):
+        """
+        Uses debug interface to reboot device (Depends what resources are available on device).
+        To utilize end device must be connected to host device using debugger.
+        """
+        raise NotImplementedError
 
     def _reboot_dut_gpio(self, hard):
         if hard:
@@ -99,7 +119,7 @@ class Rebooter:
             self._reboot_soft()
 
     def _reboot_dut_text(self, hard):
-        pass
+        raise NotImplementedError
 
     def _set_flash_mode(self, flash):
         pass
--- a/trunner/target/armv7a9.py
+++ b/trunner/target/armv7a9.py
@@ -19,7 +19,7 @@ from trunner.harness import (
 )
 from trunner.harness import TerminalHarness
 from trunner.host import Host
-from trunner.tools import Phoenixd, OpenocdGdbServer, wait_for_vid_pid
+from trunner.tools import Phoenixd, OpenocdGdbServer, OpenocdProcess, wait_for_vid_pid
 from trunner.types import TestResult, TestOptions
 from .base import TargetBase, find_port
 
@@ -27,16 +27,45 @@ from .base import TargetBase, find_port
 class ARMv7A9TargetRebooter(Rebooter):
     # TODO add text mode reboot
 
+    def __call__(self, flash=False, hard=False):
+        """Sets flash mode and perform hard or soft & debugger reboot based on `hard` flag."""
+
+        self._set_flash_mode(flash)
+        if hard:
+            if self.host.has_gpio():
+                self._reboot_dut_gpio(hard=hard)
+            else:
+                self._reboot_dut_text(self)
+        else:
+            self._reboot_by_debugger()
+
+    def _reboot_by_debugger(self):
+        OpenocdProcess(
+            board="digilent_zedboard",
+            extra_args=[
+                "-c adapter srst pulse_width 250",
+                "-c adapter srst delay 250",
+                "-c adapter speed 10000",
+                "-c init; reset; continue",
+            ],
+        ).run()
+
     def _reboot_soft(self):
         self._reboot_hard()
 
     def _reboot_hard(self):
+
         self.host.set_power(False)
         # optimal power off time to prevent sustaining chips, e.g. flash memory, related to #540 issue
         time.sleep(0.75)
         self.dut.clear_buffer()
         self.host.set_power(True)
-        time.sleep(0.05)
+        # needed to make sure that smt2 debugger device has disappeared
+        time.sleep(0.5)
+        # after reboot, we need to wait for starting the smt2 device
+        wait_for_vid_pid(vid=0x0403, pid=0x6014, device="smt2", timeout=5)
+        # after power off it is necessary to wait more time to free all resources
+        time.sleep(2)
 
     def _set_flash_mode(self, flash):
         self.host.set_flash_mode(not flash)
@@ -52,9 +81,6 @@ class ZynqZedboardGdbPloLoader(TerminalHarness, PloInterface):
 
     def __call__(self):
         """Loads plo image to RAM using gdb."""
-
-        # after reboot we need to wait for smt2 device
-        wait_for_vid_pid(vid=0x0403, pid=0x6014, device="smt2", timeout=5)
 
         with self.gdbserver.run():
             try:
--- a/trunner/target/armv7m4.py
+++ b/trunner/target/armv7m4.py
@@ -18,15 +18,28 @@ from trunner.harness import (
     HarnessBuilder,
     FlashError,
 )
-from trunner.tools import GdbInteractive, OpenocdGdbServer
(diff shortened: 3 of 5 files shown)


trunner/harness/base.py
```diff
@@ -92,9 +92,14 @@ def _reboot_hard(self):
         self.host.set_power(True)
         time.sleep(0.5)
 
-    def _reboot_dut_gpio(self, hard):
+    def _reboot_via_jtag(self, ftdi):
+        pass
+
+    def _reboot_dut_gpio(self, hard, ftdi=False):
```
nalajcie, line 98: this is generic `Rebooter` class used for all targets. Please don't use `ftdi` term here.

Maybe we should describe (as docstrings) which methods of rebooting we do support and what is needed for them to work.

The third generic option to reboot you're adding is "rebooting by debugger" (regardless of the debug protocol (jtag/swd) or debugger device (jlink/mcu-link/ftdi-based/lauterbach, etc.))

trunner/target/armv7a9.py
```diff
@@ -72,7 +84,9 @@ def __call__(self):
             except subprocess.CalledProcessError as e:
                 raise FlashError(msg=str(e), output=e.stdout) from e
             except subprocess.TimeoutExpired as e:
-                raise FlashError(msg=str(e), output=e.stdout.decode("ascii") if e.stdout else None) from e
+                raise FlashError(
```
nalajcie, line 87: invalid change, please setup your linter correctly (line length: 120)

trunner/harness/base.py
```diff
@@ -71,7 +71,21 @@ class FlashError(ProcessError):
...
+
+    Possible ways to reboot targets:
+
+    _reboot_soft() -> Uses signal to reset device using build in reset button/function.
+                      To utilize this type of reboot use pin declared in host.py
+
+    _reboot_hard() -> Uses signal to cut down power from device and perform clean start of all components.
+                      To utilize use pin declared in host.py
+
+    _reboot_via_debugger() -> Uses debug interface to reboot device (Depends what resources are available on device).
```
nalajcie, line 85: please describe functions as docstrings to these function

trunner/harness/base.py
```diff
@@ -92,9 +106,14 @@ def _reboot_hard(self):
         self.host.set_power(True)
         time.sleep(0.5)
 
-    def _reboot_dut_gpio(self, hard):
+    def _reboot_via_debugger(self, ftdi):
```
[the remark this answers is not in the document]
nalajcie: why the soft reboot will always need to be done by the debugger?

trunner/harness/base.py
```diff
@@ -92,9 +106,14 @@ def _reboot_hard(self):
...
+        pass
+
+    def _reboot_dut_gpio(self, hard, gdb_dev_rst: Optional[bool] = False):
```
nalajcie, line 112: note: the type is not `Optional[bool]`', just `bool` (`Optional` means that you can pass None, not an optional function parameter)

trunner/harness/base.py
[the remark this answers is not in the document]
damianloew: On targets where debugger is available I'd stick to this approach to unify the environment

trunner/target/armv7a9.py
```diff
@@ -19,26 +19,59 @@
...
+
+        self._set_flash_mode(flash)
+
+        if hard:
+            if self.host.has_gpio():
+                self._reboot_dut_gpio(hard=hard)
+            else:
+                self._reboot_by_debugger()
+        else:
+            self._reboot_by_debugger()
```
mateusz-bloch, line 41: What about the` _reboot_dut_text() `?

I know we don’t have it implemented yet, but we want to have it

mateusz-bloch, line 41:
```suggestion
        if hard: and  self.host.has_gpio():
            self._reboot_dut_gpio(hard=hard)
        else:
            self._reboot_by_debugger()
```

trunner/harness/base.py
```diff
@@ -86,12 +95,23 @@ def _reboot_soft(self):
...
         self.dut.clear_buffer()
         self.host.set_power(True)
         time.sleep(0.5)
 
+    def _reboot_by_debugger(self):
+        """
+        Uses debug interface to reboot device (Depends what resources are available on device).
+        To utilize end device must be connected to host device using debugger.
+        """
+        pass
```
nalajcie, line 113: `raise NotImplementedError` ?

trunner/target/armv7a9.py
```diff
@@ -19,26 +19,56 @@
...
 
+    def __call__(self, flash=False, hard=False):
+        """Sets flash mode and perform hard or soft & debugger reboot based on `hard` flag."""
+
+        self._set_flash_mode(flash)
+
+        if hard and self.host.has_gpio():
+            self._reboot_dut_gpio(hard=hard)
+        else:
+            self._reboot_by_debugger()
```
mateusz-bloch, line 38: In such a situation with hard reboot, could we provide text instructions for the user on how to proceed with the campaign manually. If they don’t want to set up the Raspberry Pi?

mateusz-bloch (changes requested):

trunner/target/armv7m4.py
```diff
@@ -18,15 +18,35 @@
...
+
+    def _reboot_by_debugger(self):
+        # after flashing device is necessary to wait bit of time to free all resources
+        Openocd(
+            self,
+            interface="stlink",
+            target="stm32l4x",
+            extra_args=["-c", "reset_config srst_only srst_nogate connect_assert_srst", "-c init; reset; exit"],
+        )
+        time.sleep(0.5)
```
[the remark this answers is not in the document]
maska989: Depending on the target.

trunner/target/armv7a9.py
maska989: That's why we check self.host.has_gpio to utilize in this case reboot by debugger.

trunner/target/armv7m4.py
```diff
@@ -18,15 +18,34 @@
...
+    def __call__(self, flash=False, hard=False):
+        """Sets flash mode and perform hard or soft & debugger reboot based on `hard` flag."""
+
+        if hard:
+            if self.host.has_gpio():
+                self._reboot_dut_gpio(hard=hard)
+            else:
+                self._reboot_by_debugger()
+        else:
+            self._reboot_by_debugger()
```
mateusz-bloch, line 39:
```suggestion
        if hard and self.host.has_gpio():
            self._reboot_dut_gpio(hard=hard)
        else:
            self._reboot_by_debugger()
```

trunner/target/armv7a9.py
mateusz-bloch: As far as I know, it currently works correctly on the nucleo. However, on the zedboard, a manual hard reboot is required.  I would suggest either adding a TODO  or something like:
```suggestion
        if hard:
        	if self.host.has_gpio():
            		self._reboot_dut_gpio(hard=hard)
            	else:
            		print("Do a hard reboot.")
            		answer = input("If you perform reboot write Y")[0]
            		if answer != "Y":
            			raise Excpetion
        else:
            self._reboot_by_debugger()
```

mateusz-bloch (dismissed):
LGTM,  only few comments

mateusz-bloch (commented):
Also, i would change commit name to be more specific

image

maybe smth like:

`trunner: implement reboot  by device debugger`

trunner/target/armv7a9.py
damianloew: I am not 100% sure whether this hard reboot with power off is necessary after changing boot mode. Running the hw targets using a laptop host will be one step closer after this change, but it's still not supported on armv7a9 target, because changing boot modes is missing. I'd suggest sticking to the current solution.

mateusz-bloch: Understand. I would still consider an exception indicating that this is not supported

trunner/tools/gdb.py
```diff
@@ -151,6 +151,45 @@ def _close(self):
         self.logfile.close()
 
 
+def OpenocdProcess(
```
damianloew, line 154: TBH my intention was creating another class for simply calling openocd commands without a need for `with`, but the only difference is removing yield from a run or simply __call__ method in our new class. Probably we can create a parent class `Openocd` and making the `OpenocdGdbServer` a child class. @mateusz-bloch could you guide @maska989 how we can do it?

damianloew: Please also note that we want to use this new class in the flashing process of stm target.

trunner/target/armv7a9.py
```diff
@@ -19,26 +19,55 @@
...
+        OpenocdProcess(
+            board="digilent_zedboard",
+            extra_args=[
+                "-c adapter srst pulse_width 250",
+                "-c adapter srst delay 250",
+                "-c adapter speed 10000",
+                "-c init; reset; continue",
+            ],
+        )
+        time.sleep(0.5)
```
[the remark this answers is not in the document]
damianloew: From a quick look on target, that's not needed

trunner/tools/gdb.py
```diff
@@ -151,6 +147,34 @@ def _close(self):
...
 
 
+class OpenocdGdbServer(OpenocdProcess):
+    """Handler for OpenocdGdbServer process"""
+
+    @contextmanager
+    @add_output_to_exception(OpenocdError)
+    def run(self):
+        try:
+            # Use pexpect.spawn to run a process as PTY, so it will flush on a new line
```
damianloew, line 157: As you can see the whole

```
            if self.board:
                args = ["-f", f"board/{self.board}.cfg"]
            else:
                args = ["-f", f"interface/{self.interface}.cfg", "-f", f"target/{self.target}.cfg"]

            if self.extra_args:
                args.extend(self.extra_args)

            self.proc = pexpect.spawn("openocd", args, encoding="ascii", logfile=self.logfile)

            try:
                self.proc.expect_exact("Info : Listening on port 3333 for gdb connections")
            except (pexpect.TIMEOUT, pexpect.EOF) as e:
                raise OpenocdError("Failed to start gdb server", self.logfile.getvalue()) from e
```

part is common for both the classes, that's why we can create the additional method, for example called `_start()` which will do this part and then it can be used in:

 - `__call__()` in OpenocdProcess
 - `run()` in `OpenocdGdbServer`

! Important note
Please remember that you have to overwrite the `__call__` method to pass in `OpenocdGdbServer` to avoid executing the code from parent class. For any additional explanations please contact @mateusz-bloch

damianloew (changes requested):

damianloew (approved):

PR #385, opened 2024-11-28
 trunner: improve formatting of multi-line subresult messages 

adamdebek:
JIRA: CI-519

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/busybox/busybox.py
+++ b/busybox/busybox.py
@@ -50,14 +50,12 @@ def harness(dut: Dut, ctx: TestContext, result: TestResult):
             continue
 
         if idx == 3:
-            # "\n\t\t" used to create more readable multiline message when printed
-            line_msg = "\n\t\t" + parsed["line"]
             # append extra message to the last test (if available)
             if subresult and subresult.status == Status.FAIL:
-                subresult.msg += line_msg
+                subresult.msg += parsed["line"]
             else:
                 # message may also appear during the test
-                msg.append(line_msg)
+                msg.append(parsed["line"])
 
         if idx == 1:
             status = Status.from_str(parsed["status"])
--- a/trunner/types.py
+++ b/trunner/types.py
@@ -342,10 +342,29 @@ class TestSubResult(TestResult):
         self.set_stage(TestStage.RUN)
 
     def __str__(self) -> str:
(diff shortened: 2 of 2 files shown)


mateusz-bloch (commented):
Overall, it's a very good idea. 

I'm just thinking if we should go with a fixed spacing , it might look better, in my opinion. What do you think @damianloew  ?

nalajcie:
Why do we need this? 

IMHO no offset is OK (as subresults are indented, so everything with "negative indent" is clearly distinguishable)?

Using `lstrip` here is invalid, any output which aligns message by hand will be distorted (eg. gcc output highlighting the error, python traceback, etc.)

adamdebek:
I agree with removal of `lstrip()`

Example output with this change:
image

Output without this change:
image

nalajcie:
Thanks for the visual example. Some notes:
- the trailing newline in every `msg` seems like a problem in that particular harness
- I'm a little bit afraid about aligning it as the (sub)test names might be long - eg. mbedTLS subtests can have len up to 70 chars - so aligning the output might actually decrease the readability in that case (might break long messages)?

If the above is a non-issue, we can merge it.

nalajcie:
> Yes, trailing new lines in 1 line message is from harness, but this harness took into account that pr and `splitlines()` trim them.

Nah, this PR handles only writing to screen case, it will still look ugly in XML output. Please correct the msg on harness side if possible.

> Maybe when (subtest name + longest fail message from msg lines) < some_edge_value. Example abstract output: \t very-long-testnameeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee: FAIL \t\t message line 1 \t\t message line 2 ... \t\t message line n

IMHO we should always present the result in the same way - otherwise the user might get confused. 

I'm ok with custom per-test indentation if we acknowledge the potential drawbacks. I also don't have a problem with not-indenting multiline messages (as it's done for normal test results). Feel free to choose the solution you think works best.

adamdebek:
I think it is still clear to the user when FAIL status and some indented text below is presented and IMHO it is the most readable

damianloew (approved):
LGTM, thanks for making it a bit cleaner + leaving some description in the commit

PR #386, opened 2025-01-13
trunner: add missing expect during flash procedure

maska989:
## Description

This change resolves the issue where the flashing process executes a reboot after successful flashing, causing the next test case to receive output from the flash reboot and detect the PSH prompt too early

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/test_runner.py
+++ b/trunner/test_runner.py
@@ -244,6 +244,10 @@ class TestRunner:
         # Ensure first test will start with reboot
         last_test_failed = True
 
+        # Unload data before starting tests
+        if self.ctx.should_flash is not False and self.ctx.target.experimental is False:
+            self.target.dut.expect(r".+")
+
         for test in tests:
             # By default we don't want to reboot the entire device to speed up the test execution)
             # if not explicitly required by the test.


nalajcie:
hmm, I think it's invalid to assume every device will boot to `psh` prompt. 

If I understand correctly - the underlying issue is that we're rebooting twice between reboot and first test - so we should focus on removing the underlying issue instead :).

maska989:
> hmm, I think it's invalid to assume every device will boot to `psh` prompt.
> 
> If I understand correctly - the underlying issue is that we're rebooting twice between reboot and first test - so we should focus on removing the underlying issue instead :).

I don't think this is an issue, every target after flashing is rebooted to ensure that process ended correctly (Right now we know only about correct pass of data from host device, and nothing about that system start or not after flash) + 70% of our physical devices are using changing mode after flash and need to be shut down using power cut, without this it will hang on reboot without effect. This change will pinpoint this behavior in the future and separate flashing process from tests cases.
In addition, we're using often different reboots in test cases and flash procedure, removing one of reboot will clog test campaign (using --no-flash flag or using environment with old setting from previous campaign) or flash (lack of state change on device).

But it is my personal opinion, and I'm open to discussion about improvements :)

nalajcie:
> > hmm, I think it's invalid to assume every device will boot to `psh` prompt.
> > If I understand correctly - the underlying issue is that we're rebooting twice between reboot and first test - so we should focus on removing the underlying issue instead :).
> 
> I don't think this is an issue (ash only used in QEMU right now), every target after flashing is rebooted to ensure that process ended correctly (Right now we know only about correct pass of data from host device, and nothing about that system start or not after flash) + 70% of our physical devices are using changing mode after flash and need to be shut down using power cut, without this it will hang on reboot without effect. This change will pinpoint this behavior in the future and separate flashing process from tests cases. In addition, we're using often different reboots in test cases and flash procedure, removing one of reboot will clog test campaign (using --no-flash flag or using environment with old setting from previous campaign) or flash (lack of state change on device).
> 
> But it is my personal opinion, and I'm open to discussion about improvements :)

Some more remarks:
 - flashing procedure correctness is checked by inspecting the logs while flashing
 - whether the device would boot after the flashing procedure or not is not an indicator of flashing failure (flashing might be ok, firmware might be broken) - so IMHO it should be visible in other test result than `flash`
 - we can't universally assume that every device with `plo` would boot to `psh` shell - we have this information on per-target basis encoded as harness chain - so **the proposed change breaks the architecture of `trunner`** - that's why I'm **against** merging it in current form
 - we should either optimize the second reboot (in a graceful way - handling special cases as `--no-flash` flag, etc.) or eg. drain the currently buffered data before next test execution (this is currently done only after each test) or while rebooting, etc.

maska989:
> > > hmm, I think it's invalid to assume every device will boot to `psh` prompt.
> > > If I understand correctly - the underlying issue is that we're rebooting twice between reboot and first test - so we should focus on removing the underlying issue instead :).
> > 
> > 
> > I don't think this is an issue (ash only used in QEMU right now), every target after flashing is rebooted to ensure that process ended correctly (Right now we know only about correct pass of data from host device, and nothing about that system start or not after flash) + 70% of our physical devices are using changing mode after flash and need to be shut down using power cut, without this it will hang on reboot without effect. This change will pinpoint this behavior in the future and separate flashing process from tests cases. In addition, we're using often different reboots in test cases and flash procedure, removing one of reboot will clog test campaign (using --no-flash flag or using environment with old setting from previous campaign) or flash (lack of state change on device).
> > But it is my personal opinion, and I'm open to discussion about improvements :)
> 
> Some more remarks:
> 
>     * flashing procedure correctness is checked by inspecting the logs while flashing
> 
>     * whether the device would boot after the flashing procedure or not is not an indicator of flashing failure (flashing might be ok, firmware might be broken) - so IMHO it should be visible in other test result than `flash`
> 
>     * we can't universally assume that every device with `plo` would boot to `psh` shell - we have this information on per-target basis encoded as harness chain - so **the proposed change breaks the architecture of `trunner`** - that's why I'm **against** merging it in current form
> 
>     * we should either optimize the second reboot (in a graceful way - handling special cases as `--no-flash` flag, etc.) or eg. drain the currently buffered data before next test execution (this is currently done only after each test) or while rebooting, etc.

You're absolutely right, after rethink this approach is destructive for trunner, I will consider soften up this second reboot with some buffer changes to nullify this behavior of double reboot situation. 

Thanks for visiting this draft pr and give me some advices :)

trunner/test_runner.py
```diff
@@ -244,6 +244,10 @@ def run_tests(self, tests: Sequence[TestOptions]) -> Sequence[TestResult]:
         # Ensure first test will start with reboot
         last_test_failed = True
 
+        # Unload data before starting tests
+        if self.ctx.should_flash is not False and self.ctx.target.experimental is False:
```
nalajcie, line 248: I don't understand why it's disabled for experimental targets.

Can't we just always drain the buffer (without any IF?) - the same way as below:
`self.target.dut.read(timeout=0.1)`

Please note that this is still a race - new data might be produced between reading the buffer and issuing reboot before the first test.

PR #387, opened 2025-01-15
trunner: armv7a9: add missing clear_buffer

maska989:
JIRA: CI-521

## Description
Adds omitted clear_buffer to prevent caching output from previous step in to another one. This situation occurs only on start of campaign when step `Flash` is present. Excessive output is connected with changing state of device from loading image to launching it. 

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/target/armv7a9.py
+++ b/trunner/target/armv7a9.py
@@ -50,6 +50,9 @@ class ARMv7A9TargetRebooter(Rebooter):
             ],
         ).run()
 
+        # TODO: resolve race condition with clearing first logs after reset
+        self.dut.clear_buffer()
+
     def _reboot_soft(self):
         self._reboot_hard()
 


damianloew (commented):
I suppose that errors like this one are no longer visible thanks to this change?
https://github.com/phoenix-rtos/phoenix-rtos-project/actions/runs/12721408032/job/35464329095

nalajcie:
there is still a slight race condition - ideally the buffer should be emptied under reset (the current approach means we could remove first logs from the new run). 

Probably this is not doable without changing how openocd is being called right now, so probably a comment with fixme/todo would be enough?

damianloew:
> there is still a slight race condition - ideally the buffer should be emptied under reset (the current approach means we could remove first logs from the new run).
> 
> Probably this is not doable without changing how openocd is being called right now, so probably a comment with fixme/todo would be enough?

Yeah, good to remember, we would like to leave a TODO for now (to eliminate potential CI fails) as @maska989 started working on separating outputs from flashing step and first test (currently streaming logs from the reboot after flashing are visible in the first test section).

damianloew (approved):
LGTM

PR #388, opened 2025-01-21
libc/scanf-advanced: fix type usage

lukileczo:
## Description

`hhx` format specifier was used with `int` type in test `modifiers_mix,` which is incorrect. This caused the test to fail on big endian architectures (SPARC), as the value was written to the wrong address - the most significant byte. Type was changed to `unsigned char` to be compatible with the specifier.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `sparcv8leon-generic-qemu`

## Checklist:

## Special treatment

### DIFF
--- a/libc/scanf-advanced/stdio_scanf_advanced.c
+++ b/libc/scanf-advanced/stdio_scanf_advanced.c
@@ -1559,22 +1559,23 @@ TEST_TEAR_DOWN(stdio_scanf_rest)
 
 TEST(stdio_scanf_rest, modifiers_mix)
 {
-	int int1, int2, int3, int4, int5;
+	int int1, int2, int3, int4;
 	unsigned int res2;
 	float flt1, flt2, flt3;
 	long long int llint1;
 	char char1;
+	unsigned char uchar1;
 	ptrdiff_t ptr1;
 	char strTab[4][MAX_TESTSTR_WORDLEN] = { 0 };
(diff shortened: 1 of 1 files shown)


libc/scanf-advanced/stdio_scanf_advanced.c
```diff
@@ -1559,11 +1559,12 @@ TEST_TEAR_DOWN(stdio_scanf_rest)
...
-	int int1, int2, int3, int4, int5;
+	int int1, int2, int3, int4;
 	unsigned int res2;
 	float flt1, flt2, flt3;
 	long long int llint1;
 	char char1;
+	unsigned char uchar1;
 	ptrdiff_t ptr1;
 	char strTab[4][MAX_TESTSTR_WORDLEN] = { 0 };
 	const char *str = "~~1`2l 0.1!_2@lorem#0x233$ 1.2e-5 % nowy 1.200020e-5 nal^ 132 *{}:|?><[]',./5/123456/+123456-a(loremipsum\0)";
```
adamdebek, line 1570: The part of string that %hhx tries to match should be changed to fit `unsigned char` range (0-255), now it reads `123456` as hex value and cut it to 8 bit, which works, but as `POSIX` states it is undefined behaviour.

[_(...) if the result of the conversion cannot be represented in the space provided, the behavior is undefined._](https://pubs.opengroup.org/onlinepubs/9799919799/)

lukileczo: Done

adamdebek (approved):

PR #399, opened 2025-06-18
trunner: add aarch64a53-zynqmp-qemu target

KArkadiusz:
JIRA: CI-550

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/test.yaml
+++ b/libc/test.yaml
@@ -76,13 +76,18 @@ test:
     - name: unix-socket
       execute: test-libc-unix-socket
       targets:
-        exclude: [armv7a9-zynq7000-qemu]
+        exclude: [armv7a9-zynq7000-qemu, aarch64a53-zynqmp-qemu]
         include: [host-generic-pc]
 
     - name: unix-socket
       execute: test-libc-unix-socket --extra-poll-delay-ms 150
       targets:
         value: [armv7a9-zynq7000-qemu]
+    
+    - name: unix-socket # Issue: 1342 | https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1342
+      execute: test-libc-unix-socket --extra-poll-delay-ms 150 --transfer-loop-cnt 30
+      targets:
+        value: [aarch64a53-zynqmp-qemu]
 
     - name: inet-socket
       execute: test-libc-inet-socket
--- a/libc/socket/unix-socket.c
+++ b/libc/socket/unix-socket.c
@@ -60,6 +60,7 @@
 static char data[DATA_SIZE];
 static char buf[DATA_SIZE];
 static int pollTimeoutDelay = 30;
+static int transferLoopCnt = TRANSFER_LOOP_CNT;
 
 
 static ssize_t unix_named_socket(int type, const char *name)
@@ -692,7 +693,7 @@ TEST(test_unix_socket, transfer)
 {
 	unsigned int i;
 
-	for (i = 0; i < TRANSFER_LOOP_CNT; ++i) {
+	for (i = 0; i < transferLoopCnt; ++i) {
 		unix_transfer(SOCK_STREAM);
 		unix_transfer(SOCK_DGRAM);
 	}
@@ -1520,7 +1521,13 @@ int main(int argc, char *argv[])
 				fprintf(stderr, "--extra-poll-delay-ms argument is not positive integer\n");
 				exit(EXIT_FAILURE);
 			}
-			break;
+		}
+		if (strcmp(argv[i], "--transfer-loop-cnt") == 0) {
+			transferLoopCnt = atoi(argv[i + 1]);
+			if (transferLoopCnt < 0 || transferLoopCnt > 50) {
+				fprintf(stderr, "--transfer-loop-cnt shall be between 0 and 50\n");
+				exit(EXIT_FAILURE);
+			}
 		}
 	}
 


nalajcie:
PR template description not filled, converting to draft. Please do so on your own in the future.

mateusz-bloch (changes requested):
Please also check the issues occurring in CI:
- If the  cause is in the test itself or in our test runner, fix it.
- If the problem is from the new architecture, report it and skip the test or test case.

image
image

trunner/target/emulated.py
```diff
@@ -107,3 +107,16 @@ def __init__(self):
     @classmethod
     def from_context(cls, _: TestContext):
         return cls()
+
+
+class AARCH64a53ZynqmpQemuTarget(QemuTarget):
```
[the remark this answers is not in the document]
damianloew: Yeah, that should be the name

.github/workflows/ci.yml
```diff
@@ -14,5 +14,5 @@ on:
 
 jobs:
   call-ci:
-    uses: phoenix-rtos/phoenix-rtos-project/.github/workflows/ci-submodule.yml@master
+    uses: phoenix-rtos/phoenix-rtos-project/.github/workflows/ci-submodule.yml@KArkadiusz/aarch64_ci_test
```
damianloew, line 17: We rather do such temporary changes in the separate commit, marking that they won't be added to the project and it's only to check the result on CI before merge.

trunner/target/emulated.py
```diff
@@ -107,3 +107,16 @@ def __init__(self):
...
         return cls()
+
+
+class AARCH64A53ZynqmpQemuTarget(QemuTarget):
+    name = "aarch64a53-zynqmp-qemu"
+    rootfs = True
+
+    def __init__(self):
+        super().__init__("aarch64a53-zynqmp-qemu.sh")
+        self.prompt_timeout = 30
```
mateusz-bloch, line 118: If I remember correctly, 30 seconds is the default timeout in pexpect, but you can check to be sure.

libc/printf/stdio_printf.c
```diff
@@ -2036,7 +2036,9 @@ TEST(stdio_printf_cspn, S)
 
 TEST(stdio_printf_cspn, p)
 {
-#ifdef __phoenix__
+#if defined(__phoenix__) && defined(__CPU_ZYNQMP)
```
damianloew, line 2039: It should be done in a separate PR - if it's the temporary change needed to see the results from CI - it should be placed in the temporary commit within this PR (but then you should just check the results and after that move it to the valid PR).

nalajcie: don't depend on `__CPU`, this should be true for all 64-bit architectures.

libc/printf/stdio_printf.c
```diff
@@ -2036,8 +2036,10 @@ TEST(stdio_printf_cspn, S)
 
 TEST(stdio_printf_cspn, p)
 {
-#ifdef __phoenix__
-	char *expect = "deadbeef    (nil) 7fffffff 80000000";
+#if defined(__phoenix__) && defined(__aarch64__)
+	char *expect = "00000000deadbeef            (nil) 7fffffffffffffff 8000000000000000";
+#elif defined(__phoenix__)
+	char *expect = "0xdeadbeef (nil) 0x7fffffffffffffff 0x8000000000000000";
```
mateusz-bloch, line 2042:
```suggestion
#elif defined(__phoenix__)
	char *expect = "deadbeef    (nil) 7fffffff 80000000";
```

libc/test.yaml
```diff
@@ -76,13 +76,18 @@ test:
...
+        exclude: [armv7a9-zynq7000-qemu, aarch64a53-zynqmp-qemu]
         include: [host-generic-pc]
 
     - name: unix-socket
       execute: test-libc-unix-socket --extra-poll-delay-ms 150
       targets:
         value: [armv7a9-zynq7000-qemu]
+    
+    - name: unix-socket
+      execute: test-libc-unix-socket --extra-poll-delay-ms 150 --transfer_loop_cnt 10
```
damianloew, line 88: As you mentioned there is a significant change in execution times between `aarch64a53-zynqmp-qemu` and `armv7a9-zynq7000-qemu` so we should raise the issue about it.

Providing you add the link to the issue here it will be reasonable to use custom argument. However I see that that this test is still failing in CI - please resolve it + raise the issue and then we can merge those changes

libc/test.yaml
```diff
@@ -76,13 +76,18 @@ test:
...
+        exclude: [armv7a9-zynq7000-qemu, aarch64a53-zynqmp-qemu]
         include: [host-generic-pc]
 
     - name: unix-socket
       execute: test-libc-unix-socket --extra-poll-delay-ms 150
       targets:
         value: [armv7a9-zynq7000-qemu]
+    
+    - name: unix-socket # Issue: 1342 | https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1342
+      execute: test-libc-unix-socket --extra-poll-delay-ms 150 --transfer-loop-cnt 10
```
mateusz-bloch, line 88: From what I can see, everything looks fine overall. You can remove commit: `temporary: run aarch64 tests on ci` and we could go ahead with the merge.

Also I'm think if it's worth to set it to around 30 loops, since this test case currently takes 8 seconds to run.
```suggestion
      execute: test-libc-unix-socket --extra-poll-delay-ms 150 --transfer-loop-cnt 30
```
image

damianloew (commented):
Please fix the commit message before merge:

 - `Fix zynqmp false negative testcases`

According to our golden rules (Confluence -> CI -> Golden rules), broken rules:

 - atomic changes - we should describe shortly what have we changed in the specific tests to show what fixes exactly have been introduced. So let's split it to `libc/printf: handle %p on 64bit Phoenix platforms` and `libc/socket: decrease transferLoopCnt for aarch64a53-zynqmp-qemu`
 - commit message prefixed with module - as mentioned above - we want at least a prefix with a directory name where changes have been applied.
 - describe WHY: when it's obvious like in adding a new target to test runner we can skip that, I think `libc/printf: handle %p on 64bit Phoenix platforms` also does not need additional explanations, but commits where we are adjusting the test for a platform needs leaving WHY (theoretically all of the tests should be the same on all of the targets - every time this rule is broken we should explain WHY are we doing it)

damianloew (commented):
Also please remember to fill the  `All new and existing linter checks and tests passed.` checkbox - you can see the linter checks results in `Checks` tab

PR #402, opened 2025-07-21
trunner: escape control characters in results xml

wmlynik:
Control characters are not valid in xml, they need to be escaped in output of tests, otherwise the file can't be parsed.

JIRA: CI-563

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic-qemu.

## Checklist:

## Special treatment

### DIFF
--- a/trunner/text.py
+++ b/trunner/text.py
@@ -1,9 +1,9 @@
 import itertools
 import re
-from colorama import Style, Fore
 
+from colorama import Fore, Style
 
-ANSI_ESCAPE = re.compile(r'(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]')
+ANSI_ESCAPE = re.compile(r"(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]")
 
 
 def bold(s: str) -> str:
@@ -31,7 +31,7 @@ def magenta(s: str) -> str:
 
 
 def remove_ansi_sequences(line: str) -> str:
-    return ANSI_ESCAPE.sub('', line)
+    return ANSI_ESCAPE.sub("", line)
 
 
 def escape_invalid_xml_characters(s: str) -> str:
--- a/trunner/types.py
+++ b/trunner/types.py
@@ -1,17 +1,18 @@
 from __future__ import annotations
-from datetime import datetime
-from functools import total_ordering
 
 import os
 import re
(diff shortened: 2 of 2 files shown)


trunner/text.py
```diff
@@ -30,4 +31,30 @@ def magenta(s: str) -> str:
 
 
 def remove_ansi_sequences(line: str) -> str:
-    return ANSI_ESCAPE.sub('', line)
+    return ANSI_ESCAPE.sub("", line)
+
+
+def escape_invalid_characters(s: str) -> str:
```
nalajcie, line 37:
```suggestion
def escape_invalid_xml_characters(s: str) -> str:
```

trunner/types.py
```diff
@@ -1,17 +1,18 @@
...
 import re
 import sys
 import time
 import traceback
-import junitparser
 from dataclasses import dataclass, field
+from datetime import datetime
 from enum import Enum, auto
-from typing import Callable, Dict, List, Optional
+from functools import total_ordering
```
damianloew, line 11: nitpick - it shouldn't be the part of fix formatting commit

trunner/text.py
```diff
@@ -30,4 +31,30 @@ def magenta(s: str) -> str:
...
+        }
+
+        for i in itertools.chain(
+            range(0x08 + 1), range(0x0B, 0x0C + 1), range(0x0E, 0x1F + 1),
+            range(0x7F, 0x84 + 1), range(0x86, 0x9F + 1)
+        ):
+            if i not in translation_map:
+                translation_map[i] = f"\\x{i:02x}"
+
+        translation_map[127] = r"\x7f"
```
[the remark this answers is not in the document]
nalajcie: @gemini-code-assist the ranges are taken directly from the XML 1.0 specification - so it's good to have them explicitly stated as-is. You may suggest other more pythonic solution while preserving the original ranges

nalajcie, line 56: this is not a regex, why the `r` prefix?

nalajcie: (also - already included in the `for` as gemini found out)

nalajcie:
please also rebase onto current master

trunner/text.py
nalajcie: why is this not removed?

wmlynik: now it's removed

PR #409, opened 2025-07-31
Fix and improve meterfs tests

anglov:
## Description

Fix building on phoenix
Remove obsolete FIXME
Improvements:
- Change phoenix abstration to call directly metefs instead of lookup.
    Many flashsrv just create dev for meterfs instaed of mounting it. This
    disallows lookups for files via lookup syscall. As tests is designed to
    test meterfs, not flashsrv implementation, it's desirable to call
    directly to meterfs filesystem, at least until mounting will be fully
    suported in our implementations.
- Returning ID from lookup
    It's needed for better test coverage.
- Make checks in phoenix and pc abstraction similar
    Host abstraction checks for id > INT_MAX. Added similar checks in
    phoenix abstration. Also, change return to errno and debug assert
- Add missing oid initialisation

## Motivation and Context

This commit fix building tests on pc and improve possibility to run those tests on hardware platforms

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: host-generic-pc, EBRO

## Checklist:

## Special treatment

### DIFF
--- a/meterfs/file_pc.c
+++ b/meterfs/file_pc.c
@@ -13,6 +13,7 @@
 #include <limits.h>
 #include <host-flashsrv.h>
 #include <meterfs.h>
+#include <unity.h>
 
 #include "file.h"
 
@@ -22,28 +23,34 @@
 int file_lookup(const char *name)
 {
 	id_t id;
-
-	return hostflashsrv_lookup(name, &id);
-}
-
-
-int file_open(const char *name)
-{
 	int err;
-	id_t id;
 
 	err = hostflashsrv_lookup(name, &id);
 	if (err < 0) {
 		return err;
 	}
 
-	if (id > INT_MAX) {
-		return -1;
+	TEST_ASSERT_LESS_OR_EQUAL_INT_MESSAGE(INT_MAX, id, "TEST ERROR: file ID too big");
+
+	return (int)id;
+}
+
+
+int file_open(const char *name)
+{
+	id_t id;
+	int ret;
+
+	ret = file_lookup(name);
+	if (ret < 0) {
+		return ret;
 	}
 
-	err = hostflashsrv_open(&id);
-	if (err < 0) {
-		return err;
+	id = ret;
+
+	ret = hostflashsrv_open(&id);
+	if (ret < 0) {
+		return ret;
 	}
 
 	return (int)id;
--- a/meterfs/file_phx.c
+++ b/meterfs/file_phx.c
@@ -14,10 +14,9 @@
 #include <string.h>
 #include <unistd.h>
 #include <meterfs.h>
-#include <errno.h>
+#include <unity.h>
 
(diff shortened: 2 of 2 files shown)


nalajcie:
> @nalajcie @damianloew I made some (maybe incorrect) decision here, that potential test error I verify with simple C assert, not an unity assert. I feel that this way it's much clearer and better show programmer intention. As this is not common in our code, I would like to know, what do you opinion about it.

After some careful thinking I'm against using plain `assert` in unity test code:
 - if you use plain assert - the `TEST_TEAR_DOWN` won't be run - IMHO it's good practice to reset the test env (as much as possible) after a failed test - to ensure further tests won't false-fail
 - currently our unity test harness won't catch the assertion message and would wait until timeout (https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/master/trunner/harness/unity.py#L9)
 - you've used the `assert` correctly (the test would fail even if `NDEBUG`), but less careful developer might assume `assert` is always enough
 - IMHO this is the place you could use `FAIL(msg)` unity macro which would result in better error message in CI than `assert(false)`
 - on some hardware target configurations the `assert` might make the target suspend executing if the JTAG is connected and gdb server is running - good for debugging, but bad for CI (should not be the case now, but who knows how the future targets will be connected to CI ;))

Alas, I'm all-in regarding using `assert` in other code than the actual unity tests :).

anglov:
> > @nalajcie @damianloew I made some (maybe incorrect) decision here, that potential test error I verify with simple C assert, not an unity assert. I feel that this way it's much clearer and better show programmer intention. As this is not common in our code, I would like to know, what do you opinion about it.
> 
> After some careful thinking I'm against using plain `assert` in unity test code:
> 
>     * if you use plain assert - the `TEST_TEAR_DOWN` won't be run - IMHO it's good practice to reset the test env (as much as possible) after a failed test - to ensure further tests won't false-fail
> 
>     * currently our unity test harness won't catch the assertion message and would wait until timeout (https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/master/trunner/harness/unity.py#L9)
> 
>     * you've used the `assert` correctly (the test would fail even if `NDEBUG`), but less careful developer might assume `assert` is always enough
> 
>     * IMHO this is the place you could use `FAIL(msg)` unity macro which would result in better error message in CI than `assert(false)`
> 
>     * on some hardware target configurations the `assert` might make the target suspend executing if the JTAG is connected and gdb server is running - good for debugging, but bad for CI (should not be the case now, but who knows how the future targets will be connected to CI ;))
> 
> 
> Alas, I'm all-in regarding using `assert` in other code than the actual unity tests :).

Thanks for your insights! I wasn't concious about that. So I put unity asserts with message begins with "TEST ERROR:" – I hope it will be clear enough

anglov:
Bump

PR #411, opened 2025-08-11
Add script for test execution times comparison

wmlynik:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/test_compare/cmp.py
+++ b/test_compare/cmp.py
@@ -665,8 +665,8 @@ class Testcase(ResultNode):
 
 @dataclass
 class Args:
-    file_old: RootResultNode
-    file_new: RootResultNode
+    file1: RootResultNode
+    file2: RootResultNode
     verbose: int
     benchmarks: {}
     targets: []
@@ -686,8 +686,8 @@ def parse_args():
         epilog="Example: python3 cmp.py old_results.xml new_results.xml -v --threshold 5.0 -t ia32-generic-qemu",
     )
 
-    parser.add_argument("file_old", metavar="FILE_OLD", help="Path to the old test results file.")
-    parser.add_argument("file_new", metavar="FILE_NEW", help="Path to the new test results file.")
+    parser.add_argument("file1", metavar="FILE1", help="Path to the first test results file.")
+    parser.add_argument("file2", metavar="FILE2", help="Path to the second test results file.")
 
     parser.add_argument(
         "-b",
@@ -798,8 +798,8 @@ def process_args(args: argparse.Namespace) -> Args:
         except yaml.YAMLError as e:
             print(f"Error parsing YAML file: {e}")
     return Args(
-        parse_xml(args.file_old),
-        parse_xml(args.file_new),
+        parse_xml(args.file1),
+        parse_xml(args.file2),
         args.verbose,
         benchmarks,
         args.targets,
@@ -957,12 +957,12 @@ def print_rows(rows: list[TimeRow | StatusRow | EmptyRow | StatusRowHeader | Tim
 
 def main() -> None:
     args: Args = process_args(parse_args())
-    for file, name in [(args.file_old, "old"), (args.file_new, "new")]:
+    for file, name in [(args.file1, "file1"), (args.file2, "file2")]:
         fails = file.failures(args)
         if fails:
             fails_filtered = file.failures(args, filtered=True)
             if args.show_fails:
-                print(f"Failed tests in {name} file:")
+                print(f"Failed tests in {name}:")
                 fails_filtered.print(args)
             else:
                 print(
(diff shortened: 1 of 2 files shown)


wmlynik:
@gemini-code-assist review

test_compare/cmp.py
```diff
@@ -0,0 +1,679 @@
...
+def print_rows(rows):
+    if not rows:
+        return
+    max_name_len = max(len(row.name) for row in rows)
+    max_name_len = max(max_name_len, 4)
+    for row in rows:
+        if row.name == "":
+            print()
+            continue
+        if row.name == "H":
```
nalajcie, line 448: ?

test_compare/cmp.py
```diff
@@ -0,0 +1,679 @@
...
+            print(f"{ESCAPE_BOLD}{'':>{max_name_len}}║{'OLD':^9}║{'NEW':^9}║{'DIFFERENCE':^20}{ESCAPE_RESET}")
+            print(
+                f"{ESCAPE_BOLD}{ESCAPE_UNDERLINE}{'NAME':^{max_name_len}}║"
+                f"{'TIME':^9}║{'TIME':^9}║{'TIME':^10}║{'PERCENTAGE':^9}{ESCAPE_RESET}"
+            )
+            continue
+        print(
+            f"{row.style}{row.name:<{max_name_len}}{ESCAPE_RESET}{row.separator}"
```
nalajcie, line 456: most of this function should probably be done as `Row` class function. The `max_len_name` seems problematic as we need to scan all rows, but it could be an input parameter, maybe we would like to shorten the names regardless of the max len name (one test might have a really long name which would decrease the readability of all of the results.

I would appreciate adding second function to `Row` class - `to_csv()` and an option to output machine-readable results.

Not sure how we want to shorten max_name_len then

test_compare/cmp.py
```diff
@@ -0,0 +1,679 @@
...
+    ]
+    with_children = [
+        {"name": result["name"], "status_old": PRESENT, "status_new": PRESENT, "children": children}
+        for result, children in ((result, find_missing(result, level + 1)) for result in results["children"])
+        if children
+    ]
+    return only_old + only_new + with_children
+
+
+def find_fails(results, args, unfiltered=False, level=0, path=None):
```
nalajcie, line 561: could probably be implemented inside Testcase / Testsuite class

test_compare/cmp.py
```diff
@@ -0,0 +1,679 @@
...
+    if level == 3:
```
nalajcie, line 562: hard-coding max recursion level in multiple places, would be better to just always check if we have sub-results or not

test_compare/cmp.py
```diff
@@ -0,0 +1,679 @@
...
+        fails = find_fails(file, args, unfiltered=True)
+        if fails:
+            fails_filtered = find_fails(file, args)
+            if args.show_fails:
+                print(f"Failed tests in {name} file:")
+                print_fails(fails_filtered, args)
+            else:
+                print(
+                    f"{ESCAPE_BOLD}{ESCAPE_YELLOW}Warning:{ESCAPE_RESET} "
+                    f"{count_fails(fails)} tests failed in the {name} file "
```
nalajcie, line 665: use real filename not an alias

test_compare/cmp.py
```diff
@@ -0,0 +1,679 @@
...
+    if args.status_diff:
+        status_diff = find_missing(output)
+        print_status_rows(generate_status_rows(status_diff, args))
+    elif not args.show_fails:
+        rows = generate_time_rows(output["children"], args)
+        print_rows(rows)
+
+
+if __name__ == "__main__":
+    main()
```
nalajcie, line 679: think when the program should exit with non-zero return code

test_compare/cmp.py
```diff
@@ -0,0 +1,679 @@
...
+    args: Args = process_args(parse_args())
+    for file, name in [(args.file_old, "old"), (args.file_new, "new")]:
+        fails = find_fails(file, args, unfiltered=True)
+        if fails:
+            fails_filtered = find_fails(file, args)
+            if args.show_fails:
+                print(f"Failed tests in {name} file:")
+                print_fails(fails_filtered, args)
+            else:
+                print(
```
nalajcie, line 663: this is a script for diff'ing 2 JUnit XML files - use-case wise - I would like to see that the test started (or stopped) failing without any extra params as this is crucial info.

test_compare/cmp.py
```diff
@@ -0,0 +1,679 @@
...
+        args.threshold_relative,
+        args.threshold_filter,
+        args.status_diff,
+        args.show_fails,
+    )
+
+
+def split_path(s):
+    index = s.find("/")
+    index = max(index, s.find("/", index + 1))
```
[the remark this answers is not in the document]
wmlynik: The idea here was to take at most two segments of the path as location so we a hierarchy - location, test suite, test case. Test suites might be organized  in sub-directories within location, but they are still at the same level of this hierarchy.

PR #413, opened 2025-08-22
trunner: make kwargs have test name and cmd by default 

adamdebek:
Some tests need to parse the command or its arguments themselves, so the appropriate field has to be passed to the harness.

JIRA: CI-584

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -152,6 +152,7 @@ class ConfigParser:
 
             execute_binary = False
 
+        self.test.cmd = cmd
         cmd = shlex.split(cmd)
         if not cmd:
             raise ParserError("execute/run attribute cannot be empty")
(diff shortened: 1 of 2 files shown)


damianloew (commented):
@adamdebek what do you mean by `Some tests`, for what tests do you need such changes?

damianloew (approved):
After looking at the use case (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/416/files#diff-3d935f1933fe56848b078ef47bd3c40bba805cc9bf859ba0527bd7e4cff35251) I think it's worth to add those arguments as default to access from the harness.

About parsing raw cmd, in this particular case we probably should parse command in lsb vsx harness as the `testname{testcases_ids}` semantic is lsb vsx specific

PR #434, opened 2025-10-23
armv8m33: split MCXN94x target for tests on CPU0/1

maska989:
JIRA: CI-592

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv8m33-mcxn94x-frdm.

## Checklist:

## Special treatment

 * https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1442


mateusz-bloch (commented):
At the moment, I can see that the CoreMark tests are not running:

Image

mateusz-bloch (dismissed):
LGTM

damianloew (changes requested):
Same suggestions about commit title as in the second PR.

In this commit:

```
tests: exclude non-supported tests
Disable execution for non supported tests on MCXN947 platform
```

Also our golden rules are broken - in commit description it's not WHY - in this case additional explanation is not necessary, so only commit title would be fine, sth like `exclude non-supported tests on MCXN947 platform` wouldn't be too long.

damianloew (commented):
Still the commit description is not 'why' in the second commit:

```
tests: exclude non-supported tests
Exclude non-supported tests on MCXN947 platform
```

The commit title should show on what architecture you are excluding the particular tests. 

```
tests: exclude non-supported tests on mcxn94x-frdm_cpu1
```
wouldn't be too long 

If you would want to add some explanation in the commit message body, you can also leave sth like:

Needed after adding cpu1 configuration for mcxn94x to exclude same tests as for the cpu0 configuration.

Above that lgtm

damianloew (commented):
Remember about marking `All new and existing linter checks and tests passed.`

trunner/target/armv8m33.py
```diff
@@ -134,3 +133,11 @@ def build_test(self, test: TestOptions):
         builder.add(test.harness)
 
         return builder.get_harness()
+
+
+class MCXN94xCPU0Target(MCXN94xTarget):
+    name = "armv8m33-mcxn94x-frdm"
```
lukkrusz, line 139: nit-picking but wouldn't it be better to call it cpu0?

```suggestion
    name = "armv8m33-mcxn94x-frdm_cpu0"
```

lukkrusz (changes requested):
Code-wise, looks OK, except for the target naming that I'd reconsider.
Two issues remain:
- `All new and existing linter checks and tests passed.`  is left unchecked
- commit message needs to be updated as previously suggested

coremark_pro/timeouts.yaml
```diff
@@ -118,4 +118,8 @@ armv8m33-mcxn94x-frdm:
   core: [2, 92]
   linear_alg-mid-100x100-sp: [2, 802]
   nnet_test: [2, 4588]
-  
\ No newline at end of file
+  
+armv8m33-mcxn94x-frdm_cpu1:
+  core: [2, 92]
+  linear_alg-mid-100x100-sp: [2, 802]
+  nnet_test: [2, 4588]
```
mateusz-bloch: Image

psh/test.yaml
```diff
@@ -54,7 +55,7 @@ test:
                 sparcv8leon-generic-qemu,
                 armv7r5f-zynqmp-qemu,
                 armv8m33-mcxn94x-frdm,
-                armv8m55-stm32n6-nucleo
+                armv8m33-mcxn94x-frdm_cpu1,
```
mateusz-bloch: why deleting n6?
Image

trunner/target/armv8m33.py
maska989: About nitpick, it cannot be easily done because the runner picks up the folder from _boot and _fs via the built target name rn. We're building these two. It may change someday, but right now it must be done this way.

mateusz-bloch: yeah,  I’d stick with the target names defined in the project
image

coremark_pro/test_coremark_pro.yaml
```diff
@@ -4,12 +4,12 @@ test:
     nightly: true
 
     targets:
-      exclude: [host-generic-pc, sparcv8leon-generic-qemu, armv7r5f-zynqmp-qemu]
+      exclude: [host-generic-pc, sparcv8leon-generic-qemu, armv7r5f-zynqmp-qemu, amrv8m33-mcxn94x-frdm, amrv8m33-mcxn94x-frdm_cpu1]
```
mateusz-bloch:
```suggestion
      exclude: [host-generic-pc, sparcv8leon-generic-qemu, armv7r5f-zynqmp-qemu]
```

lukkrusz (approved):
LGTM

PR #441, opened 2025-12-31
trunner: add armv8m55-stm32n6-nucleo target

wmlynik:
JIRA: CI-598

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

  - https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1533
  - https://github.com/phoenix-rtos/phoenix-rtos-docker/pull/43
  - https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1625

### DIFF
--- a/coremark_pro/timeouts.yaml
+++ b/coremark_pro/timeouts.yaml
@@ -106,3 +106,10 @@ armv7r5f-zynqmp-qemu:
   linear_alg-mid-100x100-sp: [2, 9]
   parser-125k: [2, 3]
   zip-test: [36, 2]
+
+armv8m55-stm32n6-nucleo:
+  cjpeg-rose7-preset: [2, 3]
+  core: [2, 14]
+  nnet_test: [2, 160]
+  sha-test: [2, 3]
+  linear_alg-mid-100x100-sp: [2, 8]
--- a/trunner/harness/plo.py
+++ b/trunner/harness/plo.py
@@ -140,9 +140,9 @@ class PloInterface:
         self.send_cmd(cmd)
 
         try:
-            self.dut.expect_exact("Serious risk of data loss, type YES! to proceed.")
-            self.send_cmd("YES!")
-            self.dut.expect_exact("Erased", timeout=timeout)
+            if self.dut.expect_exact(["Serious risk of data loss, type YES! to proceed.", "Erased"]) == 0:
+                self.send_cmd("YES!")
+                self.dut.expect_exact("Erased", timeout=timeout)
         except pexpect.TIMEOUT as e:
             raise PloError("Wrong erase command output!", cmd=cmd, output=self.dut.before) from e
 
--- a/trunner/harness/plo.py
+++ b/trunner/harness/plo.py
@@ -140,9 +140,14 @@ class PloInterface:
         self.send_cmd(cmd)
 
         try:
-            if self.dut.expect_exact(["Serious risk of data loss, type YES! to proceed.", "Erased"]) == 0:
+            erase_needs_confirming = (
+                self.dut.expect_exact(["Serious risk of data loss, type YES! to proceed.", "Erased"]) == 0
+            )
+
+            if erase_needs_confirming:
                 self.send_cmd("YES!")
                 self.dut.expect_exact("Erased", timeout=timeout)
+
         except pexpect.TIMEOUT as e:
             raise PloError("Wrong erase command output!", cmd=cmd, output=self.dut.before) from e
 
--- a/trunner/target/armv8m55.py
+++ b/trunner/target/armv8m55.py
@@ -1,0 +1,229 @@
+import time
+from pathlib import Path
+from typing import Callable, Optional, Sequence, TextIO
+
+from trunner.ctx import TestContext
+from trunner.dut import Dut, SerialDut
+from trunner.host import Host
+from trunner.harness import (
+    IntermediateHarness,
+    TerminalHarness,
+    PloInterface,
+    PloHarness,
+    PloError,
+    ShellHarness,
+    TestStartRunningHarness,
+    Rebooter,
+    RebooterHarness,
+    HarnessBuilder,
+    FlashError,
+)
+from trunner.tools import GdbInteractive, OpenocdGdbServer, OpenocdError
+from trunner.types import AppOptions, TestOptions, TestResult
+from .base import TargetBase, find_port
+
+
+class ARMv8M55Rebooter(Rebooter):
+    # NOTE: changing boot modes not needed/supported for this target
+
+    def __init__(self, host, dut, gdb):
+        super().__init__(host, dut)
+        self.gdb = gdb
+
+    def __call__(self, flash=False, hard=False):
+        """Sets flash mode and perform hard or soft & debugger reboot based on `hard` flag."""
+
+        if hard and self.host.has_gpio():
+            self._reboot_dut_gpio(hard=hard)
+        else:
+            self._reboot_by_debugger()
+
+    def _reboot_by_debugger(self):
+        with self.gdb.run():
+            self.gdb.set_architecture("arm")
+            self.gdb.connect()
+            self.gdb.reset()
+            self.gdb.cont()
+            time.sleep(0.5)
+
+
+class STM32N6STLinkGdbServerHarness(IntermediateHarness):
+    """Harness that runs other harness in ST-LINK gdb server context.
+
+    It calls passed harness under gdb server context and then continue normal
+    harness execution flow.
+
+    Attributes:
+        harness: Harness that will be run under gdb server context.
+
+    """
+
+    def __init__(self, harness: Callable[[TestResult], TestResult]):
+        super().__init__()
+        self.harness = harness
+
+    def __call__(self, result: TestResult) -> TestResult:
+        with OpenocdGdbServer(interface="stlink-dap", target="stm32n6x").run():
+            self.harness(result)
+
+        return self.next_harness(result)
+
+
+class STM32N6PloAppLoader(TerminalHarness, PloInterface):
+    def __init__(self, dut: Dut, apps: Sequence[AppOptions], gdb: GdbInteractive):
+        TerminalHarness.__init__(self)
+        PloInterface.__init__(self, dut)
+        self.dut = dut
+        self.apps = apps
+        self.gdb = gdb
+        self.load_offset = 0x00
+        self.ram_addr = 0x34100000
+        self.page_sz = 0x200
+
+    def _aligned_app_size(self, path: Path):
+        sz = path.stat().st_size
+        # round up to the size of the page
+        offset = (self.page_sz - sz) % self.page_sz
+        return sz + offset
+
+    def __call__(self):
+        # First, load the apps into ram memory using gdb
+        with self.gdb.run():
+            offset = self.load_offset
+            self.gdb.set_architecture("arm")
+            self.gdb.connect()
+
+            for app in self.apps:
+                path = self.gdb.cwd / Path(app.file)
+                sz = self._aligned_app_size(path)
+
+                self.gdb.load(path, self.ram_addr + offset)
+                offset += sz
+
+            self.gdb.cont()
+
+            # When closing right after continuing plo will get stuck
+            time.sleep(0.5)
+
+        # Secondly, map loaded binaries as runnable programs
+        offset = self.load_offset
+        for app in self.apps:
+            path = self.gdb.cwd / Path(app.file)
+            sz = path.stat().st_size
+
+            self.alias(app.file, offset=offset, size=sz)
+            self.app("ramdisk", app.file, "axi_app", "axi_app")
+
+            offset += self._aligned_app_size(path)
+
+
+class STM32N6Target(TargetBase):
+    name = "armv8m55-stm32n6-nucleo"
+    rootfs = False
+    experimental = False
+    image_file = "phoenix.disk.bin"
+    plo_file = "plo-ram.elf"
+    image_addr = 0x70000000
+    ram_addr = 0x34100000
+
+    def __init__(self, host: Host, port: Optional[str] = None, baudrate: int = 115200):
+        if port is None:
+            # Try to find ST-LINK
+            port = find_port("ST-LINK")
+        self.port = port
+        self.host = host
+        self.dut = SerialDut(port, baudrate, encoding="utf-8", codec_errors="ignore")
+        self.rebooter = ARMv8M55Rebooter(host, self.dut, GdbInteractive(port=3333))
+        self.baudrate = baudrate
+        super().__init__()
+
+    @classmethod
+    def from_context(cls, ctx: TestContext):
+        return cls(ctx.host, ctx.port, ctx.baudrate)
+
+    def flash_dut(self, host_log: TextIO):
+        try:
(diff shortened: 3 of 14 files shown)


coremark_pro/timeouts.yaml
```diff
@@ -106,3 +106,10 @@ armv7r5f-zynqmp-qemu:
   linear_alg-mid-100x100-sp: [2, 9]
   parser-125k: [2, 3]
   zip-test: [36, 2]
+
+armv8m55-stm32n6-nucleo:
+  cjpeg-rose7-preset: [2, 3]
+  core: [2, 14]
+  nnet_test: [2, 160]
+  sha-test: [2, 3]
+  linear_alg-mid-100x100-sp: [2, 8]
```
lukkrusz, line 115: Is it intended to not have `parser-125k` and `zip-test` workloads here or is it something to be updated?

trunner/harness/plo.py
```diff
@@ -140,9 +140,9 @@ def erase(self, device: str, offset: int, size: int, timeout: Optional[int] = No
         self.send_cmd(cmd)
 
         try:
-            self.dut.expect_exact("Serious risk of data loss, type YES! to proceed.")
-            self.send_cmd("YES!")
-            self.dut.expect_exact("Erased", timeout=timeout)
+            if self.dut.expect_exact(["Serious risk of data loss, type YES! to proceed.", "Erased"]) == 0:
```
lukkrusz: Here's a minor readibility improvement I suggest

```suggestion
            erase_needs_confirming = self.dut.expect_exact(["Serious risk of data loss, type YES! to proceed.", "Erased"]) == 0
            if erase_needs_confirming:
                ...
```

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,240 @@
...
+    part_plo = "part_plo.img"
+    part_user = "part_user.img"
+    plo_file = "plo-ram.elf"
+    part_user_offset = 0x10000
+    ram_addr = 0x34100000
+
+    def __init__(self, host: Host, port: Optional[str] = None, baudrate: int = 115200):
+        if port is None:
+            # Try to find ST-LINK
+            port = find_port("ST-LINK")
```
lukkrusz, line 136: I find the comment here redundant
```suggestion
        if port is None:
            port = find_port("ST-LINK")
```

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,240 @@
...
+        super().__init__()
+
+    @classmethod
+    def from_context(cls, ctx: TestContext):
+        return cls(ctx.host, ctx.port, ctx.baudrate)
+
+    def flash_dut(self, host_log: TextIO):
+        try:
+            self.rebooter(hard=True, flash=True)
+            with OpenocdGdbServer(interface="stlink", target="stm32n6x").run():
```
lukkrusz: Why are we not setting `host_log` here?

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,240 @@
...
+                    # Copy from ramdisk to flash
+                    # TODO: use image size rather  than ramdisk size
+                    plo.copy("ramdisk", 0x0, "flash0", 0x0, 0x80000, 0x80000)
+
+                    self.rebooter(hard=True, flash=False)
+
+        except FileNotFoundError as e:
+            raise FlashError(msg=str(e)) from e
+        except OpenocdError as e:
+            raise FlashError(msg=str(e)) from e
```
lukkrusz:
```suggestion
        except (FileNotFoundError, OpenocdError) as e:
            raise FlashError(msg=str(e)) from e
```
or
```suggestion
        except (FileNotFoundError, OpenocdError, PloError) as e:
            raise FlashError(msg=str(e)) from e
```
if the previous comment remains unchallenged

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,240 @@
...
+                syspage_loader = None
+
+                if test.bootloader and test.bootloader.apps:
+                    syspage_loader = STM32N6PloAppLoader(
+                        dut=self.dut,
+                        apps=test.bootloader.apps,
+                        gdb=GdbInteractive(port=3333, cwd=self.root_dir() / test.shell.path),
+                    )
+
+                builder.add(PloHarness(self.dut, syspage_loader=syspage_loader))
```
lukkrusz: This needs to be re-evalued, if the behaviour is as expected, then the logic can be simplified and extra indentation removed:

```suggestion
            shell_path = getattr(test.shell, "path", None) if test.shell else None
            if test.bootloader:
                syspage_loader = STM32N6PloAppLoader(
                    dut=self.dut,
                    apps=test.bootloader.apps,
                    gdb=GdbInteractive(port=3333, cwd=self.root_dir() / shell_path),
                ) if test.bootloader.apps and shell_path else None

                builder.add(PloHarness(self.dut))
```

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,240 @@
...
+            builder.add(RebooterHarness(self.rebooter, flash=False))
+
+            if test.bootloader is not None:
+                syspage_loader = None
+
+                if test.bootloader and test.bootloader.apps:
+                    syspage_loader = STM32N6PloAppLoader(
+                        dut=self.dut,
+                        apps=test.bootloader.apps,
+                        gdb=GdbInteractive(port=3333, cwd=self.root_dir() / test.shell.path),
```
lukkrusz: Magic number `port=3333` - probably should become a static constant of the class for better readability

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,240 @@
...
+                    # The board reboots after loading PLO for the first time,
+                    # load it again if plo prompt doesn't appear
+                    try:
+                        plo.wait_prompt(timeout=2)
+                    except PloError:
+                        gdb.pause()
+                        gdb.load_program(self.plo_file)
+                        gdb.cont()
+
+                        plo.wait_prompt()
```
lukkrusz, line 174: A second failure would raise `PloError` without catching it. Don't we want it to be caught and raised as FlashError in a similar fashion as at the end of this function?

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,240 @@
...
+)
+from trunner.tools import GdbInteractive, OpenocdGdbServer, OpenocdError
+from trunner.types import AppOptions, TestOptions, TestResult
+from .base import TargetBase, find_port
+
+
+class ARMv8M55Rebooter(Rebooter):
+    # NOTE: changing boot modes not needed/supported for this target
+
+    def __init__(self, host, dut, gdb):
```
lukkrusz: Missing types

lukkrusz (changes requested):
Noticed a couple of things that could be improved or need explaining

trunner/target/armv8m55.py
lukkrusz: `target="stm32n6x"`

OpenOCD 0.12.0 version, which we currently use for [gh-runners and devel](https://github.com/phoenix-rtos/phoenix-rtos-docker/blame/97c1d6d0eb60d859cf3644cda932ceece303eeab/gh-runner/Dockerfile#L31) does not contain the config script for this target.

This PR is dependent on https://github.com/phoenix-rtos/phoenix-rtos-docker/pull/43
Otherwise, we need to get the config script directly and run openocd with it

lukkrusz (changes requested):

trunner/harness/plo.py
maska989: This change was in first iteration in commits, which doesn't contain automation of state changing and needed manual labor.

trunner/target/armv8m55.py
maska989: We need approach like this, because we use a non-standard FlexRAM configuration the board will reset immediately after continue if this is the first run after power on.

trunner/target/armv8m55.py
maska989: Added to `Special treatment`

lukkrusz: Following an internal discussion, that PR (https://github.com/phoenix-rtos/phoenix-rtos-docker/pull/43) is no longer relevant and we might need to create a new one, in a similar fashion as here: https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1582

We will then need to change our approach a bit and parse the config path into the openocd command, instead of relying on it being natively supported

trunner/target/armv8m55.py
lukkrusz: My question was more about what happens after the prompt doesn't appear the second time, at `plo.wait_prompt()` (or comes with an error) - we're not catching it at `except (FileNotFoundError, OpenocdError) as e:`, line 194.

I think it'd have to be extended to `except (FileNotFoundError, OpenocdError, PloError) as e:`

coremark_pro/timeouts.yaml
lukkrusz: @maska989 could you confirm that we do not want to include those workloads here?

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,237 @@
...
+class STM32N6STLinkGdbServerHarness(IntermediateHarness):
+    """Harness that runs other harness in ST-LINK gdb server context.
+
+    It calls passed harness under gdb server context and then continue normal
+    harness execution flow.
+
+    Attributes:
+        harness: Harness that will be run under gdb server context.
+
+    """
```
lukkrusz, line 60: Since I'm at it, the extra newline should be deleted
```suggestion
    Attributes:
        harness: Harness that will be run under gdb server context.
    """
```

trunner/target/armv8m55.py
lukkrusz: OK, the PR for that is already approved, not yet merged - https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1625

I suggest to begin working on changing the approach - parsing the config path into openocd command

trunner/harness/plo.py
lukkrusz: But it's still included, I only suggest adding an intermediate variable to help the readability of this code (a result holder)

lukkrusz (changes requested):

coremark_pro/timeouts.yaml
maska989: I'm not familiar with coremark sadly, more information about those changes can provide @wmlynik

wmlynik: Timeouts are only defined for the tests that run on given target. Some tests are excluded on this target due to its limitations.

trunner/target/armv8m55.py
maska989: Done

trunner/harness/plo.py
maska989: Done

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def flash_dut(self, host_log: TextIO):
+        try:
+            self.rebooter(hard=True, flash=True)
+            with OpenocdGdbServer(
+                interface="stlink",
+                # TODO after merge https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1625
+                # use this approach
+                # target_path=[self._project_dir(), "stm32n6/stm32n6x.cfg"],
+                target="stm32n6x",
```
lukkrusz: The PR is now merged, so we can update this code.

I still have some reservations about this approach that I'll address in a separate comment, though

trunner/tools/gdb.py
```diff
@@ -106,6 +144,7 @@ def __init__(
         self,
         interface: Optional[str] = None,
         target: Optional[str] = None,
+        target_path: Optional[tuple[str, str]] = None,
```
lukkrusz: I don't think this approach is good practice, the `OpenocdProcess` is getting bloated with an extra case-specific variable.

I was shown an alternative approach for r5 target: https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/ipcei/1.4/trunner/target/armv7r5f.py

Unlike there, I'd keep using `OpenGdbServer` but the openocd script path should, imho, be in the STM32N6Target class, just like in the linked approach:
`self.openocd_config = f"{self._project_dir()}/{self.openocd_config_file}"`

Now, we'd need to come up with something that would not add an extra variable and complex logic. Maybe pass the full path to the config as a `target` and do `elif self.target.endswith(".cfg")` in `OpenocdProcess._start()`?

trunner/tools/gdb.py
```diff
@@ -124,14 +164,21 @@ def __init__(
...
+            raise OpenocdError("Target or Interface arguments missing")
 
     def _start(self):
         # Use pexpect.spawn to run a process as PTY, so it will flush on a new line
         if self.board:
             args = ["-f", f"board/{self.board}.cfg"]
+        elif self.target_path:
+            args = [
+                "-f",
+                f"interface/{self.interface}.cfg",
```
lukkrusz: If target path is provided but not the interface, will it not be "interface/None.cfg"?

lukkrusz (changes requested):
The OpenocdProcess changes need to be reconsidered, I dropped some suggestions.

Otherwise looks OK

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,302 @@
...
+    Attributes:
+        harness: Harness that will be run under gdb server context.
+    """
+
+    def __init__(self, harness: Callable[[TestResult], TestResult]):
+        super().__init__()
+        self.harness = harness
+
+    def __call__(self, result: TestResult) -> TestResult:
+        openocd_config_file = "scripts/openocd/stm32n6/stm32n6x.cfg"
```
lukkrusz: This should be a class constant, it already is defined in STM32N6Target class. Multiple definitions of the same are suboptimal.
If we were lazy, we could do:
```suggestion
        openocd_config_file = STM32N6Target.openocd_config_file
```
Ideally, the openocd_config_file would only be defined and used (passed) within a single class

trunner/target/armv8m55.py
```diff
@@ -0,0 +1,302 @@
...
+        except (pexpect.TIMEOUT, pexpect.EOF) as e:
+            self.output = proc.before
+            raise OpenocdError("Failed to connect to target", output=self.output) from e
+        finally:
+            proc.close()
+
+    def flash_dut(self, host_log: TextIO):
+        try:
+            self.rebooter(hard=True, flash=True)
+            with self._run_openocd():
```
lukkrusz: I'm not sure about this. We already have CustomOpenocdProcess and CustomOpenocdGdbServer, yet we run openocd with a separate function only to accommodate the openocd_config_file path argument

It works but it's a lot of added code

lukkrusz (changes requested):
I'm still at odds with this approach. To reiterate my previous suggestion, I'd keep using `OpenGdbServer` and we could pass the full path to the config as a target and then, inside the `OpenOcdProcess`._start(), add:
```python
    elif self.target.endswith(".cfg"):
        args = ["-f", f"interface/{self.interface}.cfg", "-f", self.target]
```
or add it as an extra arg, at the beginning.

With this, we could also use `OpenocdGbServer` in `STM32N6STLinkGdbServerHarness`, thus eliminating the need for custom solutions defined at the top.

lukkrusz (approved):
LGTM

PR #445, opened 2026-01-19
trunner: Implementing pytest support

lukkrusz:
**PyTest support implementation for our existing testing framework.**

## Description
The implementation includes
- pytest harness
- pytest plugins:
  - plugin for IO Stream control
  - plugin bridging PyTest with our TestRunner (report)
- sample pytest tests with conftest.py
- TestType Enum for simpler and more controllable input validation

## Types of changes

- [x] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work ([#449](https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/449)).

## Previous conversation: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/443

### DIFF
--- a/sample/pytest/conftest.py
+++ b/sample/pytest/conftest.py
@@ -1,0 +1,25 @@
+import pytest
+
+
+@pytest.fixture(scope="session")
+def pexpect_bin(dut):
+    p = dut.pexpect_proc
+    p.expect_exact("[Commence Fake Communication]\r\n")
+
+    yield p
+
+    p.sendline("EXIT")
+    p.expect_exact("[Success!]")
+
+
+@pytest.fixture(scope="session")
+def pexpect_psh(dut, ctx):
+    p = dut.pexpect_proc
+    exec_cmd = "/bin/psh" if ctx.target.rootfs else "sysexec psh"
+    p.sendline(exec_cmd)
+    p.expect(rf"{exec_cmd}(\r+)\n")
+
+    yield p
+
+    p.send("\x04")
+    p.expect("exit(\r+)\n")
--- a/sample/pytest/main.c
+++ b/sample/pytest/main.c
@@ -1,0 +1,71 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Simple C app for simulating serial communication
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Lukasz Kruszynski
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#define BUF_SZ 256U
+
+
+static int main_processCommand(char *input)
+{
+	const char *echoCmd = "echo ";
+
+	if (strcmp(input, "EXIT") == 0) {
+		return 0;
+	}
+	else if (strncmp(input, echoCmd, strlen(echoCmd)) == 0) {
+		(void)printf("%s\n", input + strlen(echoCmd));
+	}
+	else if (strcmp(input, "echo") == 0 || (strcmp(input, "ping") == 0)) {
+		(void)printf("main: [OK]\n");
+	}
+	else if (strlen(input) > 0U) {
+		(void)printf("main: %s [FAIL]\n", input);
+	}
+	return 1;
+}
+
+
+int main(void)
+{
+	char c = '\0';
+	char buffer[BUF_SZ] = { 0 };
+	size_t pos = 0U;
+	int returnCode = 1;
+
+	(void)printf("main: [Commence Fake Communication]\n");
+
+	/* Run the loop until `EXIT` */
+	while (returnCode != 0) {
+		if (read(STDIN_FILENO, &c, 1) != 1) {
+			return -1;
+		}
+
+		if (c == '\n') {
+			buffer[pos] = '\0';
+			returnCode = main_processCommand(buffer);
+			pos = 0U;
+		}
+		else if (pos < BUF_SZ - 1U) {
+			buffer[pos] = c;
+			pos++;
+		}
+		else {
+			(void)printf("main: [Failure!]\n");
+			return -1;
+		}
+	}
+
+	(void)printf("main: [Success!]\n");
+	return 0;
+}
--- a/sample/pytest/test-psh.py
+++ b/sample/pytest/test-psh.py
@@ -1,0 +1,39 @@
+import psh.tools.psh as psh
+import pytest
+
+TEST_DIR = "test_dir"
+
+echo_resp_code_list = [
+    ("", "", "success"),
+    ("hello", "hello", "success"),
+    ("hello world", "hello world", "success"),
+    ("lorem $ipsum dolor$ales", "lorem  dolor", "success"),
+    ("hello > /", ".*EISDIR", "success"),
+    ("-illegal_arg", ".*", "fail"),
+]
+
+
+@pytest.mark.parametrize("msg, resp, exp_res", echo_resp_code_list)
+def test_echo(pexpect_psh, msg, resp, exp_res):
+    psh.assert_cmd(
+        pexpect_psh,
+        f"echo {msg}",
+        expected=rf"{resp}\r+\n",
+        result=exp_res,
+        msg=f'echo "{msg}" failed',
+        is_regex=True,
+    )
+
+
+def is_dir_in_files(dirname: str, files: list[psh.File]) -> bool:
+    return any(file.is_dir and file.name == dirname for file in files)
+
+
+def test_mkdir_rmdir(pexpect_psh):
+    psh.assert_cmd(pexpect_psh, f"mkdir {TEST_DIR}")
+
+    files = psh.ls(pexpect_psh, "/")
+    assert is_dir_in_files(TEST_DIR, files)
+    psh.assert_cmd(pexpect_psh, f"rmdir {TEST_DIR}")
+    files = psh.ls(pexpect_psh, "/")
+    assert not is_dir_in_files(TEST_DIR, files)
--- a/sample/pytest/test.py
+++ b/sample/pytest/test.py
@@ -1,0 +1,44 @@
+import pytest
+
+
+test_data = [
+    ("ping", "[OK]"),
+    ("echo", "[OK]"),
+    ("abcd", "[FAIL]"),
+]
+
+
+def test_ctx(ctx):
+    assert ctx is not None
+
+
+def test_dut(dut):
+    assert dut is not None
+
+
+@pytest.mark.parametrize("cmd, exp", test_data)
+def test_dut_parameterized_commands(pexpect_bin, cmd, exp):
+    pexpect_bin.sendline(cmd)
+    pexpect_bin.expect_exact(f"{exp}\r\n")
+
+
+@pytest.mark.usefixtures("pexpect_bin")
+class TestClass:
+    test_val = 0
+
+    def setup_method(self):
+        self.__class__.test_val = 5
+
+    def teardown_method(self):
+        self.__class__.test_val = 0
+
+    def test_send_hello_from_class(self, pexpect_bin):
+        pexpect_bin.sendline("echo hello from class")
+        pexpect_bin.expect_exact("\r\nhello from class\r\n")
+
+    def test_test_val_value(self):
+        assert self.__class__.test_val == 5
+
+
+def test_class_static_val_is_zero():
+    assert TestClass.test_val == 0
--- a/sample/pytest/test.yaml
+++ b/sample/pytest/test.yaml
@@ -1,0 +1,12 @@
+test:
+    tests:
+        - name: binary
+          type: pytest
+          execute: test-fake-comm
+          script: test.py
+          targets:
+              include: [host-generic-pc]
+
+        - name: psh
+          type: pytest
+          script: test-psh.py
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -6,9 +6,8 @@ from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Set
 
 import yaml
-
 from trunner.ctx import TestContext
-from trunner.harness import PyHarness, unity_harness
+from trunner.harness import PyHarness, unity_harness, pytest_harness
 from trunner.types import AppOptions, BootloaderOptions, FileOptions, TestOptions, ShellOptions, TestType
 
 
@@ -59,6 +58,8 @@ class ConfigParser:
 
         if t_type in (TestType.EMPTY, TestType.HARNESS):
             self._parse_pyharness(config)
+        elif t_type is TestType.PYTEST:
+            self._parse_pytest(config)
         elif t_type is TestType.UNITY:
             self._parse_unity()
         else:
@@ -69,13 +70,7 @@ class ConfigParser:
         if path is None:
             raise ParserError("test is of type 'harness' but there is no \"harness\" keyword")
 
-        path = Path(path)
-
-        if not path.is_absolute():
-            # If path is not absolute then it must be relative to the directory of yaml
-            path = self.yaml_path.parent / path
-            if not path.is_absolute():
-                raise ParserError("yml path is not absolute!")
+        path = self._get_absolute_path(path)
 
         spec = importlib.util.spec_from_file_location("harness", path.absolute())
         if not spec:
@@ -94,6 +89,21 @@ class ConfigParser:
             raise ParserError(f"harness function has not been found in {path}")
 
         self.test.harness = PyHarness(self.ctx.target.dut, self.ctx, harness_fn, self.test.kwargs)
+
+    def _parse_pytest(self, config: dict) -> None:
+        script = config.get("script")
+        if not script:
+            raise ParserError("pytest `script` name must be provided!")
+        script_path = self._get_absolute_path(script)
+        if script_path.suffix != ".py":
+            raise ParserError("provided `script` is not a python script!")
+        if not script_path.exists():
+            raise ParserError(f"pytest script not found: {script_path}")
+        # Creating a new kwarg is better than expanding TestContext
+        # with a field required only for this specific case
+        # TODO Think of a better way of parsing special arguments
+        self.test.kwargs["script"] = script_path.absolute()
+        self.test.harness = PyHarness(self.ctx.target.dut, self.ctx, pytest_harness, self.test.kwargs)
 
     def _parse_unity(self):
         self.test.harness = PyHarness(self.ctx.target.dut, self.ctx, unity_harness, self.test.kwargs)
@@ -335,6 +345,10 @@ class ConfigParser:
 
         self.main.ignore = config.get("ignore", False)
         self.main.nightly = config.get("nightly", False)
+
+    def _get_absolute_path(self, path: str) -> Path:
+        p = Path(path)
+        return self.yaml_path.parent / p
 
     def parse(self, path: Path) -> List[TestOptions]:
         self.yaml_path = path
--- a/trunner/config.py
+++ b/trunner/config.py
@@ -6,9 +6,8 @@ from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Set
 
 import yaml
-
 from trunner.ctx import TestContext
-from trunner.harness import PyHarness, unity_harness
+from trunner.harness import PyHarness, unity_harness, pytest_harness
 from trunner.types import AppOptions, BootloaderOptions, FileOptions, TestOptions, ShellOptions, TestType
 
 
@@ -59,6 +58,8 @@ class ConfigParser:
 
         if t_type in (TestType.EMPTY, TestType.HARNESS):
             self._parse_pyharness(config)
+        elif t_type is TestType.PYTEST:
+            self._parse_pytest(config)
         elif t_type is TestType.UNITY:
             self._parse_unity()
         else:
@@ -69,13 +70,7 @@ class ConfigParser:
         if path is None:
             raise ParserError("test is of type 'harness' but there is no \"harness\" keyword")
 
-        path = Path(path)
-
-        if not path.is_absolute():
-            # If path is not absolute then it must be relative to the directory of yaml
-            path = self.yaml_path.parent / path
-            if not path.is_absolute():
-                raise ParserError("yml path is not absolute!")
+        path = self._get_absolute_path(path)
 
         spec = importlib.util.spec_from_file_location("harness", path.absolute())
         if not spec:
@@ -94,6 +89,21 @@ class ConfigParser:
             raise ParserError(f"harness function has not been found in {path}")
 
         self.test.harness = PyHarness(self.ctx.target.dut, self.ctx, harness_fn, self.test.kwargs)
+
+    def _parse_pytest(self, config: dict) -> None:
+        script = config.get("script")
+        if not script:
+            raise ParserError("pytest `script` name must be provided!")
+        script_path = self._get_absolute_path(script)
+        if script_path.suffix != ".py":
+            raise ParserError("provided `script` is not a python script!")
+        if not script_path.exists():
+            raise ParserError(f"pytest script not found: {script_path}")
+        # Creating a new kwarg is better than expanding TestContext
+        # with a field required only for this specific case
+        # TODO Think of a better way of parsing special arguments
+        self.test.kwargs["script"] = script_path.absolute()
+        self.test.harness = PyHarness(self.ctx.target.dut, self.ctx, pytest_harness, self.test.kwargs)
 
     def _parse_unity(self):
         self.test.harness = PyHarness(self.ctx.target.dut, self.ctx, unity_harness, self.test.kwargs)
@@ -335,6 +345,14 @@ class ConfigParser:
 
         self.main.ignore = config.get("ignore", False)
         self.main.nightly = config.get("nightly", False)
+
+    def _get_absolute_path(self, path: str) -> Path:
+        resolved_path = self.yaml_path.parent / Path(path)
+
+        if not resolved_path.is_absolute():
+            raise ParserError("yml path is not absolute!")
+
+        return resolved_path
 
     def parse(self, path: Path) -> List[TestOptions]:
         self.yaml_path = path
--- a/trunner/harness/__init__.py
+++ b/trunner/harness/__init__.py
@@ -24,6 +24,7 @@ from .plo import (
 from .psh import ShellHarness
 from .pyharness import PyHarness
 from .unity import unity_harness
+from .pytest import pytest_harness
 
 __all__ = [
     "HarnessBuilder",
@@ -48,4 +49,5 @@ __all__ = [
     "PloImageProperty",
     "PloJffsImageProperty",
     "unity_harness",
+    "pytest_harness",
 ]
--- a/trunner/harness/pytest.py
+++ b/trunner/harness/pytest.py
@@ -1,0 +1,365 @@
+import os
+import re
+import shlex
+import sys
+from collections.abc import Generator
+from io import TextIOWrapper
+from typing import Any
+
+import pexpect
+import pytest
+from pytest import Function, TestReport
+from trunner.ctx import TestContext
+from trunner.dut import Dut
+from trunner.text import bold, green, red, yellow
+from trunner.types import Status, TestResult
+
+
+def extract_reason_from_report(report: TestReport) -> str:
+    """Extracts and formats the skip or xfail reason from a test report"""
+    reason = ""
+    if isinstance(report.longrepr, tuple):
+        reason = report.longrepr[2].replace("Skipped: ", "")
+    elif hasattr(report, "wasxfail"):
+        reason = report.wasxfail
+        reason = f"XFAIL: {reason}" if reason else "XFAIL"
+    return reason.strip()
+
+
+def get_status(report: TestReport) -> str:
+    """Converts a PyTest report status into a formatted trunner status string"""
+    was_xfail = hasattr(report, "wasxfail")
+    if report.failed and report.when != "call":
+        return red("ERROR")
+
+    if report.passed and not was_xfail:
+        return green("PASSED")
+    if report.passed and was_xfail:
+        return yellow("XPASS")
+    if report.failed:
+        return red("FAILED")
+    if report.skipped:
+        return yellow("XFAIL" if was_xfail else "SKIPPED")
+
+    return bold("UNKNOWN")
+
+
+def get_str_attr(obj: object, attribute: str) -> str:
+    """Safely retrieves an object attribute and ensures it is returned as a clean string"""
+    raw_attr = getattr(obj, attribute, "")
+    raw_attr = "" if raw_attr is None else raw_attr
+
+    attribute_str = raw_attr.decode("utf-8", "ignore") if isinstance(raw_attr, bytes) else str(raw_attr)
+    return attribute_str.replace("\r", "")
+
+
+class PytestLogInterceptorPlugin:
+    """Plugin that intercepts and modifies the log output
+
+    It modifies only the raw PyTest output, not the live
+    output coming from the DUT, etc.
+    """
+    def __init__(self, ctx: TestContext) -> None:
+        self._ctx = ctx
+        self._screen_writer = None
+        self._sink: TextIOWrapper | None = None
+        self._last_handled_state: tuple[Any, Any, Any] | None = None
+
+    @pytest.hookimpl(trylast=True)
+    def pytest_configure(self, config: Any) -> None:
+        terminal = config.pluginmanager.get_plugin("terminalreporter")
+        if not terminal:
+            return
+
+        if self._ctx.stream_output:
+            self._screen_writer = type(terminal._tw)(sys.stdout)
+
+        self._sink = open(os.devnull, "w")
+        terminal._tw._file = self._sink
+
+        config.option.disable_warnings = True
+        config.option.no_summary = True
+        terminal.summary_stats = lambda: None
+
+    @pytest.hookimpl(trylast=True)
+    def pytest_unconfigure(self, config: Any) -> None:
+        if self._sink is not None:
+            self._sink.close()
+            self._sink = None
+
+    @pytest.hookimpl(tryfirst=True)
+    def pytest_runtest_logstart(self, nodeid: str) -> None:
+        """Manually print start header to screen"""
+        if self._screen_writer is None:
+            return
+
+        self._flush_and_ensure_newline()
+        self._screen_writer.write(f"\t{bold(nodeid)}...")
+        self._screen_writer.write("\n")
+
+    @pytest.hookimpl(tryfirst=True)
+    def pytest_runtest_logreport(self, report: TestReport) -> None:
+        """Manually print status to screen."""
+        should_print = (report.when == "call") or (not report.passed)
+
+        if not should_print or self._screen_writer is None:
+            return
+
+        status = get_status(report)
+        reason = extract_reason_from_report(report)
+
+        self._flush_and_ensure_newline()
+
+        self._screen_writer.write(f"\t{bold(report.nodeid)} ")
+        self._screen_writer.write(status)
+
+        if reason:
+            self._screen_writer.write(f" ({reason})")
+
+        self._screen_writer.write("\n")
+
+    def _flush_and_get_last_char(self, pexpect_p: Any) -> str:
+        drained_last_char = ""
+        try:
+            while True:
+                chunk = pexpect_p.read_nonblocking(size=4096, timeout=0)
+                if not chunk:
+                    break
+
+                if isinstance(chunk, bytes):
+                    drained_last_char = chunk[-1:].decode("utf-8", "ignore")
+                else:
+                    drained_last_char = chunk[-1:]
+
+                # NOTE: The buffer appending operations below do not cause logfile duplication/
+                # The data is manually placed into the internal pexpect buffer, subsequent expect()
+                # calls omit the OS buffer pull operation that triggers the logfile write.
+                if isinstance(pexpect_p.buffer, bytes) and isinstance(chunk, str):
+                    pexpect_p.buffer += chunk.encode("utf-8", "ignore")
+                elif isinstance(pexpect_p.buffer, str) and isinstance(chunk, bytes):
+                    pexpect_p.buffer += chunk.decode("utf-8", "ignore")
+                else:
+                    pexpect_p.buffer += chunk
+        except (pexpect.TIMEOUT, pexpect.EOF):
+            pass
+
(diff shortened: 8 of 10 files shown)


mateusz-bloch (commented):
Only quick look.

I’m not sure what the expected behaviour were, but wouldn’t it be better to parse specific failures per test case? For example if there were hundreds of test cases, this could be more readable, and we would still have the full output available via -S. @damianloew 

Image

damianloew (commented):
First suggestions, just to order the changes first

trunner/harness/pytest.py
```diff
@@ -0,0 +1,124 @@
...
+        exit_code = pytest.main(cmd_args, plugins=[bridge_plugin, log_plugin])
+    except Exception as e:
+        result.fail_harness_exception(e)
+        return result
+
+    if any(sub.status == Status.FAIL for sub in result.subresults):
+        result.status = Status.FAIL
+    elif all(sub.status == Status.SKIP for sub in result.subresults) and result.subresults:
+        result.status = Status.SKIP
+    elif not result.subresults and exit_code != 0:
```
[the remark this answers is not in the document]
lukkrusz: While I'm on it, in a case when the teardown fails, do we want to treat the entire case as FAIL or just the ultimate test result?

One could argue that the test case passed but it was a post-test operation that failed.
As is now, in case of such a failure, the full pytest trace is being released like on the provided screenshot

trunner/harness/pytest.py
```diff
@@ -0,0 +1,124 @@
...
+
+    bridge_plugin = PytestBridgePlugin(dut, ctx, result, kwargs)
+    log_plugin = PytestLogCapturePlugin(ctx.stream_output)
+
+    cmd_args = [
+        str(test_path),
+        *options,
+        "-v",
+        "-s",
+        "--tb=no",
```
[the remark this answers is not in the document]
damianloew: Sorry about misleading screenshot - of course the overall result should be `FAIL`.

sample/pytest/test-pytest.py
```diff
@@ -0,0 +1,141 @@
...
+    def test_no_ok_in_log(self, fake_dut_class):
+        assert "OK" not in fake_dut_class.get_log()
+
+    def test_send_hello(self, fake_dut_class):
+        fake_dut_class.send("hello")
+        assert "OK" in fake_dut_class.get_log()
+
+    def test_send_world_check_for_2_OKs(self, fake_dut_class):
+        fake_dut_class.send("hello")
+        assert fake_dut_class.get_log().count("OK") == 2
```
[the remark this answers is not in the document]
lukkrusz: We are not actually testing anything. The purpose of the sample/pytest is to showcase the usage of our newly added support.

trunner/harness/pytest.py
```diff
@@ -0,0 +1,124 @@
...
+        if report.failed:
+            status = Status.FAIL
+        elif report.skipped and not hasattr(report, "wasxfail"):
+            status = Status.SKIP
+
+        self._result.add_subresult(
+            subname=report.nodeid.split("::")[-1],
+            status=status,
+            msg=""
+        )
```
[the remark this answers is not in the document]
lukkrusz: Somewhat resolved by implementing traceback in case of failure

trunner/harness/pytest.py
```diff
@@ -0,0 +1,194 @@
...
+        config.option.no_summary = True
+        terminal.summary_stats = lambda: None
+
+        if self._suppress:
+            terminal._tw._file = self.buffer
+
+    def get_logs(self) -> str:
+        """Get the full, unsuppressed pytest log
+        """
+        return self.buffer.getvalue()
```
nalajcie: does it work if we're streaming? (we're attaching the `buffer` only if `_suppress == true`

sample/pytest/test.yaml
```diff
@@ -0,0 +1,8 @@
+test:
+    tests:
+        - name: pytest
+          type: pytest
+          kwargs:
+              path: phoenix-rtos-tests/sample/pytest/test-pytest.py
```
nalajcie: IMHO this should be relative to the test.yaml, not to the CWD we expect the end user to run `trunner` from.

Maybe it should be a custom field for `pytest` test type (handled in a special way when parsing the config)?

nalajcie (changes requested):
Pointing out some issues in the current code. Please note that the commit list is too verbose, the final history should not include fixes to your own code (added within the same PR) - I would expect 2-3 commits in the final history. 

Adding separate commits after review might be beneficial, for tracking review progress, but they should be squashed sooner than later.

trunner/harness/pytest.py
```diff
@@ -0,0 +1,194 @@
...
+                "msg": f"{error_msg}",  # Newline for improved readability
+            }
+            if fixture_name != self._unknown_fixture_str:
+                # It is assumed that only a fixture that fails at startup will
+                # be marked as 'broken', thus disabling all test-cases using it
+                self._broken_fixtures.add(fixture_name)
+            return
+        if report.skipped:
+            skip_msg = self._extract_reason(report)
+            self._test_outcomes[report.nodeid] = {"status": Status.SKIP, "msg": skip_msg}
```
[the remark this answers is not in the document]
lukkrusz: Pytest supports running tests in parallel and so I wanted to support it here as well.
However, we decided that it isn't necessary to support now, therefore I'll remove the use of dictionary from here, so that the code doesn't appear overengineered.

damianloew (changes requested):
Suggestions for the code before the recent force push. I haven't seen much unnecessary code, which is good - changes are quite clear for now

trunner/types.py
```diff
@@ -388,9 +388,9 @@ def __str__(self) -> str:
                 )
             # multi-line message that does not exceed width limit
             elif remaining_lines:
-                out += ": " + first_line + "\n" + "\n".join(
+                out += ":\n" + "\n".join(
```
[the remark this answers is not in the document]
lukkrusz: We will create a separate issue to apply this change.

trunner/harness/pytest.py
```diff
@@ -0,0 +1,233 @@
...
+    options = shlex.split(kwargs.get("options", ""))
+
+    bridge_plugin = PytestBridgePlugin(dut, ctx, result, kwargs)
+    log_plugin = PytestLogCapturePlugin(ctx.stream_output)
+
+    cmd_args = [str(test_path), *options, "-s"]
+
+    try:
+        exit_code = pytest.main(cmd_args, plugins=[bridge_plugin, log_plugin])
+    except Exception as e:
```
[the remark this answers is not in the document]
lukkrusz: Hmmm, by what I can see, pytest handles the exceptions quite well, see below for UnicodeDecodeError:
image

EDIT: I will look into how different the exception handling is compared to, e.g., PyHarness psh test

trunner/harness/pytest.py
```diff
@@ -0,0 +1,194 @@
...
+
+    cmd_args = [str(test_path), *options, "-v", "-s"]
+
+    try:
+        exit_code = pytest.main(cmd_args, plugins=[bridge_plugin, log_plugin])
+    except Exception as e:
+        result.fail_harness_exception(e)
+        return result
+
+    if any(sub.status == Status.FAIL for sub in result.subresults):
```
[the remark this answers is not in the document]
lukkrusz: A new PR[#449](https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/449) implementing this feature is now open

trunner/types.py
```diff
@@ -89,6 +89,28 @@ def __lt__(self, other):
         if self.__class__ is other.__class__:
             return self.value < other.value
         return NotImplemented
+    
+
+class TestType(str, Enum):
```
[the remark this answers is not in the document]
lukkrusz: It's a subtask now

sample/pytest/test-pytest.py
```diff
@@ -0,0 +1,126 @@
...
+def test_always_fails():
+    assert False, "Always Fail"
+
+
+@pytest.mark.skip(reason="Should not fail because it's skipped")
+def test_always_skip():
+    assert False
+
+
+@pytest.mark.xfail(reason="Always fails, as expected")
```
damianloew: This case will just show another option (probably even better one) when sth is not working as expected yet. Still, we want to keep `#x issue`  format.
```suggestion
@pytest.mark.xfail(reason="#x issue")
```

sample/pytest/test-psh.py
```diff
@@ -0,0 +1,46 @@
...
+    ("-illegal_arg", ".*", 1),
+]
+
+
+@pytest.mark.parametrize("msg, resp, code", echo_resp_code_list)
+def test_echo(pexpect_psh, msg, resp, code):
+    psh.assert_cmd(
+        pexpect_psh,
+        f"echo {msg}",
+        expected=f"{resp}\r+\n",
```
damianloew:
```suggestion
        expected=fr"{resp}\r+\n",
```

damianloew (changes requested):
Looking only at tests for now

sample/pytest/test-pytest.py
```diff
@@ -0,0 +1,126 @@
...
+    assert global_session_resource["status"] == "busy"
+
+
+def test_subresult_time_by_delaying():
+    time.sleep(2)
+    assert True
+
+
+def test_send_hello(pexpect_bin):
+    pexpect_bin.sendline("hello")
```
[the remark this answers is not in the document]
lukkrusz: I decided for the app to either recognise the commands or not, and so this test case is now redundant and will be removed.

sample/pytest/test.py
```diff
@@ -0,0 +1,126 @@
...
+    test_val = 0
+
+    def setup_method(self):
+        TestClass.test_val = 5
+
+    def teardown_method(self):
+        TestClass.test_val = 0
+
+    def test_send_hello_from_class(self, pexpect_bin):
+        pexpect_bin.sendline("hello from class")
```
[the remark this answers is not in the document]
lukkrusz: As per my previous response, I will create a special command for this test case

trunner/harness/pytest.py
```diff
@@ -0,0 +1,282 @@
...
+        self._screen_writer.write(status)
+
+        if reason:
+            self._screen_writer.write(f" ({reason})")
+
+        self._screen_writer.write("\n")
+
+    def _shell_prompt_present(self) -> bool:
+        target = self._ctx.target
+        shell_prompt = "(psh)% "
```
lukkrusz: This needs a deeper look into; it solves pytest's output appearing on the same line as prompt but it seems too hacky and suffers from exceptions like this:
Image

We're dealing with a live output, which is tricky

trunner/harness/pytest.py
```diff
@@ -0,0 +1,120 @@
...
+        self._result.add_subresult(
+            subname=report.nodeid.split("::")[-1], 
+            status=status, 
+            msg=msg
+        )
+
+
+def pytest_harness(dut: Dut, ctx: TestContext, result: TestResult, **kwargs) -> TestResult:
+    test_path = ctx.project_path / kwargs.get("path")
+    options = kwargs.get("options", "").split()
```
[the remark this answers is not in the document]
damianloew: After some thinking, we can bring it back - it's helpful when we want to add a custom argument for a test - for example to skip initialization phase, which is not always necessary.

trunner/config.py
```diff
@@ -303,6 +311,17 @@ def _preprocess_main_config(self, config: dict):
...
+    def _get_absolute_path(self, path: str) -> Path:
+        p = Path(path)
+
+        if not p.is_absolute():
+            # If path is not absolute then it must be relative to the directory of yaml
+            absolute_path = self.yaml_path.parent / p
+            if not absolute_path.is_absolute():
+                raise ParserError("yml path is not absolute!")
+
+        return absolute_path
```
damianloew: NameError when path is already absolute: the absolute_path variable is only assigned inside the if not p.is_absolute() branch, but return absolute_path is unconditional. Calling _get_absolute_path with an absolute path (e.g. a YAML that uses /abs/path.py as script) raises UnboundLocalError: local variable 'absolute_path' referenced before assignment.

```suggestion
            p = self.yaml_path.parent / p
            if not p.is_absolute():
        return p
```

trunner/config.py
```diff
@@ -95,6 +90,19 @@ def _parse_pyharness(self, config: dict):
 
         self.test.harness = PyHarness(self.ctx.target.dut, self.ctx, harness_fn, self.test.kwargs)
 
+    def _parse_pytest(self, config):
```
damianloew:
```suggestion
    def _parse_pytest(self, config: dict) -> None:
```
The return type should be added to the rest of the methods in this class too.

trunner/config.py
```diff
@@ -95,6 +90,19 @@ def _parse_pyharness(self, config: dict):
...
+        script = config.get("script", None)
```
damianloew: dict.get() returns None by default — passing None explicitly is redundant.

```suggestion
        script = config.get("script")
```

trunner/harness/__init__.py
```diff
@@ -46,4 +47,5 @@
     "PloImageProperty",
     "PloJffsImageProperty",
     "unity_harness",
+    "pytest_harness"
```
damianloew: to be consistent with the previous approach - more readable diffs when adding new elements.
```suggestion
    "pytest_harness",
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,302 @@
...
+        return yellow("XPASS")
+    if report.failed:
+        return red("FAILED")
+    if report.skipped:
+        return yellow("XFAIL" if was_xfail else "SKIPPED")
+
+    return bold("UNKNOWN")
+
+
+def get_str_attr(obj: object, attribute: str):
```
damianloew:
```suggestion
def get_str_attr(obj: object, attribute: str) -> str:
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,302 @@
...
+class PytestLogInterceptorPlugin:
+    """Plugin that intercepts and modifies the log output
+
+    It modifies only the raw PyTest output, not the live
+    output coming from the DUT, etc.
+    """
+    def __init__(self, ctx: TestContext) -> None:
+        self._ctx = ctx
+        self._screen_writer = None
+        self._first_test_ran = False
```
damianloew: _first_test_ran is assigned in __init__ but never read or written anywhere else in the class. Dead code.
```suggestion
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,302 @@
...
+        config.option.no_summary = True
+        terminal.summary_stats = lambda: None
+
+    @pytest.hookimpl(tryfirst=True)
+    def pytest_runtest_logstart(self, nodeid: str) -> None:
+        """Manually print start header to screen.
+
+        Retains pexpect post-initalization buffer for the first
+        test-case, based on an assumption that pexpect object is
+        spawned only once within a test-suite.
```
damianloew: It looks like it doesn't work like that:

Image

And I like the current behavior - we see what happened during the initialization and what during the test  case execution. Probably the docstring is outdated (?)

trunner/harness/pytest.py
```diff
@@ -0,0 +1,302 @@
...
+
+        if call.when != "call" or not report.failed or not call.excinfo:
+            return
+
+        exc_instance = call.excinfo.value
+        report_add_info = None
+
+        before = get_str_attr(self._dut, "before")
+
+        if isinstance(exc_instance, UnicodeDecodeError) and before:
```
damianloew: how about pexpect.TIMEOUT and pexpect.EOF?

trunner/harness/pytest.py
```diff
@@ -0,0 +1,302 @@
...
+            return decorator_match.group(1)
+        return cls._unknown_fixture_str
+
+
+def pytest_harness(dut: Dut, ctx: TestContext, result: TestResult, **kwargs) -> TestResult:
+    if "script" not in kwargs:
+        result.fail(msg="missing `script` from test configuration!")
+        return result
+
+    test_path = ctx.project_path / kwargs["script"]
```
damianloew: kwargs["script"] is already an absolute Path (guaranteed by _parse_pytest calling _get_absolute_path). Joining it onto ctx.project_path via / is a silent no-op in Python (an absolute right operand replaces the left), which is misleading — it looks like a path construction but actually does nothing. Use the value directly.


```suggestion
    test_path = kwargs["script"]
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,302 @@
...
+        exit_code = pytest.main(cmd_args, plugins=[bridge_plugin, log_plugin])
+    except Exception:
+        result.fail_unknown_exception()
+        return result
+
+    accepted_codes = (pytest.ExitCode.OK, pytest.ExitCode.TESTS_FAILED, pytest.ExitCode.NO_TESTS_COLLECTED)
+
+    if exit_code not in accepted_codes:
+        result.status = Status.FAIL
+        result.msg = f"Pytest execution failed (Exit Code {exit_code})."
```
damianloew: Direct mutation of result.status and result.msg bypasses the TestResult.fail() API used everywhere else.

```suggestion
    result.fail(msg=f"Pytest execution failed (Exit Code {exit_code}).")
```

sample/pytest/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+    yield p
+
+    p.sendline("EXIT")
+    p.expect_exact("[Success!]")
+
+
+@pytest.fixture(scope="session")
+def pexpect_psh(dut):
+    p = dut.pexpect_proc
+    exec_cmd = "/bin/psh" if trunner.ctx.target.rootfs else "sysexec psh"
```
damianloew: PytestBridgePlugin already injects ctx as a session-scoped fixture - use it instead.

```suggestion
def pexpect_psh(dut, ctx):
    p = dut.pexpect_proc
    exec_cmd = "/bin/psh" if ctx.target.rootfs else "sysexec psh"
```

sample/pytest/test.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+@pytest.mark.xfail(reason="#x issue")
+def test_always_xfail():
+    assert False
+
+
+@pytest.mark.xfail()
+def test_always_xpass():
+    assert True
```
damianloew:
```suggestion
```

Should be moved to docs - it only presents how we can use this pytest support in our tests, it doesn't test nothing - to test that we would have to add test runner tests which would assert the correct test results in such cases. For now, I'd remove that.

sample/pytest/test.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def test_global_resource_ready_id_999(global_session_resource):
+    assert global_session_resource["status"] == "ready"
+    assert global_session_resource["id"] == 999
+
+
+def test_subresult_time_by_delaying():
+    time.sleep(0.5)
+    assert True
```
damianloew:
```suggestion
# TODO: move this test to trunner tests which would assert the execution time for the test case in xml
def test_subresult_time_by_delaying():
    time.sleep(0.5)
    assert True
```

or sth similar

sample/pytest/test.py
```diff
@@ -0,0 +1,102 @@
...
+    pexpect_bin.sendline(cmd)
+    pexpect_bin.expect_exact(f"{exp}\r\n")
+
+
+@pytest.mark.usefixtures("pexpect_bin")
+class TestClass:
+    test_val = 0
+
+    def setup_method(self):
+        TestClass.test_val = 5
```
damianloew: setup_method and teardown_method referenced the class by its literal name (TestClass.test_val). If the class were ever subclassed, the base-class name is hardcoded and subclass state would not be modified. Use self.__class__ to be subclass-safe and match Python convention.

```suggestion
        self.__class__.test_val = 5
```

sample/pytest/test.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def teardown_method(self):
+        TestClass.test_val = 0
```
damianloew:
```suggestion
        self.__class__.test_val = 0
```

damianloew (changes requested):

trunner/harness/pytest.py
lukkrusz: Yes, along with earlier `_first_test_ran`, this explained the now abandoned approach of keeping the initialisation logs until the first test ran for rare occurences of a missing newline or colliding buffers like this [(observed on armv7m7-imxrt106x-evk target)](https://github.com/phoenix-rtos/phoenix-rtos-tests/actions/runs/22722709472/job/65889554740?pr=445):

Image

This code will now be removed

damianloew: As discussed f2f, the situation where the system output is visible before the first test case header is acceptable (the test setup stage is visible then), but you can check whether it's possible to print the header in the new line in such scenario while keeping the code clean. If not, leave the appropriate comment.

trunner/config.py
lukkrusz: Following a F2F discussion, the wider effort to add the return types to other function definitions will be a part of a separate PR. This suggestion, however, will be applied to the additions made in this PR

trunner/config.py
lukkrusz: Done

trunner/config.py
lukkrusz: Done

trunner/harness/__init__.py
lukkrusz: Done

trunner/harness/pytest.py
lukkrusz: Done

trunner/harness/pytest.py
lukkrusz: Done

trunner/harness/pytest.py
lukkrusz: I included them now too

trunner/harness/pytest.py
lukkrusz: Done

trunner/harness/pytest.py
lukkrusz: Good catch, fixed now

sample/pytest/conftest.py
lukkrusz: Done

sample/pytest/test.py
lukkrusz: Following a F2F discussion, we established that only the tests testing the trunner framework (pexpect, ctx and dut fixtures) shall remain in the sample test suite. The test cases showcasing the use of pytest will have their place in docs

sample/pytest/test.py
lukkrusz: Done

trunner/harness/pytest.py
```diff
@@ -0,0 +1,309 @@
...
+        after_str = after.decode("utf-8", "ignore") if isinstance(after, bytes) else str(after) if after else ""
+        buffer_str = buffer.decode("utf-8", "ignore") if isinstance(buffer, bytes) else str(buffer) if buffer else ""
+
+        last_text_on_screen = buffer_str if buffer_str else after_str
+
+        # if state did not change there is nothing to do
+        if getattr(self, "_last_handled_state", None) == current_state or not last_text_on_screen:
+            return
+
+        self._last_handled_state = current_state
```
damianloew: _last_handled_state was used as a lazily-initialized instance attribute accessed via getattr(..., None). Good to be declared in __init__ for explicit class interface.

trunner/harness/pytest.py
```diff
@@ -0,0 +1,309 @@
...
+    def pytest_configure(self, config: Any) -> None:
+        terminal = config.pluginmanager.get_plugin("terminalreporter")
+        if not terminal:
+            return
+
+        if self._ctx.stream_output:
+            self._screen_writer = type(terminal._tw)(sys.stdout)
+
+        # Redirects PyTest output to /dev/null
+        terminal._tw._file = open(os.devnull, "w")
```
damianloew: File handle opened with open(os.devnull, 'w') was never closed — the fd was leaked for the lifetime of the process. Store as self._sink and close in pytest_unconfigure, which pytest calls after the session ends.
```suggestion
         self._sink = open(os.devnull, "w")
         terminal._tw._file = self._sink
```


```
     def pytest_unconfigure(self, config: Any) -> None:
         if self._sink is not None:
             self._sink.close()
             self._sink = None
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,309 @@
...
+        dut = getattr(self._ctx.target, "dut", None)
+        pexpect_p = getattr(dut, "pexpect_proc", None)
+        if dut is None or pexpect_p is None:
+            return
+
+        before = getattr(pexpect_p, "before", "")
+        after = getattr(pexpect_p, "after", "")
+        buffer = getattr(pexpect_p, "buffer", "")
+
+        # capture unique set of IDs to monitor the change of state
```
damianloew: Not necessary comment - we use comments only when sth is not conventional.
```suggestion
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,309 @@
...
+        current_state = (id(before), id(after), id(buffer))
+
+        after_str = after.decode("utf-8", "ignore") if isinstance(after, bytes) else str(after) if after else ""
+        buffer_str = buffer.decode("utf-8", "ignore") if isinstance(buffer, bytes) else str(buffer) if buffer else ""
+
+        last_text_on_screen = buffer_str if buffer_str else after_str
+
+        # if state did not change there is nothing to do
```
damianloew: Not necessary comment.
```suggestion
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,309 @@
...
+        if report.when == "call":
+            self._record_call_phase(report, error_msg)
+
+        if report.when == "teardown":
+            self._finalize_and_report(report, error_msg)
+
+    def _record_setup_phase(self, report: TestReport, error_msg: str) -> None:
+        if report.failed:
+            fixture_name = self._extract_broken_fixture(error_msg)
+            self._pending_result = (Status.FAIL, f"{error_msg}")
```
damianloew:
```suggestion
            self._pending_result = (Status.FAIL, error_msg)
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,309 @@
...
+            status = Status.SKIP
+            msg = extract_reason_from_report(report)
+
+        self._pending_result = (status, msg)
+
+    def _finalize_and_report(self, report: TestReport, error_msg: str) -> None:
+        stored = self._pending_result
+        final_status = stored[0]
+        final_msg = stored[1]
+        teardown_msg = error_msg
```
damianloew: Remove unnecessary `teardown_msg = error_msg` intermediate variable and redundant f-string f"{teardown_msg}".
```suggestion
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+class PytestBridgePlugin:
+    """Plugin that bridges the TestRunner and PyTest frameworks.
+
+    Each unit test is run as a TestRunner's subtest, accurately
+    capturing each subresult's runtime and status.
+    """
+    def __init__(self, dut, ctx, result, kwargs):
+        self._dut = dut
```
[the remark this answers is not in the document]
damianloew: It's also mentioned in [PEP 8](https://peps.python.org/pep-0008/#naming-conventions) `Use one leading underscore only for non-public methods and instance variables.`, so we will change it in the whole Test Runner - task has been created for that.

trunner/harness/pytest.py
lukkrusz: Upon re-reviewing this, I agree, and will fix this as suggested

trunner/harness/pytest.py
```diff
@@ -0,0 +1,354 @@
...
+                    pexpect_p.buffer += chunk
+        except (pexpect.TIMEOUT, pexpect.EOF):
+            pass
+
+        return drained_last_char
+
+    def _ensure_newline(self) -> None:
+        """Check if there was a newline before and write it if not.
+
+        This drains pending OS logs so they print continuously, restores them
```
damianloew, line 152: Then it's not only ensure_newline, current function name is misleading. The flushing utility is also crucial here.

Also, I don't like the this statement `restores them so tests don't break`, maybe `restores them so they can be parsed in the incoming test`, or sth like that.

trunner/harness/pytest.py
lukkrusz: Done

trunner/harness/pytest.py
lukkrusz: Done

trunner/harness/pytest.py
lukkrusz: Done, removed some newer comments too

trunner/harness/pytest.py
lukkrusz: Done

trunner/harness/pytest.py
lukkrusz: Done

trunner/harness/pytest.py
```diff
@@ -0,0 +1,354 @@
...
+import pluggy
+import pytest
+from pytest import Function, TestReport
+from trunner.ctx import TestContext
+from trunner.dut import Dut
+from trunner.text import bold, green, red, yellow
+from trunner.types import Status, TestResult
+
+
+def extract_reason_from_report(report: TestReport) -> str:
```
damianloew, line 18: Good to add short docstrings for top-level functions.

sample/pytest/main.c
```diff
@@ -0,0 +1,53 @@
...
+int main(int argc, char **argv)
+{
+	char c = '\0';
+	char buffer[BUF_SZ] = { 0 };
+	int pos = 0;
+	int ret_code = 1;
+
+	printf("[Commence Fake Communication]\n");
+
+	// Run the loop until `EXIT`
```
damianloew: It does not follow our convention, please see https://docs.phoenix-rtos.com/latest/coding/index.html

Please update the whole test according to that rules.

sample/pytest/main.c
```diff
@@ -0,0 +1,53 @@
...
+#define BUF_SZ 256
+
+int process_command(char *input)
+{
+	const char *echo_cmd = "echo ";
+
+	if (strcmp(input, "EXIT") == 0)
+		return 0;
+	else if (strncmp(input, echo_cmd, strlen(echo_cmd)) == 0)
+		printf("%s\n", input + strlen(echo_cmd));
```
damianloew: PS-C-REC-037, PS-C-REC-038, PS-C-REC-039

trunner/harness/pytest.py
```diff
@@ -0,0 +1,354 @@
...
+    @pytest.hookimpl(trylast=True)
+    def pytest_configure(self, config: Any) -> None:
+        terminal = config.pluginmanager.get_plugin("terminalreporter")
+        if not terminal:
+            return
+
+        if self._ctx.stream_output:
+            self._screen_writer = type(terminal._tw)(sys.stdout)
+
+        # Redirects PyTest output to /dev/null
```
damianloew: what comment

trunner/config.py
```diff
@@ -95,6 +90,19 @@ def _parse_pyharness(self, config: dict):
 
         self.test.harness = PyHarness(self.ctx.target.dut, self.ctx, harness_fn, self.test.kwargs)
 
+    def _parse_pytest(self, config: dict) -> None:
+        script = config.get("script")
+        if not script:
+            raise ParserError("pytest `script` name must be provided!")
+        script_path = self._get_absolute_path(script)
+        if script_path.suffix != ".py":
+            raise ParserError("provided `script` is not a python script!")
```
damianloew, line 99: _parse_pytest validates the .py suffix but not whether the file actually exists. Good to add sth like:
```suggestion
            raise ParserError("provided `script` is not a python script!")
        if not script_path.exists():
            raise ParserError(f"pytest script not found: {script_path}")
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,354 @@
...
+
+        drained_last_char = self._drain_and_restore_buffer(pexpect_p)
+        last_char = ""
+
+        if drained_last_char:
+            last_char = drained_last_char
+            self._last_handled_state = (
+                id(getattr(pexpect_p, "before", "")),
+                id(getattr(pexpect_p, "after", "")),
+                id(getattr(pexpect_p, "buffer", "")),
```
damianloew: id() was used to detect whether pexpect state objects changed between calls. This is unreliable: CPython can reuse a memory address for a new object after the old one is garbage-collected, making two distinct states appear identical and causing _ensure_newline to silently skip writing a needed newline. Compare actual attribute values instead. Note that type hint should be also updated.

```suggestion
                getattr(pexpect_p, "before", ""),
                getattr(pexpect_p, "after", ""),
                getattr(pexpect_p, "buffer", ""),
...
            current_state = (before, after, buffer)
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,354 @@
...
+        report_add_info = None
+
+        before = get_str_attr(self._dut, "before")
+
+        if isinstance(exc_instance, (UnicodeDecodeError, pexpect.TIMEOUT, pexpect.EOF)) and before:
+            report_add_info = bold("OUTPUT CAUGHT BEFORE EXCEPTION:")
+            report_add_info += before
+
+        if report_add_info:
+            report.longrepr = str(report.longrepr) + "\n\n" + report_add_info
```
damianloew: hookwrapper=True is deprecated since pytest 8.1 (we will probably update the version in the near future), please check the suggested approach.

```suggestion
    @pytest.hookimpl(wrapper=True)
    def pytest_runtest_makereport(self, item: Function, call: pytest.CallInfo) -> Generator[None, None, TestReport]:
        """Post-process PyTest's report making

        Inject output caught before exception for a more detailed report.

        Currently, TestResult failing methods fail the entire test with the
        most recent exception, with no regard to individual subresults.

        TODO: Modify TestResult failing methods to support single subresult
        exceptions and use them here to avoid reimplementation.
        """
        report: TestReport = yield

        if call.when != "call" or not report.failed or not call.excinfo:
            return report

        exc_instance = call.excinfo.value
        report_add_info = None

        before = get_str_attr(self._dut, "before")

        if isinstance(exc_instance, (UnicodeDecodeError, pexpect.TIMEOUT, pexpect.EOF)) and before:
            report_add_info = bold("OUTPUT CAUGHT BEFORE EXCEPTION:")
            report_add_info += before

        if report_add_info:
            report.longrepr = str(report.longrepr) + "\n\n" + report_add_info
        return report
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,354 @@
...
+        error_msg = getattr(report, "longreprtext", str(report.longrepr)) if report.longrepr else ""
+
+        if report.when == "setup":
+            self._record_setup_phase(report, error_msg)
+
+        if report.when == "call":
+            self._record_call_phase(report, error_msg)
+
+        if report.when == "teardown":
+            self._finalize_and_report(report, error_msg)
```
damianloew: report.when can only be one value at a time, so the three conditions are mutually exclusive - use elifs

trunner/harness/pytest.py
```diff
@@ -0,0 +1,354 @@
...
+
+        self._screen_writer.write(f"\t{bold(report.nodeid)} ")
+        self._screen_writer.write(status)
+
+        if reason:
+            self._screen_writer.write(f" ({reason})")
+
+        self._screen_writer.write("\n")
+
+    def _drain_and_restore_buffer(self, pexpect_p: Any) -> str:
```
damianloew: IMHO more readable name in this case
```suggestion
    def _flush_and_get_last_char(self, pexpect_p: Any) -> str:
```

trunner/harness/pytest.py
```diff
@@ -0,0 +1,354 @@
...
+                if not chunk:
+                    break
+
+                if isinstance(chunk, bytes):
+                    drained_last_char = chunk[-1:].decode("utf-8", "ignore")
+                else:
+                    drained_last_char = chunk[-1:]
+
+                if isinstance(pexpect_p.buffer, bytes) and isinstance(chunk, str):
+                    pexpect_p.buffer += chunk.encode("utf-8", "ignore")
```
damianloew, line 139: Please prove in the comment that it won't cause the data duplication in logfile (streamed in -S option in our case) - it's not so obvious and needs explanation. We could also use expect for that, but `read_nonblocking` should be faster.

damianloew (changes requested):
Some new findings, hopefully we're close to merging

trunner/harness/pytest.py
lukkrusz: Done

trunner/harness/pytest.py
lukkrusz: I hope that the newly added comment offers the desired level of clarity when it comes to logfile duplication concern.

trunner/harness/pytest.py
lukkrusz: Agreed, it's done

trunner/harness/pytest.py
lukkrusz: Done

trunner/harness/pytest.py
lukkrusz: It works, thanks for this suggestion

trunner/harness/pytest.py
lukkrusz: Done

trunner/config.py
lukkrusz: Done, thanks

sample/pytest/main.c
lukkrusz: The test should be fully compliant now

damianloew (dismissed):
LGTM

@nalajcie if you don't mind, I'd like to merge this PR

trunner/config.py
```diff
@@ -69,13 +70,7 @@ def _parse_pyharness(self, config: dict):
         if path is None:
             raise ParserError("test is of type 'harness' but there is no \"harness\" keyword")
 
-        path = Path(path)
-
-        if not path.is_absolute():
```
nalajcie, line 74: are we intentionally dropping support for using absolute path to specify the harness? (I don't see it mentioned in any commit message)

nalajcie (commented):
ok for me, I don't have time for full review, but quickly skimming through changes in common code I see that the current approach seems well-made

trunner/config.py
lukkrusz: Well the support wasn't dropped but it wasn't enforced either. I reintroduced an `.is_absolute()` check.
`.parent / path` works for both, relative and absolute paths.

damianloew (approved):
Only one additional defensive check added in the code since the last approve

PR #470, opened 2026-03-27
libc: add tests for readdir behaviour at end-of-stream

oI0ck:
We want to require filesystems to return -ENOENT predictably when approaching end of directory stream in readdir().

This was not the case in JFFS2, hence this test to ensure conformance in the future.

JIRA: RTOS-1088

Related to [phoenix-rtos-tests#163](https://github.com/phoenix-rtos/phoenix-rtos-filesystems/pull/163)

### DIFF
--- a/libc/unistd/dir.c
+++ b/libc/unistd/dir.c
@@ -1,0 +1,184 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Tests for *dir functions behaviour
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Michal Lach
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdlib.h>
+#include <dirent.h>
+#include <errno.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/msg.h>
+
+#include "unity.h"
+#include "unity_fixture.h"
+
+#define TEST_DIR      "/test_dir"
+#define TEST_FILENAME "/test_file"
+
+
+TEST_GROUP(dirent);
+
+
+TEST_SETUP(dirent)
+{
+	TEST_ASSERT_EQUAL_INT(0, mkdir(TEST_DIR, 0666));
+
+	FILE *empty = fopen(TEST_FILENAME, "w+");
+	TEST_ASSERT_NOT_NULL(empty);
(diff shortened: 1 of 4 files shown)


libc/unistd/dir.c
```diff
@@ -0,0 +1,61 @@
...
+
+	DIR *d = opendir(TEST_DIR);
+	TEST_ASSERT_NOT_NULL(d);
+	/* clang-format off */
+	errno = 0;
+	while (readdir(d));
+	/* clang-format on */
+	saved = errno;
+
+	TEST_ASSERT_EQUAL_INT(0, saved);
```
Darchiv, line 70: How does this test test for `-ENOENT`?

oI0ck: This is linked to https://github.com/phoenix-rtos/libphoenix/pull/433.

There, a check for `-ENOENT` is introduced in `readdir()` to not set `errno`, as this means that the filesystem server reached end of child directory stream. If `readdir()` terminates, it will not set `errno` in two cases.
Either if the filesystem server did not return an error on `mtReaddir` or, if it returned an `-ENOENT` which in this case should signal reaching end of directory stream. In all other cases, `errno` shall be set, as the filesystem server returned an error, which is not `-ENOENT.`

oI0ck: This test fails without https://github.com/phoenix-rtos/phoenix-rtos-filesystems/pull/163 on JFFS2 platforms, as it returns `-1` on end of directory stream, this then does not pass the aforementioned check and sets the errno incorrectly to `-1`, which on our system is `-EPERM`

libc/unistd/dir.c
```diff
@@ -0,0 +1,61 @@
...
+
+TEST(readdir, enoent)
+{
+	int saved;
+
+	DIR *d = opendir(TEST_DIR);
+	TEST_ASSERT_NOT_NULL(d);
+	/* clang-format off */
+	errno = 0;
+	while (readdir(d));
```
julianuziemblo: ik MISRA doesn't apply here, but I'd still suggest `while (readdir(d) != NULL);` for readability

oI0ck: Done

oI0ck:
As this pull request now includes tests for https://github.com/phoenix-rtos/libphoenix/pull/433, and the branch is named differently, hence the CI does not pass, I'm closing this PR in favour of https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/477

PR #471, opened 2026-03-31
port: add tests for register and devs

julianuziemblo:
Add tests for portRegister and portUnregister syscalls, create_dev() and destroy_dev() functions.

YT: CI-651

## Motivation and Context
Depends-On: libphoenix:julianuziemblo/remove_dev
Depends-On: phoenix-rtos-kernel:julianuziemblo/port-unregister

## Types of changes

## How Has This Been Tested?

- [x] New test added: *this*
- [x] Tested by hand on: `ia32-generic-qemu`, `armv7m7-imxrt117x-evk`

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work:
    - https://github.com/phoenix-rtos/libphoenix/pull/466
    - https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/750

### DIFF
--- a/port/test_register.c
+++ b/port/test_register.c
@@ -1,0 +1,172 @@
+/*
+ * Phoenix-RTOS
+ *
+ * test_register
+ *
+ * tests for `portRegister` and `portUnregister` syscalls
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Julian Uziembło
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/msg.h>
+#include <unistd.h>
+
+#include "unity_fixture.h"
+
+#define LETTERS_IN_ALPHABET ('z' - 'a' + 1)
+#define NDEVS               (LETTERS_IN_ALPHABET * LETTERS_IN_ALPHABET)
+#define DEVNAMSIZ           2
+
+#define SIMPLE_PORT_NAME "/simple_test"
+
+#define TEST_PORT_REGISTER(name, oid) \
+	do { \
+		uint32_t _port = (oid).port; \
+		id_t _id = (oid).id; \
+		oid_t _odev, _ofil; \
+		TEST_ASSERT_EQUAL_MESSAGE(0, portRegister(_port, (name), &(oid)), (name)); \
+		TEST_ASSERT_EQUAL_MESSAGE(_port, (oid).port, (name)); \
+		TEST_ASSERT_EQUAL_MESSAGE(_id, (oid).id, (name)); \
+		TEST_ASSERT_EQUAL_MESSAGE(0, lookup(name, &_ofil, &_odev), name); \
+	} while (0)
+
+
+static uint32_t port;
+
+
+TEST_GROUP(test_register);
+
+TEST_SETUP(test_register)
+{
+}
+
+TEST_TEAR_DOWN(test_register)
+{
+	/* in any case, unregister the SIMPLE_PORT_NAME */
+	portUnregister(SIMPLE_PORT_NAME);
+}
+
+TEST(test_register, simple)
+{
+	const id_t id = 0;
+	oid_t oid = { .port = port, .id = id }, ofil, odev;
+	TEST_PORT_REGISTER(SIMPLE_PORT_NAME, oid);
+
+	TEST_ASSERT_EQUAL(0, portUnregister(SIMPLE_PORT_NAME));
+
+	TEST_ASSERT_LESS_THAN(0, lookup(SIMPLE_PORT_NAME, &ofil, &odev));
+
+	TEST_ASSERT_EQUAL(-ENOENT, portUnregister(SIMPLE_PORT_NAME));
+}
+
+TEST(test_register, register_existing)
+{
+	const id_t id = 0;
+	oid_t oid = { .port = port, .id = id };
+	TEST_PORT_REGISTER(SIMPLE_PORT_NAME, oid);
+
+	TEST_ASSERT_EQUAL(-EEXIST, portRegister(port, SIMPLE_PORT_NAME, &oid));
+
+	TEST_ASSERT_EQUAL(0, portUnregister(SIMPLE_PORT_NAME));
+
+	TEST_ASSERT_EQUAL(-ENOENT, portUnregister(SIMPLE_PORT_NAME));
+}
+
+TEST(test_register, bad_port)
+{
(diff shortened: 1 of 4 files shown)


port/test_register.c
```diff
@@ -0,0 +1,164 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <sys/msg.h>
+#include <unistd.h>
+
+#include "unity_fixture.h"
+
+#define LETTERS_IN_ALPHABET ('z' - 'a' + 1)
+#define NDEVS               (LETTERS_IN_ALPHABET * LETTERS_IN_ALPHABET)
+#define DEVNAMSIZ           2
```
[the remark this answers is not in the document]
julianuziemblo: why?

port/test_register.c
```diff
@@ -0,0 +1,164 @@
...
+		}
+
+		char currName[3];
+		currName[0] = firstLetter;
+		currName[1] = (i % LETTERS_IN_ALPHABET) + 'a';
+		currName[2] = '\0';
+		memcpy(devs[i].name, currName, sizeof(currName));
+
+		devs[i].oid.port = port;
+		devs[i].oid.id = currName[0] + (LETTERS_IN_ALPHABET * currName[1]);
```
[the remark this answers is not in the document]
julianuziemblo: it doesn't matter that we depend on cwd here as it should work the same regardless

port/test_register.c
```diff
@@ -0,0 +1,164 @@
...
+TEST(test_register, bad_port)
+{
+	const id_t id = 0;
+	oid_t oid = { .port = port, .id = id }, odev;
+
+	/* make sure this port doesn't exist by destroying it */
+	portDestroy(port);
+
+	/* port should still be registered */
+	TEST_ASSERT_EQUAL(0, portRegister(port, SIMPLE_PORT_NAME, &oid));
```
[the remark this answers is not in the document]
julianuziemblo: ports and devs (which we're effectively registering here) are not the same thing, so we should treat them in this way to avoid confusion down the road.

Darchiv (dismissed):

port/test_register.c
```diff
@@ -0,0 +1,179 @@
...
+
+#define TEST_PORT_REGISTER(name, oid) \
+	do { \
+		uint32_t _port = (oid).port; \
+		id_t _id = (oid).id; \
+		oid_t _odev, _ofil; \
+		TEST_ASSERT_EQUAL_MESSAGE(0, portRegister(_port, (name), &(oid)), (name)); \
+		TEST_ASSERT_EQUAL_MESSAGE(_port, (oid).port, (name)); \
+		TEST_ASSERT_EQUAL_MESSAGE(_id, (oid).id, (name)); \
+		TEST_ASSERT_GREATER_OR_EQUAL_MESSAGE(0, lookup(name, &_ofil, &_odev), name); \
```
ziemleszcz: Why OR_GREATER? By the way, I find these kinds of macros hard to read.

port/test_register.c
```diff
@@ -0,0 +1,179 @@
...
+	TEST_PORT_REGISTER(SIMPLE_PORT_NAME, oid);
+
+	TEST_ASSERT_EQUAL(-EEXIST, portRegister(port, SIMPLE_PORT_NAME, &oid));
+
+	TEST_ASSERT_EQUAL(0, portUnregister(SIMPLE_PORT_NAME));
+
+	TEST_ASSERT_EQUAL(-ENOENT, portUnregister(SIMPLE_PORT_NAME));
+}
+
+TEST(test_register, bad_port)
```
ziemleszcz, line 84: Please add a brief explanation describing the purpose of this test.

port/test_register.c
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		char currName[3];
+		currName[0] = firstLetter;
+		currName[1] = (i % LETTERS_IN_ALPHABET) + 'a';
+		currName[2] = '\0';
+		memcpy(devs[i].name, currName, sizeof(currName));
+
+		devs[i].oid.port = port;
+		devs[i].oid.id = currName[0] + (LETTERS_IN_ALPHABET * currName[1]);
```
ziemleszcz: Why not `devs[i].oid.id = i`?

port/test_register.c
```diff
@@ -0,0 +1,179 @@
...
+
+		if (i != 0 && i % LETTERS_IN_ALPHABET == 0) {
+			firstLetter++;
+		}
+
+		char currName[3];
+		currName[0] = firstLetter;
+		currName[1] = (i % LETTERS_IN_ALPHABET) + 'a';
+		currName[2] = '\0';
+		memcpy(devs[i].name, currName, sizeof(currName));
```
ziemleszcz: Why `memcpy`? Just set `devs[i].name[]` directly.

port/test_register.c
```diff
@@ -0,0 +1,179 @@
...
+TEST_GROUP(test_register_many);
+
+TEST_SETUP(test_register_many)
+{
+	/* prepare names and oids */
+	int firstLetter = 'a';
+	for (int i = 0; i < NDEVS; i++) {
+		/* names will follow pattern: aa, ab, ..., ba, bb, ... */
+
+		if (i != 0 && i % LETTERS_IN_ALPHABET == 0) {
```
ziemleszcz: No need for `firstLetter`:


devs[i].name[0] = 'a' + (i / 26);
devs[i].name[1] = 'a' + (i % 26);
devs[i].name[2] = '\0';

port/test_register.c
julianuziemblo: do you think it's better to copy-paste it? or can it stay this way?

port/test_register.c
julianuziemblo: it's purpose is to ensure we can register a port to any, even non-existent oid. I'll add a FIXME for the future to revise if it's really the intended behavior we want.

port/test_register.c
ziemleszcz: It can stay, but I would prefer an inline function instead of a function-like macro.

ziemleszcz: Still waiting for explanation why `lookup` result is tested with `GREATER_OR_EQUA`.

nalajcie: > It can stay, but I would prefer an inline function instead of a function-like macro.

There is one clear advantage of using macros vs inline functions in test procedures - in fail you will get real line number within the failed test (for inline function you will get line no within this function - makes a lot of difference if you call the same macro a few times in the same testcase).

The rationale was written here: https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/master/libc/common.h#L25 , probably lost along the way while copying the common practice.

I would add similar comment here.

ziemleszcz (approved):

PR #483, opened 2026-05-14
libc/socket: new unix socket tests

ziemleszcz:
When passing a file descriptor with sendmsg(), file status flags are shared with the receiver, while file descriptor flags (such as FD_CLOEXEC) are not.

JIRA: RTOS-1323

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/socket/unix-socket.c
+++ b/libc/socket/unix-socket.c
@@ -40,7 +40,7 @@
 
 #define FAIL_OR_EXIT(pid, msg) \
 	do { \
-		if (pid) \
+		if (pid != 0) \
 			FAIL(msg); \
 		else \
 			exit(1); \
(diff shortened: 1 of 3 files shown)


libc/socket/unix-socket.c
```diff
@@ -530,6 +530,124 @@ TEST(test_unix_socket, dgram_sock_data_and_fd)
...
+		CHILD_ASSERT((flags & O_APPEND) != 0);
+		flags = get_fd_flags(rfd[4]);
+		CHILD_ASSERT(flags >= 0);
+		CHILD_ASSERT((flags & FD_CLOEXEC) == 0);
+
+		if (close_files(rfd, rfdcnt) < 0)
+			exit(2);
+
+		if (stat_files(rfd, rfdcnt, 0) < 0)
+			exit(1);
```
julianuziemblo: 3?

libc/socket/unix-socket.c
```diff
@@ -530,6 +530,124 @@ TEST(test_unix_socket, dgram_sock_data_and_fd)
...
+
+
+	sfdcnt = sizeof(oflags) / sizeof(oflags[0]);
+
+	if (socketpair(AF_UNIX, type, 0, fd) < 0)
+		FAIL("socketpair");
+
+	pid = safe_fork();
+
+	if (pid) {
```
julianuziemblo: `pid != 0` for clarity please

libc/socket/unix-socket.c
```diff
@@ -530,6 +530,124 @@ TEST(test_unix_socket, dgram_sock_data_and_fd)
...
+		O_RDWR | O_CLOEXEC
+	};
+	pid_t pid;
+	size_t sfdcnt, rfdcnt;
+
+
+	sfdcnt = sizeof(oflags) / sizeof(oflags[0]);
+
+	if (socketpair(AF_UNIX, type, 0, fd) < 0)
+		FAIL("socketpair");
```
julianuziemblo, line 554: `TEST_ASSERT_GREATER_OR_EQUAL_MESSAGE`? similar in other cases

libc/socket/unix-socket.c
```diff
@@ -530,6 +530,124 @@ TEST(test_unix_socket, dgram_sock_data_and_fd)
...
+		O_RDWR,
+		O_RDWR | O_NONBLOCK,
+		O_RDWR | O_APPEND,
+		O_RDWR | O_CLOEXEC
+	};
+	pid_t pid;
+	size_t sfdcnt, rfdcnt;
+
+
+	sfdcnt = sizeof(oflags) / sizeof(oflags[0]);
```
julianuziemblo, line 551: its not changed if I see correctly - const? maybe even make it a `#define` so you don't have to allocate `MAX_FD_CNT` ints on the stack? If this is not an issue let me know, but seems a little wastefull for effective 5 sockets used

ziemleszcz: I was following Copilot suggestions here. const or stack usage is not really a concern in test code.

libc/socket/unix-socket.c
ziemleszcz: I’m following the established pattern used in this file. In my opinion, this version is more readable.

julianuziemblo: ok, didn't see it

libc/socket/unix-socket.c
```diff
@@ -530,6 +530,124 @@ TEST(test_unix_socket, dgram_sock_data_and_fd)
...
+		int rfd[MAX_FD_CNT];
+		ssize_t n;
+		int flags;
+
+		n = msg_recv(fd[1], buf, sizeof(buf), rfd, &rfdcnt);
+		if (n != 1 || rfdcnt != sfdcnt)
+			exit(1);
+
+		flags = get_flags(rfd[0]);
+		CHILD_ASSERT(flags >= 0);
```
julianuziemblo, line 613: nit: if this fails, we don't know which condition did it. maybe you could stringify the condition and print it in the macro alongside `__FILE__` and `__LINE__`, essentially mocking the TEST_ASSERT functionality, so that we know what failed and where?

julianuziemblo (commented):
last little point, otherwise LGTM

libc/socket/unix-socket.c
adamgreloch: > print it in the macro alongside __FILE__ and __LINE__

Unsure whether this is CI-friendly. One way around it is to expand `CHILD_ASSERT` to accept the exit code, assign unique exit codes to each assertion and check them in the parent waitpid against zero. That way you can trace the child assertion by parent assertion only

julianuziemblo: would be good too

ziemleszcz: Good idea, but it’s outside the scope of this commit. I can do it later in a separate PR.

julianuziemblo (approved):

PR #488, opened 2026-05-22
libc/stdio: add tests for flockfile, ftrylockfile, funlockfile

julianuziemblo:
YT: RTOS-1334

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `ia32-generic-qemu`

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work:
    - https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/785
    - https://github.com/phoenix-rtos/libphoenix/pull/477

### DIFF
--- a/libc/stdio/stdio_file.c
+++ b/libc/stdio/stdio_file.c
@@ -34,6 +34,7 @@
 #include <errno.h>
 #include <string.h>
 #include <limits.h>
+#include <pthread.h>
 #include <stdlib.h>
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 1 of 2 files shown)


libc/stdio/stdio_file.c
```diff
@@ -1799,3 +1800,120 @@ TEST_GROUP_RUNNER(stdio_fflush)
...
+	arg.f = filep;
+
+	/* lock stream in current thread */
+	flockfile(filep);
+
+	TEST_ASSERT_EQUAL_INT(0, pthread_create(&tid, NULL, flock_thread, &arg));
+
+	while (arg.locked == 0) {
+		usleep(1000);
+	}
```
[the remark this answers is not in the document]
julianuziemblo: CI has its own, externally managed timeout. Adding it here would reduce code managability as we want only one timeout.

sys/mutex/mutex.c
```diff
@@ -233,6 +247,38 @@ static void recursive_thread(void *arg)
...
+	threadArg_t *targ = (threadArg_t *)arg;
+	for (int i = 0; i < 100; i++) {
+		if ((mutexLock(mt_common.mutex)) < 0) {
+			mt_common.thrErrors[targ->id]++;
+			endthread();
+		}
+		if ((mutexTry(mt_common.mutex)) < 0) {
+			mt_common.thrErrors[targ->id]++;
+			endthread();
+		}
```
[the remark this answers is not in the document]
julianuziemblo: If it fails, the test will fail with the external timeout.

libc/stdio/stdio_file.c
```diff
@@ -1799,3 +1800,120 @@ TEST_GROUP_RUNNER(stdio_fflush)
...
+
+	/* another thread must not acquire the lock */
+	TEST_ASSERT_NOT_EQUAL(0, arg.try_result);
+
+	funlockfile(filep);
+
+	pthread_join(tid, NULL);
```
lukkrusz: line 1904: can cause a long jump, which in turn omits the `phtread_join`,  causing the thread to leak.
I'd ensure it's handled properly in the tear-down, before freeing `filep`

lukkrusz (changes requested):
A small potential resource-leak fix and we are good to merge

libc/stdio/stdio_file.c
```diff
@@ -1799,3 +1800,127 @@ TEST_GROUP_RUNNER(stdio_fflush)
...
+TEST_SETUP(stdio_flockfile)
+{
+	filep = NULL;
+}
+
+
+TEST_TEAR_DOWN(stdio_flockfile)
+{
+	if (tid != 0) {
+		pthread_join(tid, NULL);
```
lukkrusz, line 1829: A bit nitpicky but I'd add `tid = 0;` here

libc/stdio/stdio_file.c
```diff
@@ -1799,3 +1800,127 @@ TEST_GROUP_RUNNER(stdio_fflush)
...
+	}
+	if (filep != NULL) {
+		funlockfile(filep);
+		fclose(filep);
+		filep = NULL;
+	}
```
lukkrusz, line 1835: I'm not sure if it's 100% safe. Perhaps we should add a check to test if the file was locked before unlocking it?

lukkrusz (changes requested):

libc/stdio/stdio_file.c
julianuziemblo: maybe we just fclose the file as this destroys the lock and kernel will forcibly unlock it?

lukkrusz: If the lock is set, we'd be destroying a locked mutex, which I'm not sure what it would result in. If we are certain that it's safe then OK but it sounds a bit like a possible undefined behaviour

julianuziemblo: that behavior is defined in our kernel, though maybe not POSIX.

The problem with the alternative solution which is to `ftrylockfile()` is TOCTOU - we can't guarantee that the mutex is still locked after the check, only sliding the window further. Maybe I'll remove any longjpms while we're holding the lock.

libc/stdio/stdio_file.c
lukkrusz, line 923: It's not in the TEST_GROUP_RUNNER

lukkrusz: possibility of a leak, either wrap inside a TEST_PROTECT() or remove in teardown

lukkrusz: wrong stream checked?

lukkrusz: 2nd and 3rd arguments swapped?

lukkrusz: 2nd and 3rd arguments swapped?

lukkrusz, line 1025: The buffer must remain valid until fclose() or another setvbuf, so we need to either make the buf2 global or ensure that filep is closed inside each test

libc/stdio/stdio_file.c
```diff
@@ -1799,3 +1800,146 @@ TEST_GROUP_RUNNER(stdio_fflush)
...
+	TEST_ASSERT_NOT_NULL(filep);
+
+	arg.f = filep;
+
+	/* lock stream in current thread */
+	flockfile(filep);
+	do {
+		err = pthread_create(&tid, NULL, flock_thread, &arg);
+		if (err != 0) {
+			break;
```
lukkrusz, line 1928: tid may hold garbage here

libc/stdio/stdio_file.c
```diff
@@ -1799,3 +1800,146 @@ TEST_GROUP_RUNNER(stdio_fflush)
...
+		}
+
+		while (arg.tried == 0) {
+			usleep(1000);
+		}
+	} while (0);
```
lukkrusz, line 1940: I'd put some kind of a timeout here just for sanity, as a good practice

lukkrusz (changes requested):
I run a coding agent through this file and found a couple of new issues we should address

libc/stdio/stdio_file.c
julianuziemblo: filep and filep2 are closed inside each test in TEST_TEAR_DOWN

lukkrusz: Sorry, didn't explain well, we should ensure that buf2 outlives filep. One way could be making buf2 global static or closing filep before the test terminates (buf2 ceases to exist before the teardown)

lukkrusz (changes requested):
Test failures are somewhat related to this PR and need to be looked into

lukkrusz (dismissed):
LGTM

PR #498, opened 2026-07-02
libc: add posix tests for statvfs

lukkrusz:
Assisted-by:  Claude Opus 4.8 and Opus 5 
YT: CI-678

## Types of changes

- [x] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/libc/statvfs/statvfs.c
+++ b/libc/statvfs/statvfs.c
@@ -1,0 +1,356 @@
+/*
+ * Phoenix-RTOS
+ *
+ * POSIX.1-2017 standard library functions tests
+ * HEADER:
+ *    - <sys/statvfs.h>
+ * TESTED:
+ *    - statvfs()
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Lukasz Kruszynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <sys/statvfs.h>
+#include <sys/stat.h>
+#include <errno.h>
+#include <limits.h>
+#include <string.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <stdlib.h>
+
+#include "unity_fixture.h"
+
+#define STATVFS_TEST_DIR    "/tmp/test_statvfs_dir"
+#define STATVFS_TEST_FILE   "/tmp/test_statvfs_file"
+#define STATVFS_SYMLINK     "/tmp/test_statvfs_link"
+#define STATVFS_NOPERM_DIR  "/tmp/test_statvfs_noperm"
+#define STATVFS_NOPERM_FILE "/tmp/test_statvfs_noperm/inner"
+#define STATVFS_LOOP_LINK_A "/tmp/test_statvfs_loop_a"
+#define STATVFS_LOOP_LINK_B "/tmp/test_statvfs_loop_b"
+
+static struct {
+	int fd;
+} test_common;
+
+TEST_GROUP(statvfs_statvfs);
+
+TEST_SETUP(statvfs_statvfs)
+{
+	test_common.fd = -1;
+	unlink(STATVFS_TEST_FILE);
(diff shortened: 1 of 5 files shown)


libc/test.yaml
```diff
@@ -113,3 +113,8 @@ test:
       targets:
         include: [host-generic-pc]
 
+    - name: statvfs
+      execute: test-libc-statvfs
+      targets:
+        include: [host-generic-pc]
+        exclude: [armv7m7-imxrt117x-evk, armv7m7-imxrt106x-evk, armv7m4-stm32l4x6-nucleo, armv7r5f-zynqmp-qemu]
```
mateusz-bloch: Why are these targets excluded?
```
        exclude: [armv7m7-imxrt117x-evk, armv7m7-imxrt106x-evk, armv7m4-stm32l4x6-nucleo, armv7r5f-zynqmp-qemu]
```
I'm pretty sure that armv7m7-imxrt117x-evk, armv7m7-imxrt106x-evk should be able to run these tests. They probably only need the /tmp directory to be created, as it doesn't exist on the dummy fs.

libc/statvfs/statvfs.c
```diff
@@ -0,0 +1,330 @@
...
+	ret = statvfs("/proc", &buf);
+	if (ret == -1) {
+		TEST_IGNORE_MESSAGE("/proc not available");
+	}
+	TEST_ASSERT_EQUAL_INT(0, ret);
+
+	if ((buf.f_flag & ST_RDONLY) == 0) {
+		TEST_IGNORE_MESSAGE("/proc not reported as read-only on this system");
+	}
+	TEST_ASSERT_EQUAL_UINT(ST_RDONLY, (buf.f_flag & ST_RDONLY));
```
mateusz-bloch: The test will never fail, it will return ignore or pass

libc/statvfs/statvfs.c
```diff
@@ -0,0 +1,330 @@
...
+	int ret;
+
+	errno = 0;
+	ret = statvfs("", &buf);
+	TEST_ASSERT_EQUAL_INT(-1, ret);
+	TEST_ASSERT_EQUAL_INT(ENOENT, errno);
+}
+
+
+TEST(statvfs_statvfs, statvfs_enotdir_prefix_not_dir)
```
mateusz-bloch, line 226: Which POSIX version is this expected to comply with? From what I can see, POSIX.1-2017 includes the following ENOTDIR case:

Image

So should also be a test case where the path points to a regular file and ends with a trailing slash.

mateusz-bloch (changes requested):

libc/statvfs/statvfs.c
lukkrusz: Fixed

libc/statvfs/statvfs.c
lukkrusz: I added a case below to exercise the said scenario. Thanks

libc/test.yaml
lukkrusz: Fixed, with the /tmp dir now being ensured to exist before the test runs.
I'm a bit mixed with leaving it post test in case it's created but it seems to be the case with libc/misc tests

lukkrusz: Update: added armv8m55 target to the excluded list.

mateusz-bloch: > Fixed, with the /tmp dir now being ensured to exist before the test runs. I'm a bit mixed with leaving it post test in case it's created but it seems to be the case with libc/misc tests

Also don't like leaving anything after the tests, ofc on targets that only use dummyfs, it will get cleaned up after a device reboot anyway

mateusz-bloch: It would probably be worth adding a short comment explaining why this test is disabled on these targets

mateusz-bloch (approved):
LGTM

damianloew (commented):
Let's use the model name in the following format in git commit message:
`Assisted by: claude-opus-4.8`

